PR #79, opened 2014-01-28
Proposal Status cannot transition from any state to any other. This will...

miteshashar:
... be validated. Needed to eliminate some evident edge use cases while writing the ticketing automation logic.


jace:
This is no longer needed as we have workflows now.

PR #110, opened 2015-01-13
added map marker field for venue form

shreyas-satish:


shreyas-satish:
This depends on https://github.com/hasgeek/baseframe/pull/75

PR #112, opened 2015-01-30
Optional RSVP for Proposal Spaces

shreyas-satish:


alembic/versions/14d1424b47_.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('rsvp',
+    sa.Column('created_at', sa.DateTime(), nullable=False),
+    sa.Column('updated_at', sa.DateTime(), nullable=False),
+    sa.Column('user_id', sa.Integer(), nullable=False),
+    sa.Column('proposal_space_id', sa.Integer(), nullable=False),
+    sa.Column('rsvp_action', sa.Enum('M', 'N', 'Y', name='rsvp_action'), nullable=True),
```
jace: Prefix with table name to make it `rsvp_rsvp_action`. Nullable should be `False` is there is no occasion for a record with no value.

alembic/versions/14d1424b47_.py
```diff
@@ -0,0 +1,38 @@
...
+    sa.ForeignKeyConstraint(['proposal_space_id'], ['proposal_space.id'], ),
+    sa.ForeignKeyConstraint(['user_id'], ['user.id'], ),
+    sa.PrimaryKeyConstraint('user_id', 'proposal_space_id'),
```
jace: Your primary key order should be `('proposal_space_id', 'user_id')` as you'll be looking up RSVPs from the ProposalSpace table most of the time (SQL indexes are left to right). Also reverse the order of the columns above.

alembic/versions/14d1424b47_.py
```diff
@@ -0,0 +1,38 @@
...
+    sa.UniqueConstraint('user_id', 'proposal_space_id')
+    )
+    op.add_column(u'proposal_space', sa.Column('allow_rsvp', sa.Boolean(), nullable=True))
+    ### end Alembic commands ###
+
+
+def downgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column(u'proposal_space', 'allow_rsvp')
+    op.drop_table('rsvp')
```
jace: You need to explicitly drop the enum type in PostgreSQL. Since Alembic doesn't have an operation for this, use `op.execute(sa.text('DROP TYPE rsvp_rsvp_action_enum'))`.

alembic/versions/14d1424b47_.py
```diff
@@ -0,0 +1,38 @@
...
+    sa.Column('updated_at', sa.DateTime(), nullable=False),
+    sa.Column('user_id', sa.Integer(), nullable=False),
+    sa.Column('proposal_space_id', sa.Integer(), nullable=False),
+    sa.Column('rsvp_action', sa.Enum('M', 'N', 'Y', name='rsvp_action'), nullable=True),
+    sa.ForeignKeyConstraint(['proposal_space_id'], ['proposal_space.id'], ),
+    sa.ForeignKeyConstraint(['user_id'], ['user.id'], ),
+    sa.PrimaryKeyConstraint('user_id', 'proposal_space_id'),
+    sa.UniqueConstraint('user_id', 'proposal_space_id')
+    )
+    op.add_column(u'proposal_space', sa.Column('allow_rsvp', sa.Boolean(), nullable=True))
```
jace: Make this `nullable=False, server_default='0'` and follow up with `op.alter_column('proposal_space', 'allow_rsvp', server_default=None)`.

funnel/models/rsvp.py
```diff
@@ -0,0 +1,45 @@
...
+from coaster.utils import LabeledEnum
+from baseframe import __
+
+__all__ = ['RSVP', 'RSVP_ACTION', 'RSVPMixin']
+
+
+class RSVP_ACTION(LabeledEnum):
+    RSVP_Y = ('Y', {'label': "I'm going", 'category': 'success', 'order': 1})
+    RSVP_N = ('N', {'label': 'Not going', 'category': 'danger', 'order': 2})
+    RSVP_M = ('M', {'label': 'Maybe', 'category': 'default', 'order': 3})
```
jace: Labels need to be wrapped in `__(...)`

funnel/models/rsvp.py
```diff
@@ -0,0 +1,45 @@
...
+    user_id = db.Column(db.Integer, db.ForeignKey('user.id'), nullable=False, primary_key=True)
+    proposal_space_id = db.Column(db.Integer, db.ForeignKey('proposal_space.id'), nullable=False, primary_key=True)
+    rsvp_action = db.Column(db.Enum(*RSVP_ACTION.keys(), name='rsvp_action'), default=RSVP_ACTION)
+
+    def __init__(self, **kwargs):
+        super(RSVP, self).__init__(**kwargs)
+
+class RSVPMixin(object):
+
+    allow_rsvp = db.Column(db.Boolean, default=False)
```
jace: Add `nullable=False`.

funnel/models/rsvp.py
```diff
@@ -0,0 +1,45 @@
...
+
+class RSVP_ACTION(LabeledEnum):
+    RSVP_Y = ('Y', {'label': "I'm going", 'category': 'success', 'order': 1})
+    RSVP_N = ('N', {'label': 'Not going', 'category': 'danger', 'order': 2})
+    RSVP_M = ('M', {'label': 'Maybe', 'category': 'default', 'order': 3})
+
+
+class RSVP(TimestampMixin, db.Model):
+    __tablename__ = 'rsvp'
+    __table_args__ = (db.UniqueConstraint('user_id', 'proposal_space_id'),)
```
jace: ProposalSpace followed by User, and move `__table_args__` below the column definitions.

funnel/models/rsvp.py
```diff
@@ -0,0 +1,45 @@
...
+    
+    user_id = db.Column(db.Integer, db.ForeignKey('user.id'), nullable=False, primary_key=True)
+    proposal_space_id = db.Column(db.Integer, db.ForeignKey('proposal_space.id'), nullable=False, primary_key=True)
+    rsvp_action = db.Column(db.Enum(*RSVP_ACTION.keys(), name='rsvp_action'), default=RSVP_ACTION)
+
+    def __init__(self, **kwargs):
+        super(RSVP, self).__init__(**kwargs)
+
+class RSVPMixin(object):
```
jace: Since the RSVP model only works for ProposalSpace, you no longer need a mixin. Move all this into the ProposalSpace model.

funnel/models/rsvp.py
```diff
@@ -0,0 +1,45 @@
...
+__all__ = ['RSVP', 'RSVP_ACTION', 'RSVPMixin']
+
+
+class RSVP_ACTION(LabeledEnum):
+    RSVP_Y = ('Y', {'label': "I'm going", 'category': 'success', 'order': 1})
+    RSVP_N = ('N', {'label': 'Not going', 'category': 'danger', 'order': 2})
+    RSVP_M = ('M', {'label': 'Maybe', 'category': 'default', 'order': 3})
+
+
+class RSVP(TimestampMixin, db.Model):
```
jace: Class names shouldn't be in all uppercase. Make this `Rsvp`.

funnel/models/rsvp.py
```diff
@@ -0,0 +1,45 @@
...
+        if self.allow_rsvp:
+            return sorted(RSVP_ACTION.items(), key=lambda action: action[1]['order'])
+        else:
+            return []
+
+    def rsvp_responses_count(self, action):
+        return RSVP.query.filter_by(proposal_space_id=self.id, rsvp_action=action).count()
+
+    def user_rsvp_status(self, user_id):
+        rsvp = RSVP.query.filter_by(proposal_space_id=self.id, user_id=user_id).first()
```
jace: Use `query.get((self.id, user.id))` and take a `user` model parameter, not `user_id`. Never pass ids between functions. The risk of accidentally passing a value from the wrong model is pretty high and there's no type safety since they are all integers.

Using `query.get` does a primary key lookup using the SQLAlchemy session's identity map cache. This can save a database roundtrip if the data has already been loaded.

funnel/templates/layout.html
```diff
@@ -42,3 +42,18 @@
 {% block basecontent -%}
   {% block content %}{% endblock %}
 {%- endblock %}
+
+{% macro rsvp_script() %}
```
jace: Don't put macros in layout.html. Put them in a macros.html.

funnel/templates/space.html
```diff
@@ -123,6 +123,9 @@
     {%- endif -%}
   </strong>
 </p>
+{%- if space.allow_rsvp %}
+  {% include 'rsvp.html' %}
```
jace: Don't use includes. Move this into a macro or inline it.

funnel/views/space.py
```diff
@@ -2,20 +2,19 @@
...
 from .. import app
-from ..models import db, Profile, ProposalSpace, ProposalSpaceRedirect, ProposalSpaceSection, Proposal, PROPOSALSTATUS
+from ..models import db, Profile, ProposalSpace, ProposalSpaceRedirect, ProposalSpaceSection, Proposal, PROPOSALSTATUS, RSVP
 from ..forms import ProposalSpaceForm
 from .proposal import proposal_headers, proposal_data, proposal_data_flat
 from .schedule import schedule_data
 from .venue import venue_data, room_data
 from .section import section_data
 
-
```
jace: You need two blank lines between imports and the first function.

funnel/views/space.py
```diff
@@ -141,3 +140,18 @@ def space_edit(profile, space):
         flash(_("Your changes have been saved"), 'info')
         return redirect(space.url_for(), code=303)
     return render_form(form=form, title=_("Edit proposal space"), submit=_("Save changes"))
+
```
jace: Additional blank line needed.

funnel/views/space.py
```diff
@@ -141,3 +140,18 @@ def space_edit(profile, space):
...
+@app.route('/<space>/rsvp', methods=['POST'], subdomain='<profile>')
+@load_models(
+    (Profile, {'name': 'profile'}, 'g.profile'),
+    ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
+    permission='view')
```
jace: RSVP needs a new permission. `view` is available to anonymous users.

funnel/views/space.py
```diff
@@ -141,3 +140,18 @@ def space_edit(profile, space):
...
+def rsvp(profile, space):
+    rsvp = RSVP.query.filter_by(proposal_space_id=space.id, user_id=g.user.id).first()
```
jace: Use `query.get((space.id, g.user.id))`

funnel/views/space.py
```diff
@@ -141,3 +140,18 @@ def space_edit(profile, space):
         flash(_("Your changes have been saved"), 'info')
         return redirect(space.url_for(), code=303)
     return render_form(form=form, title=_("Edit proposal space"), submit=_("Save changes"))
+
+@app.route('/<space>/rsvp', methods=['POST'], subdomain='<profile>')
+@load_models(
```
jace: Add `@lastuser.requires_login` before `@load_models`.

funnel/views/space.py
```diff
@@ -141,3 +140,18 @@ def space_edit(profile, space):
...
+    (Profile, {'name': 'profile'}, 'g.profile'),
+    ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
+    permission='view')
+def rsvp(profile, space):
+    rsvp = RSVP.query.filter_by(proposal_space_id=space.id, user_id=g.user.id).first()
+    if rsvp:
+        rsvp.rsvp_action = request.form['rsvp_action']
```
jace: Fatal! You are not checking the value of `request.form['rsvp_action']` nor checking for CSRF. *Always* use a Form when handling POST data to ensure you don't accidentally miss these two.

funnel/views/space.py
```diff
@@ -141,3 +140,18 @@ def space_edit(profile, space):
...
+    else:
+        rsvp = RSVP(proposal_space_id=space.id, user_id=g.user.id, rsvp_action=request.form['rsvp_action'])
```
jace: When you make a new entity, pass it objects, not ids. Use `proposal_space=space, user=g.user, rsvp_action=form.rsvp.data`

alembic/versions/14d1424b47_.py
```diff
@@ -0,0 +1,40 @@
...
+# revision identifiers, used by Alembic.
+revision = '14d1424b47'
+down_revision = '1c496c114b6'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
```
jace: Since these are reviewed migrations, these lines should be removed.

alembic/versions/14d1424b47_.py
```diff
@@ -0,0 +1,40 @@
...
+    op.create_table('rsvp',
+    sa.Column('created_at', sa.DateTime(), nullable=False),
+    sa.Column('updated_at', sa.DateTime(), nullable=False),
+    sa.Column('proposal_space_id', sa.Integer(), nullable=False),
+    sa.Column('user_id', sa.Integer(), nullable=False),
+    sa.Column('rsvp_action', sa.Enum('M', 'N', 'Y', 'A', name='rsvp_rsvp_action_enum'), nullable=False, server_default='A'),
```
jace: Since `server_default` isn't present in the Python code, this technically means the migration will result in a db schema that is different from a fresh installation. Remove the server_default from the schema, or add it in Python code (our convention is to remove it from the schema with an alter_column. Since this is a new table, you don't need it in the first place).

alembic/versions/14d1424b47_.py
```diff
@@ -0,0 +1,40 @@
...
+    sa.ForeignKeyConstraint(['proposal_space_id'], ['proposal_space.id'], ),
+    sa.ForeignKeyConstraint(['user_id'], ['user.id'], ),
+    sa.PrimaryKeyConstraint('proposal_space_id', 'user_id'),
+    sa.UniqueConstraint('proposal_space_id', 'user_id')
+    )
+    op.add_column(u'proposal_space', sa.Column('allow_rsvp', sa.Boolean(), nullable=True))
+    op.alter_column('proposal_space', 'allow_rsvp', server_default=None)
```
jace: These two should be:

```python
op.add_column(u'proposal_space', sa.Column('allow_rsvp', sa.Boolean(), nullable=False, server_default='0'))
op.alter_column('proposal_space', 'allow_rsvp', server_default=None)
```

funnel/models/rsvp.py
```diff
@@ -0,0 +1,48 @@
...
+from .user import User
+
+__all__ = ['Rsvp']
+
+
+class RSVP_ACTION(LabeledEnum):
+    RSVP_Y = ('Y', {'label': __("I'm going"), 'category': 'success', 'order': 1, 'active': True})
+    RSVP_N = ('N', {'label': __("Not going"), 'category': 'danger', 'order': 2, 'active': True})
+    RSVP_M = ('M', {'label': __("Maybe"), 'category': 'default', 'order': 3, 'active': True})
+    RSVP_A = ('A', {'label': __("Awaiting"), 'category': 'default', 'order': 4, 'active': False})
```
jace: `__order__ = [RSVP_Y, RSVP_N, RSVP_M]` and remove RSVP_A for now.

funnel/models/rsvp.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Rsvp(TimestampMixin, db.Model):
+    __tablename__ = 'rsvp'
+    
+    user_id = db.Column(db.Integer, db.ForeignKey('user.id'), nullable=False, primary_key=True)
+    proposal_space_id = db.Column(db.Integer, db.ForeignKey('proposal_space.id'), nullable=False, primary_key=True)
+    proposal_space = db.relationship(ProposalSpace, primaryjoin=proposal_space_id == ProposalSpace.id)
+    user = db.relationship(User, primaryjoin=user_id == User.id)
+
+    rsvp_action = db.Column(db.Enum(*RSVP_ACTION.keys(), name='rsvp_action'), default=RSVP_ACTION, nullable=False)
```
jace: `RSVP_ACTION` is a class and can't be the default value. Drop it here, or add `default='A'`.

funnel/models/rsvp.py
```diff
@@ -0,0 +1,48 @@
...
+        super(Rsvp, self).__init__(**kwargs)
+
+    @classmethod
+    def rsvp_actions(self, space):
+        # A new list is being generated with the responses count appended
+        def append_rsvp_responses_count(action):
+            action[1]['responses_count'] = Rsvp.query.filter_by(proposal_space=space, rsvp_action=action[0]).count()
+            return action
+        return map(append_rsvp_responses_count,
+                   sorted((item for item in RSVP_ACTION.items() if item[1]['active']), key=lambda action: action[1]['order']))
```
jace: This could simply be:

```python
return OrderedDict([(key, Rsvp.query.filter_by(proposal_space=space, rsvp_action=key)) for key in RSVP_ACTION])
```

jace: Add `from collections import OrderedDict`

funnel/models/rsvp.py
```diff
@@ -0,0 +1,48 @@
...
+
+    rsvp_action = db.Column(db.Enum(*RSVP_ACTION.keys(), name='rsvp_action'), default=RSVP_ACTION, nullable=False)
+
+    __table_args__ = (db.UniqueConstraint('proposal_space_id', 'user_id'),)
+
+    def __init__(self, **kwargs):
+        super(Rsvp, self).__init__(**kwargs)
+
+    @classmethod
+    def rsvp_actions(self, space):
```
jace: Class methods take a `cls` parameter instead of `self` (it's just a convention so your code will still work, but you'll leave someone else mightily confused later).

funnel/models/rsvp.py
jace: Also, in class methods, we generally use the `cls` parameter instead of the explicit class name, so the query should be `cls.query`, not `Rsvp.query`.

funnel/models/space.py
```diff
@@ -64,6 +64,8 @@ class ProposalSpace(BaseScopedNameMixin, db.Model):
     #: Redirect URLs from Funnel to Talkfunnel
     legacy_name = db.Column(db.Unicode(250), nullable=True, unique=True)
 
+    allow_rsvp = db.Column(db.Boolean, default=False)
```
jace: Add `nullable=False`

funnel/templates/rsvp.html
```diff
@@ -0,0 +1,2 @@
+{% from "macros.html" import rsvp with context %}
```
jace: Avoid importing `with context`. Pass the macro whatever it needs. Makes debugging easier if something changes in future.

funnel/views/space.py
```diff
@@ -141,3 +141,23 @@ def space_edit(profile, space):
...
+@lastuser.requires_login
+@load_models(
+    (Profile, {'name': 'profile'}, 'g.profile'),
+    ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
+    permission='view')
+def rsvp(profile, space):
+    if request.method == 'POST' and Form().validate_csrf_data(request.form['csrf_token']):
+        rsvp = Rsvp.query.get((space.id, g.user.id))
+        if rsvp:
+            rsvp.rsvp_action = request.form['rsvp_action']
```
jace: You're still not validating the value of `rsvp_action`. Just make a new form with one RadioField, with `choices=[(key, value['label']) for key, value in RSVP_ACTION.items()]`. Use it like this:

```python
form = RsvpForm()
if form.validate_on_submit():
    rsvp = Rsvp.get_for(space, g.user, create=True)
    form.populate_obj(rsvp)
    db.session.commit()
```

1. No need to make this code aware of Form internals like `validate_csrf_data` (it changes occasionally in WTForms without warning as it's meant to be internal detail).
2. Don't deal with ids like `space.id` outside the model itself. If the primary key order switches to `user_id, space_id` in future, you'll have to refactor across your code and errors won't show up easily because there's no validation on integers. We've had serious nightmares developing hacknight.in because of practices like this where the code worked as long as there was limited data in production and broke when users started adding data.

funnel/views/space.py
```diff
@@ -141,3 +141,23 @@ def space_edit(profile, space):
...
+        else:
+            rsvp = Rsvp(proposal_space=space, user=g.user, rsvp_action=request.form['rsvp_action'])
+            db.session.add(rsvp)
+        db.session.commit()
+        return make_response(render_template('rsvp.html', space=space, rsvp=rsvp, rsvp_actions=Rsvp.rsvp_actions(space), user_rsvp_status=Rsvp.user_rsvp_status(space, g.user), rsvp_form=Form()))
+    else:
+        return make_response("Invalid request", 400)
```
jace: Use `abort(400)`. Import it form `flask`.

PR #114, opened 2015-04-06
added a buy tickets url column for proposal space

shreyas-satish:
- added buy_tickets_url column in ProposalSpace
- added the corresponding field in ProposalSpaceForm
- added a 'Buy Tickets' button, which shows if the buy_tickets_url is set.


funnel/templates/space.html
```diff
@@ -125,9 +125,16 @@
   </strong>
 </p>
 {%- if space.sessions %}
-  <p>
-    <a href="{{ space.url_for('schedule') }}" class='btn btn-primary btn-sm'>{% trans %}View Schedule{% endtrans %}</a>
-  </p>
+  <ul class="list-inline">
+    <li>
+      <a href="{{ space.url_for('schedule') }}" class='btn btn-primary btn-sm'>{% trans %}View Schedule{% endtrans %}</a>
```
jace: There needs to be an if condition for this too. A new proposal space has no schedule. However, this requires a separate issue.

PR #116, opened 2015-04-14
expose space_id in JSON api

karthikb351:
- There is no other reliable way to identify unique spaces


karthikb351:
I goofed up a one line commit.

PR #117, opened 2015-04-20
label speaker's location

shreyas-satish:
- Adds a label (outstation, same or invalid location) to each speaker's location in the proposals table.

### DIFF
--- a/funnel/models/proposal.py
+++ b/funnel/models/proposal.py
@@ -11,6 +11,7 @@ from baseframe import __
 from sqlalchemy.ext.hybrid import hybrid_property
 from flask import request
 from pytz import timezone, utc, UnknownTimeZoneError
+from funnel.util import geonameid_from_location
 
 __all__ = ['PROPOSALSTATUS', 'Proposal', 'ProposalRedirect']
 
@@ -230,6 +231,15 @@ class Proposal(BaseScopedIdNameMixin, CoordinatesMixin, db.Model):
                     votes_bydate[groupname].setdefault(date, 0)
                     votes_bydate[groupname][date] += -1 if vote.votedown else +1
         return votes_bydate
+
+    def is_outstation_speaker(self, space_location_geonameid):
+        """ determines if a proposal's location is different from the space location
+            returns None if location didn't resolve with hascore.
+        """
+        location_geonameid = geonameid_from_location(self.location)
+        if not location_geonameid:
+            return None
+        return location_geonameid != space_location_geonameid
 
     def permissions(self, user, inherited=None):
         perms = super(Proposal, self).permissions(user, inherited)
--- a/funnel/models/proposal.py
+++ b/funnel/models/proposal.py
@@ -11,7 +11,8 @@ from baseframe import __
 from sqlalchemy.ext.hybrid import hybrid_property
 from flask import request
 from pytz import timezone, utc, UnknownTimeZoneError
-from funnel.util import geonameid_from_location
+from werkzeug.utils import cached_property
+from ..util import geonameid_from_location
 
 __all__ = ['PROPOSALSTATUS', 'Proposal', 'ProposalRedirect']
 
@@ -187,6 +188,10 @@ class Proposal(BaseScopedIdNameMixin, CoordinatesMixin, db.Model):
     def confirmed(self):
         return self.status == PROPOSALSTATUS.CONFIRMED
 
+    @cached_property
+    def location_geonameid(self):
+        return geonameid_from_location(self.location)
+
     def getnext(self):
         return Proposal.query.filter(Proposal.proposal_space == self.proposal_space).filter(
             Proposal.id != self.id).filter(
@@ -231,15 +236,6 @@ class Proposal(BaseScopedIdNameMixin, CoordinatesMixin, db.Model):
                     votes_bydate[groupname].setdefault(date, 0)
                     votes_bydate[groupname][date] += -1 if vote.votedown else +1
         return votes_bydate
-
-    def is_outstation_speaker(self, space_location_geonameid):
-        """ determines if a proposal's location is different from the space location
-            returns None if location didn't resolve with hascore.
-        """
-        location_geonameid = geonameid_from_location(self.location)
-        if not location_geonameid:
-            return None
-        return location_geonameid != space_location_geonameid
 
     def permissions(self, user, inherited=None):
         perms = super(Proposal, self).permissions(user, inherited)
--- a/funnel/models/space.py
+++ b/funnel/models/space.py
@@ -5,6 +5,7 @@ from . import db, TimestampMixin, BaseScopedNameMixin, MarkdownColumn
 from .user import User, Team
 from .profile import Profile
 from .commentvote import VoteSpace, CommentSpace, SPACETYPE
+from funnel.util import geonameid_from_location
 
 __all__ = ['SPACESTATUS', 'ProposalSpace', 'ProposalSpaceRedirect']
 
@@ -123,6 +124,10 @@ class ProposalSpace(BaseScopedNameMixin, db.Model):
             confirmed=basequery.filter_by(status=PROPOSALSTATUS.CONFIRMED).order_by(db.desc('created_at')).all(),
             unconfirmed=basequery.filter(Proposal.status != PROPOSALSTATUS.CONFIRMED, Proposal.status != PROPOSALSTATUS.DRAFT).order_by(db.desc('created_at')).all())
         return response
+
+    @property
+    def location_geonameid(self):
+        return geonameid_from_location(self.datelocation)
 
     def user_in_group(self, user, group):
         for grp in self.usergroups:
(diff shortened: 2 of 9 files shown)


funnel/static/sass/_layout.scss
```diff
@@ -663,3 +663,7 @@ div.markdown-field a.button {
 .rsvp-header {
     padding: 1px 0px 1px 15px
 }
+
+.outstation-speaker {
+    vertical-align: middle;
```
jace: You shouldn't need this class. Check what is wrong in the container that icons aren't getting aligned properly.

funnel/templates/space.html
```diff
@@ -42,6 +42,20 @@
                 {%- if proposal.user != proposal.speaker %}
                   <small>via {{ proposal.user.pickername }}</small>
                 {%- endif %}
+                {%- if g.user and 'edit-space' in g.permissions %}
+                  {% with %}
+                    {% set oustation_speaker = proposal.location | is_outstation_speaker(speaker_locations, space_location) %}
```
jace: Ugh. Don't use `set` if it can be avoided, and don't use template filters. This should be a function on the model.

funnel/views/helpers.py
```diff
@@ -17,3 +23,40 @@ def localize_date(date, from_tz=utc, to_tz=utc):
             to_tz = pytz_timezone(to_tz)
         return from_tz.localize(date).astimezone(to_tz).replace(tzinfo=None)
     return date
+
+
+def location_geodata(location):
```
jace: This needs caching.

funnel/views/helpers.py
```diff
@@ -17,3 +23,40 @@ def localize_date(date, from_tz=utc, to_tz=utc):
...
+        response = requests.get(url, params={'name': location}).json()
+        if response.get('status') == 'ok':
+            result = response.get('result', {})
+            if isinstance(result, (list, tuple)):
+                result = {r['geonameid']: r for r in result}
+            return result
+    return {}
+
+
+def parsed_location_geodata(data):
```
jace: This name is confusing. Change to `parse_location_for_geodata`.

funnel/views/helpers.py
```diff
@@ -17,3 +23,40 @@ def localize_date(date, from_tz=utc, to_tz=utc):
...
+    if 'HASCORE_SERVER' in app.config:
+        url = urljoin(app.config['HASCORE_SERVER'], '/1/geo/parse_locations')
+        response = requests.get(url, params={'q': data}).json()
```
jace: Change parameter name to `text`. `data` suggests an unknown structure.

funnel/views/helpers.py
```diff
@@ -17,3 +23,40 @@ def localize_date(date, from_tz=utc, to_tz=utc):
...
+        return [field['geoname'] for field in response['result'] if 'geoname' in field.keys()][0]
+    return {}
```
jace: Your two return statements return different data types.

funnel/views/helpers.py
```diff
@@ -17,3 +23,40 @@ def localize_date(date, from_tz=utc, to_tz=utc):
...
+
+
+def format_location(location):
+    return str(location).strip().lower()
```
jace: Move this inline wherever it's used.

funnel/views/helpers.py
```diff
@@ -17,3 +23,40 @@ def localize_date(date, from_tz=utc, to_tz=utc):
...
+
+
+@app.template_filter('is_outstation_speaker')
+def is_outstation_speaker(speaker_location_name, speaker_locations, space_location):
```
jace: This should not be a template filter since it's very special purpose. Move it into the model.

funnel/views/helpers.py
jace: Actually, since the first returns a dict inside a list, you should disambiguate by returning the actual data that is interesting: the integer `geonameid` or `None`.

funnel/views/space.py
```diff
@@ -85,10 +86,16 @@ def space_new(profile):
...
     sections = ProposalSpaceSection.query.filter_by(proposal_space=space, public=True).order_by('title').all()
     rsvp_form = RsvpForm(obj=space.rsvp_for(g.user))
+    space_location = parsed_location_geodata(space.datelocation)
+    speaker_locations = {}
+    proposal_locations = set([format_location(proposal.location) for proposal in space.proposals])
+    for location in proposal_locations:
+        speaker_locations.setdefault(format_location(location), location_geodata(format_location(location)))
     return render_template('space.html', space=space, description=space.description, sections=sections,
-        PROPOSALSTATUS=PROPOSALSTATUS, rsvp_form=rsvp_form)
+        PROPOSALSTATUS=PROPOSALSTATUS, rsvp_form=rsvp_form, space_location=space_location, speaker_locations=speaker_locations)
```
jace: Don't pass these here, let them be properties on the models.

funnel/models/proposal.py
```diff
@@ -11,6 +11,7 @@
 from sqlalchemy.ext.hybrid import hybrid_property
 from flask import request
 from pytz import timezone, utc, UnknownTimeZoneError
+from funnel.util import geonameid_from_location
```
jace: Use `from ..util`

funnel/models/proposal.py
```diff
@@ -231,6 +232,15 @@ def votes_by_date(self):
                     votes_bydate[groupname][date] += -1 if vote.votedown else +1
         return votes_bydate
 
+    def is_outstation_speaker(self, space_location_geonameid):
```
jace: You could now change this to `is_matching_geonameid` and accept a `geonameid` parameter. Makes this a bit more generic.

funnel/models/space.py
```diff
@@ -5,6 +5,7 @@
 from .user import User, Team
 from .profile import Profile
 from .commentvote import VoteSpace, CommentSpace, SPACETYPE
+from funnel.util import geonameid_from_location
```
jace: Use `from ..util`

funnel/models/space.py
```diff
@@ -124,6 +125,10 @@ def proposals_by_confirmation(self):
             unconfirmed=basequery.filter(Proposal.status != PROPOSALSTATUS.CONFIRMED, Proposal.status != PROPOSALSTATUS.DRAFT).order_by(db.desc('created_at')).all())
         return response
 
+    @property
```
jace:
```python
from werkzeug.utils import cached_property
@cached_property
```

Saves a roundtrip to Redis if you use this multiple times in the same request.

funnel/models/proposal.py
jace: Actually, dump this entire function and make it `location_geonameid` like in `ProposalSpace`. Do the comparison in the template.

funnel/templates/space.html
```diff
@@ -42,6 +42,17 @@
                 {%- if proposal.user != proposal.speaker %}
                   <small>via {{ proposal.user.pickername }}</small>
                 {%- endif %}
+                {%- if g.user and 'edit-space' in g.permissions %}
+                  <span class='outstation-speaker'>
+                    {%- if proposal.is_outstation_speaker(space.location_geonameid) %}
```
jace: Change to `{% if proposal.location_geonameid != space.location_geonameid %}`

jace: Don't bother about unknown locations. You only want to mark when the speaker's location is not matching the event's location.

funnel/templates/space.html
```diff
@@ -42,6 +42,17 @@
                 {%- if proposal.user != proposal.speaker %}
                   <small>via {{ proposal.user.pickername }}</small>
                 {%- endif %}
+                {%- if g.user and 'edit-space' in g.permissions %}
+                  <span class='outstation-speaker'>
```
jace: The `<span>` is unnecessary now.

funnel/templates/space.html
```diff
@@ -42,6 +42,13 @@
                 {%- if proposal.user != proposal.speaker %}
                   <small>via {{ proposal.user.pickername }}</small>
                 {%- endif %}
+                {%- if g.user and 'edit-space' in g.permissions %}
+                  {%- if proposal.location_geonameid != space.location_geonameid %}
+                    <i class="fa fa-plane" title="Outstation Speaker - {{ proposal.location }}"></i>
+                  {%- else %}
+                    <i class="fa fa-map-marker" title="Same location - {{ proposal.location }}"></i>
```
jace: Skip the title tags -- hints on hover are bad UI and don't work on touch screens. Add an aria label instead for screen readers.

funnel/util.py
```diff
@@ -0,0 +1,19 @@
...
+def geonameid_from_location(text):
+    """ Accepts a string, checks hascore if there's a location embedded
+        in the string, and returns the geonameid for a location found.
+        Eg: "Bangalore" -> 1277333
+        To detect multiple locations, split them up and pass each location individually
+    """
+    if 'HASCORE_SERVER' in app.config:
+        url = urljoin(app.config['HASCORE_SERVER'], '/1/geo/parse_locations')
+        response = requests.get(url, params={'q': text}).json()
+        geonameids = [field['geoname']['geonameid'] for field in response['result'] if 'geoname' in field.keys()]
```
jace: This is sufficient: `if 'geoname' in field` (no need for `.keys()`)

jace:
This is good to merge after the above two are fixed.

PR #118, opened 2015-04-23
trello sync

shreyas-satish:
Synchronizes the status of proposals for a given proposal space, with the specified trello board. The script `trello_sync.py` is intended to be run as a cron job.


shreyas-satish:
Will revisit this, once labels are fixed.

shreyas-satish:
We have to get rid of PROPOSALSTATUS and use a new model for labels based on a proposal_space (with a public flag). Fortunately, we just have to change two lines of code in `trello_sync.py`, to reflect this change.

PR #123, opened 2015-05-11
Participants patch 1

shreyas-satish:
- Synchronization done through the API, instead of CSV.


funnel/explara.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class ExplaraAPI(object):
+    # explara_api = ExplaraAPI({'access_token'}: app.config.get('EXPLARA_ACCESS_TOKEN'))
+    # tickets = explara_api.get_tickets(app.config.get('EXPLARA_EVENT_IDS').get('rootconf-2015'))
+    def __init__(self, config):
+        self.access_token = config.get('access_token')
+
+    def get_orders(self, explara_event_id):
+        headers = {'Authorization': 'Bearer {0}'.format(self.access_token)}
```
jace: `''.format` doesn't automatically cast plain strings into Unicode strings if any of the parameters are Unicode, so this can bite. You'll need to start with a Unicode string. Since it's just string concatenation, use this instead: `u'Bearer ' + self.access_token`

funnel/explara.py
```diff
@@ -0,0 +1,58 @@
...
+        while not completed:
+            payload = {'eventId': explara_event_id, 'fromRecord': from_record, 'toRecord': to_record}
+            attendee_response = requests.post(attendee_list_url, headers=headers, data=payload).json()
+            if not attendee_response.get('attendee'):
+                completed = True
+            elif isinstance(attendee_response.get('attendee'), list):
+                ticket_orders.append([order for order in attendee_response.get('attendee')])
+            # after the first batch, subsequent batches are dicts with batch no. as key.
+            elif isinstance(attendee_response.get('attendee'), dict):
+                ticket_orders.append([order for id, order in attendee_response.get('attendee').iteritems()])
```
jace: `id` is a builtin function and this line reassigns `id` to the last id in the list. Not good. Use `k, v` or some other variable name. Also, just `.items()` should do. `iteritems()` is deprecated in Python 3 and its behaviour is now the default.

funnel/explara.py
```diff
@@ -0,0 +1,58 @@
...
+            from_record = to_record
+            to_record += 50
+
+        return [order for order_list in ticket_orders for order in order_list]
```
jace: You wouldn't need to unpack this if the above two lines used `ticket_orders.extend` instead of `.append`.

funnel/models/event.py
```diff
@@ -2,10 +2,12 @@
...
 from .space import ProposalSpace
 from .user import User
-import random
-import uuid
 from sqlalchemy.ext.associationproxy import association_proxy
 from datetime import datetime
+from ..util import get_rows_from_csv, format_twitter
+import logging
+import base64
+import os
```
jace: As a convention, imports should be in the order of core Python libraries, extended Python libraries, third party libraries, related project libraries and internal imports. That means `os` should be the very first import, followed by `base64`, `logging`, `datetime`, `sqlalchemy` and then the project imports.

funnel/models/event.py
```diff
@@ -52,6 +53,25 @@ class Event(BaseMixin, db.Model):
     ticket_types = db.relationship("TicketType", secondary=event_ticket_type, lazy='dynamic')
     participants = association_proxy('attendees', 'participant')
 
+    @classmethod
+    def get_or_create(cls, name, space):
```
jace: Change to `def get(cls, name, space, create=False):`

funnel/models/event.py
```diff
@@ -52,6 +53,25 @@ class Event(BaseMixin, db.Model):
...
+        event = cls.query.filter_by(name=name, proposal_space=space).first()
+        if not event:
```
jace: Change to `if not event and create:`

Also, pass `create=True` when calling it.

funnel/models/event.py
```diff
@@ -52,6 +53,25 @@ class Event(BaseMixin, db.Model):
...
+            event = cls(name=name, proposal_space=space)
+            db.session.add(event)
+            db.session.commit()
```
jace: Never commit inside the model. Commits should only be in the view.

funnel/models/event.py
```diff
@@ -52,6 +53,25 @@ class Event(BaseMixin, db.Model):
...
+        return event
+
+    @classmethod
+    def sync_from_list(cls, event_list, space):
+        for event_dict in event_list:
+            event = cls.get_or_create(event_dict.get('name').decode('utf-8'), space)
```
jace: Don't call `.decode('utf-8')` here. Clean up data before it is passed to the model.

funnel/models/event.py
```diff
@@ -52,6 +53,25 @@ class Event(BaseMixin, db.Model):
...
+            for ticket_type_name in event_dict.get('ticket_types', []):
+                if ticket_type_name.decode('utf-8') not in [ticket_type.name for ticket_type in event.ticket_types]:
+                    event.ticket_types.append(TicketType.get_or_create(ticket_type_name.decode('utf-8'), space))
+        db.session.commit()
```
jace: No commit here as well.

funnel/models/event.py
```diff
@@ -52,6 +53,25 @@ class Event(BaseMixin, db.Model):
...
+        return True
```
jace: If it always returns `True`, returning any value is pointless. Skip the `return` statement.

funnel/models/event.py
```diff
@@ -65,10 +85,25 @@ class TicketType(BaseMixin, db.Model):
         backref=db.backref('ticket_types', cascade='all, delete-orphan', lazy='dynamic'))
     events = db.relationship("Event", secondary=event_ticket_type)
 
+    @classmethod
+    def get_or_create(cls, name, space):
```
jace: Rename to `get` and add parameter `create=False`.

funnel/models/event.py
```diff
@@ -112,6 +146,43 @@ def attendees_by_event(cls, event):
         return db.session.execute(stmt).fetchall()
     __table_args__ = (db.UniqueConstraint("email", "proposal_space_id"), {})
 
+    @classmethod
+    def make_from_dict(cls, participant_dict, space):
+        return Participant(
+            fullname=participant_dict.get('fullname'),
+            email=participant_dict.get('email'),
+            phone=participant_dict.get('phone'),
+            twitter=format_twitter(participant_dict.get('twitter')),
```
jace: Move `format_twitter` out, clean up data before sending it here.

funnel/models/event.py
```diff
@@ -112,6 +146,43 @@ def attendees_by_event(cls, event):
...
+
+        for r in crew_rows:
+            p = space.participants.filter(Participant.email == r[1]).first()
+            if not p:
+                p = Participant(fullname=r[0], email=r[1], company=r[2], twitter=r[3], proposal_space=space)
+                db.session.add(p)
+                db.session.commit()
+            crew_participants.append(p)
+
+        ticket_type = space.ticket_types.filter(TicketType.name == 'Crew').first()
```
jace: You shouldn't be harcoding a `Crew` string this deep into the model. Move it to a higher level.

funnel/models/event.py
```diff
@@ -144,8 +215,59 @@ class SyncTicket(BaseMixin, db.Model):
...
+        logging.warn("Sweeping cancelled tickets..")
+        for ct in cancelled_tickets:
+            logging.warn("Removing event access for {0}".format(ct.participant.email))
+            st = SyncTicket.fetch(ct.ticket_no, ct.order_no, space)
+            event_ids = [event.id for event in st.ticket_type.events]
+            cancelled_attendees = Attendee.query.filter(Attendee.participant_id == st.participant.id).filter(Attendee.event_id.in_(event_ids)).all()
+            for ca in cancelled_attendees:
+                db.session.delete(ca)
+            db.session.delete(st)
+            db.session.commit()
```
jace: Remove all the commits from this file.

funnel/models/label.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
```
jace: Is this file copy-pasted from the master branch? It's showing up as a new file here.

funnel/util.py
```diff
@@ -0,0 +1,36 @@
+from . import app
+import requests
+from urlparse import urljoin
+from baseframe import cache
+import csv
+from urlparse import urlparse
```
jace: Reorder imports.

funnel/views/event.py
```diff
@@ -108,14 +109,26 @@ def participant(profile, space):
...
     ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
     permission='event-view')
 def events(profile, space):
-    return render_template('events.html', profile=profile, space=space, events=space.events.all())
+    attendee_sync_form = forms.Form()
+    if attendee_sync_form.validate_on_submit():
+        space_key = space.title.lower() + '-' + space.name
+        if space_key in app.config.get('EXPLARA_EVENT_IDS').keys():
+            ea = ExplaraAPI({'access_token': app.config.get('EXPLARA_ACCESS_TOKEN')})
+            tickets = ea.get_tickets(app.config.get('EXPLARA_EVENT_IDS').get(space_key))
```
jace: Move `EXPLARA_EVENT_IDS` and `EXPLARA_ACCESS_TOKEN` out of config and into the models.

funnel/views/helpers.py
```diff
@@ -50,7 +48,7 @@ def file_contents(path):
 
 def make_qrcode(data, path):
     """ Makes a QR code with a given path and returns the raw svg
-      Data Format is id:key. Eg: 1:xxxxxxxx
+        Data Format is id:key. Eg: 1:xxxxxxxx
```
jace: Docstring formatting convention is one of:

```python
def foo():
    """This is a single line docstring."""

def bar():
    """
    This is a multi-
    line docstring
    """
```

PR #125, opened 2015-06-16
Schedule fixes

shreyas-satish:
- The slot size has been set to 5 minutes in the 'edit' view and 15 minutes in the read-only view.
- The date range for the calendar now accommodates changes in a proposal space's dates.

### DIFF
--- a/funnel/views/schedule.py
+++ b/funnel/views/schedule.py
@@ -247,8 +247,13 @@ def schedule_edit(profile, space):
             } for proposal in space.proposals_all if proposal.confirmed and not proposal.session],
         'scheduled': session_data(space.sessions, timezone=space.timezone, with_modal_url='edit', with_delete_url=True)
         }
+    # Set the proper range for the calendar to allow for date changes
+    first_session_date = Session.query.filter_by(proposal_space=space).order_by("created_at asc").first().start
+    last_session_date = Session.query.filter_by(proposal_space=space).order_by("created_at desc").first().start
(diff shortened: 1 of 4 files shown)


funnel/views/schedule.py
```diff
@@ -247,8 +247,13 @@ def schedule_edit(profile, space):
             } for proposal in space.proposals_all if proposal.confirmed and not proposal.session],
         'scheduled': session_data(space.sessions, timezone=space.timezone, with_modal_url='edit', with_delete_url=True)
         }
+    # Set the proper range for the calendar to allow for date changes
+    first_session_date = Session.query.filter_by(proposal_space=space).order_by("created_at asc").first().start
+    last_session_date = Session.query.filter_by(proposal_space=space).order_by("created_at desc").first().start
+    from_date = first_session_date if first_session_date.date() < space.date else space.date
+    to_date = last_session_date if last_session_date.date() > space.date_upto else space.date_upto
```
jace: How does this work if the dates are several weeks/months apart? Do you get navigation and the ability to drag across weeks?

shreyas-satish: Interesting use case; hasn't been covered here. One option is to provide a 'bulk edit' option that remaps the dates of the sessions to the new dates. Also, dragging multiple sessions across weeks or months can be cumbersome.

jace:
Merging since we need this urgently for Fifthel, but the highlighted concern needs to be examined.

PR #128, opened 2015-08-25
catch IntegrityError and inform user with appropriate error message

shreyas-satish:


jace:
This is the wrong place to catch it. The form should validate if the name is acceptable.

PR #130, opened 2015-08-25
Add order

shreyas-satish:
- Added the `Order` model, and associates it with `SyncTicket`.
- Updated 'ExplaraApi` to format orders.


shreyas-satish:
Will be made as a PR for `master`.

PR #132, opened 2015-09-13
adds ability to import a csv of participants

shreyas-satish:


funnel/forms/space.py
```diff
@@ -92,6 +92,16 @@ class ParticipantForm(forms.Form):
         validators=[forms.validators.DataRequired(u"Select at least one event")])
 
 
+class ParticipantImportForm(forms.Form):
+    participant_list = forms.FileField(__("Participant List"),
```
jace: Should be in sentence case.

funnel/forms/space.py
```diff
@@ -92,6 +92,16 @@ class ParticipantForm(forms.Form):
...
+        description=u"CSV with Header. Order: name, email, phone, twitter, company.",
```
jace: Why do you need headers if you need a specific order? Pick one. See Hasmail for how it's handled there if you need ideas.

funnel/models/event.py
```diff
@@ -23,6 +24,14 @@ def make_private_key():
     return make_key()[:8]
 
 
+def csv_to_rows(csv_file, skip_header=True, delimiter=','):
+    with open(csv_file, 'rb') as csvfile:
```
jace: Doesn't FileField give you an open CSV file?

funnel/models/event.py
```diff
@@ -23,6 +24,14 @@ def make_private_key():
...
+        reader = csv.reader(csvfile, delimiter=delimiter)
```
jace: Use `unicodecsv`.

funnel/models/event.py
```diff
@@ -23,6 +24,14 @@ def make_private_key():
...
+        if skip_header:
+            next(reader)
```
jace: Use `DictReader` and you won't need this kludge.

funnel/models/event.py
```diff
@@ -125,6 +134,24 @@ class Participant(BaseMixin, db.Model):
     __table_args__ = (db.UniqueConstraint('proposal_space_id', 'email'),)
 
     @classmethod
+    def get(cls, space, email, participant_attrs={}, create=False):
```
jace: `participant_attrs` is a mouthful. Call this `fields` or something light like that.

funnel/models/event.py
```diff
@@ -125,6 +134,24 @@ class Participant(BaseMixin, db.Model):
...
+        participant = cls.query.filter_by(proposal_space=space, email=email).first()
```
jace: This assumes `email` is unique. How does this work with platforms like Explara and DoAttend that don't enforce a unique constraint on email?

funnel/models/event.py
```diff
@@ -136,21 +163,8 @@ def attendees_by_event(cls, event):
...
-            job_title=participant_dict.get('job_title'),
-            company=participant_dict.get('company'),
-            city=participant_dict.get('city'),
-            proposal_space=space
-        )
-
-    @classmethod
-    def update_from_dict(cls, updated_participant):
-        participant = Participant.query.filter_by(email=updated_participant.get('email')).first()
+    def update_from_dict(cls, space, updated_participant):
```
jace: `updated_participant` is ambiguous. If this is a dict of fields, call it something consistent like `fields`.

funnel/models/event.py
```diff
@@ -136,21 +163,8 @@ def attendees_by_event(cls, event):
...
+        participant = cls.query.get(space, updated_participant.get('email'))
```
jace: Did you mean `cls.get`? `cls.query.get` uses SQLAlchemy's primary key-based getter which won't work with objects like `space`.

funnel/models/event.py
```diff
@@ -163,6 +177,15 @@ def update_from_dict(cls, updated_participant):
         else:
             return None
```
jace: Silent failure is a bad idea. Also, the `db.session.add` above is a bad idea since no new participant is being created in this code block. It'll cause hard to debug problems elsewhere later.

funnel/models/event.py
```diff
@@ -163,6 +177,15 @@ def update_from_dict(cls, updated_participant):
...
 
+    @classmethod
+    def add_from_list(cls, space, filename="", events=[]):
+        """
+        Expected CSV format (with header) : name, email, phone, twitter, company
+        """
+        for row in csv_to_rows(filename):
+            participant_dict = {'events': events, 'fullname': row[0], 'email': row[1], 'phone': row[2], 'twitter': row[3], 'company': row[4]}
+            cls.get(space, participant_dict.get('email'), participant_dict, create=True)
+
```
jace: This entire method is making me uncomfortable:

1. Dealing with writeable files on what will be a read-only filesystem in future deployments
2. Hardcoding column order

funnel/models/event.py
```diff
@@ -176,6 +199,14 @@ class Attendee(BaseMixin, db.Model):
         backref=db.backref('attendees', cascade='all, delete-orphan', lazy='dynamic'))
     checked_in = db.Column(db.Boolean, default=False, nullable=False)
 
+    @classmethod
+    def get(cls, participant=None, event=None, create=False):
```
jace: participant` and `event` shouldn't have default values in this method. Both are required. `event` is the more significant parameter (for SQL left-to-right indexing order) and should be first.

funnel/models/event.py
```diff
@@ -125,6 +134,24 @@ class Participant(BaseMixin, db.Model):
     __table_args__ = (db.UniqueConstraint('proposal_space_id', 'email'),)
 
     @classmethod
+    def get(cls, space, email, participant_attrs={}, create=False):
+        participant = cls.query.filter_by(proposal_space=space, email=email).first()
+        if create and not participant:
+            participant = cls(
```
jace: What happens if you call with `create=True` and the record already exists? `participant_attrs` is ignored then, even if the data doesn't match? That makes this method's purpose and behaviour confusing.

funnel/models/event.py
```diff
@@ -163,6 +177,15 @@ def update_from_dict(cls, updated_participant):
...
+        """
+        Expected CSV format (with header) : name, email, phone, twitter, company
+        """
+        for row in csv_to_rows(filename):
+            participant_dict = {'events': events, 'fullname': row[0], 'email': row[1], 'phone': row[2], 'twitter': row[3], 'company': row[4]}
+            cls.get(space, participant_dict.get('email'), participant_dict, create=True)
+
 
 class Attendee(BaseMixin, db.Model):
     """Join model between Participant and Event."""
```
jace: Join models shouldn't inherit from BaseMixin, but that's for another pull request review.

funnel/templates/events.html
```diff
@@ -4,14 +4,18 @@
...
 <form method="POST" id="attendee_remote_sync_form" enctype="multipart/form-data" class="form-inline">
   <div class="form-group">
     {{ attendee_sync_form.hidden_tag() }}
   </div>
   <input id="attendee_sync_form-submit" class="btn btn-danger" type="submit" value="{% trans %}Sync Attendees{% endtrans %}"/>
 </form>
-</p>
+</li>
+<li><a href={{ space.url_for('new-participant') }} class='btn btn-primary'>New Participant</a></li>
+<li><a href={{ space.url_for('import-participant') }} class='btn btn-info'>Import Participants</a></li>
```
jace: Labels should be sentence case and wrapped in a `{% trans %}` block.

funnel/views/event.py
```diff
@@ -37,6 +38,10 @@ def participant_data(participant, space_id, full=False):
         }
 
 
+def allowed_file(filename, exts=[]):
+    return '.' in filename and filename.rsplit('.', 1)[1] in exts
```
jace: This method shouldn't be required.

funnel/views/event.py
```diff
@@ -69,6 +74,29 @@ def new_participant(profile, space):
...
+    (Profile, {'name': 'profile'}, 'g.profile'),
+    ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
+    permission='new-participant')
+def import_participant(profile, space):
+    form = ParticipantImportForm()
+    form.events.query = space.events
+    if form.validate_on_submit():
+        participant_list_csv = request.files['participant_list']
+        if participant_list_csv and allowed_file(participant_list_csv.filename, exts=['csv']):
+            filename = secure_filename(participant_list_csv.filename)
```
jace: Don't write to the filesystem. Work with the CSV in memory.

funnel/views/event.py
```diff
@@ -69,6 +74,29 @@ def new_participant(profile, space):
...
+            participant_list_csv.save(os.path.join(app.config['UPLOAD_FOLDER'], filename))
+            Participant.add_from_list(space, filename=filename, events=form.data['events'])
+            db.session.commit()
+            flash(_(u"Participants were imported from {0}.".format(filename)), 'info')
```
jace: Avoid ordered parameters in `format`. Use named parameters.

funnel/views/event.py
```diff
@@ -69,6 +74,29 @@ def new_participant(profile, space):
...
+            return redirect(space.url_for('events'), code=303)
+        else:
+            form.participant_list.errors.append(u'Please provide a valid CSV file.')
```
jace: Move this validation into the form.

funnel/models/event.py
shreyas-satish: The combination of (space, email) is enforced to be unique. If there are multiple entries in Explara with the same email for a given space, then it's taken to mean that, the particular participant has multiple tickets.

funnel/models/event.py
shreyas-satish: One option is to add an `update` flag, so if a record is already present (and if `update` is set), the record will be updated with the dictionary that was passed.

funnel/forms/space.py
```diff
@@ -92,6 +93,30 @@ class ParticipantForm(forms.Form):
...
+# TODO: Move to Baseframe
+class ValidFile(object):
+    def __init__(self, allowed_exts=[], message=None):
+        if not message:
+            message = __(u"Please upload a valid file.")
+        self.message = message
+        self.allowed_exts = allowed_exts
+
+    def __call__(self, form, field):
+        ext = os.path.splitext(field.data.filename)[1].strip(".")
```
jace: Have you tried this on Windows with IE? Filename is as reported by the browser based on the native OS, while `os.path` uses the conventions of the host OS. You will also get an internal server error if someone uploads a file without an extension.

jace: Corrected, `splitext` won't fail because it always returns a value in position `[1]`.

funnel/forms/space.py
```diff
@@ -92,6 +93,30 @@ class ParticipantForm(forms.Form):
...
+        if ext and ext.lower() not in self.allowed_exts:
+            raise wtforms.validators.StopValidation(self.message)
```
jace: Any file passed through MS Excel on Mac will have CR for linefeeds, not LF or CRLF, and will be in Latin1 charset instead of UTF-8. You need to detect that and massage the data.

jace: (Hasmail chokes on this, BTW.)

funnel/forms/space.py
```diff
@@ -92,6 +93,30 @@ class ParticipantForm(forms.Form):
...
+
+
+class ParticipantImportForm(forms.Form):
+    participant_list = forms.FileField(__("Participant list"),
+        description=u"Expected Attributes: name, email, phone, twitter, company.",
```
jace: "Expected headers", and no trailing period in the description.

jace: Also, if the source doesn't have a twitter header, import shouldn't fail. Minimum requirements are only name and email.

funnel/forms/space.py
```diff
@@ -92,6 +93,30 @@ class ParticipantForm(forms.Form):
...
+        validators=[forms.validators.DataRequired(u"Please upload a valid CSV file."), ValidFile(allowed_exts=['csv'], message=u"Please upload a valid CSV file.")])
```
jace: No trailing periods.

funnel/models/event.py
```diff
@@ -125,6 +126,24 @@ class Participant(BaseMixin, db.Model):
     __table_args__ = (db.UniqueConstraint('proposal_space_id', 'email'),)
 
     @classmethod
+    def get(cls, space, email, fields={}, create=False):
+        participant = cls.query.filter_by(proposal_space=space, email=email).first()
+        if create and not participant:
+            participant = cls(
+                fullname=fields.get('fullname'),
```
jace: Says `fullname` here but `name` in the CSV description above.

funnel/models/event.py
```diff
@@ -125,6 +126,24 @@ class Participant(BaseMixin, db.Model):
...
+                email=email,
+                phone=fields.get('phone'),
+                twitter=fields.get('twitter'),
+                job_title=fields.get('job_title'),
```
jace: `job_title` not in the CSV header list above.

funnel/models/event.py
```diff
@@ -125,6 +126,24 @@ class Participant(BaseMixin, db.Model):
...
+                company=fields.get('company'),
+                city=fields.get('city'),
+                events=fields.get('events', []),
```
jace: IMO, all setters should be moved out of the `get` method. Return an empty participant object and let the caller set/update the fields.

funnel/models/event.py
```diff
@@ -136,32 +155,18 @@ def attendees_by_event(cls, event):
...
+            participant.fullname = fields.get('fullname')
+            participant.phone = fields.get('phone')
+            participant.twitter = fields.get('twitter')
+            participant.job_title = fields.get('job_title')
+            participant.company = fields.get('company')
+            participant.city = fields.get('city')
             return participant
         else:
-            return None
+            raise NoResultFound
```
jace: 1. `NoResultFound` implies you were searching. `update` in the method name doesn't sound like searching.
2. No good reason for this to be a class method. Let `get(create=True)` return an empty participant and `update_from_dict` be an instance method that applies on the current instance.

funnel/models/event.py
```diff
@@ -221,12 +234,7 @@ def sync_from_list(cls, ticket_list, space, ticket_client=None):
         for ticket_dict in ticket_list:
             ticket = SyncTicket.fetch(ticket_dict.get('ticket_no'), ticket_dict.get('order_no'), space)
             ticket_ticket_type = TicketType.get(ticket_dict.get('ticket_type'), space, create=True)
-            # get or create participant
-            ticket_participant = Participant.query.filter_by(email=ticket_dict.get('email'), proposal_space=space).first()
-            if not ticket_participant:
-                # create a new participant record if required
-                ticket_participant = Participant.make_from_dict(ticket_dict, space)
-                db.session.add(ticket_participant)
+            ticket_participant = Participant.get(space, ticket_dict.get('email'), fields=ticket_dict, create=True)
```
jace: `ticket_dict.get('email')` isn't normalising for case. Email is case-insensitive but the database is case-sensitive. You have to force lowercase it.

funnel/views/event.py
```diff
@@ -7,12 +8,20 @@
...
 from sqlalchemy.exc import IntegrityError
 from ..extapi.explara import ExplaraAPI
 
 
+def lowercase_keys(dictiter):
+    return [dict((k.lower(), v) for k, v in d.iteritems()) for d in dictiter]
+
+
+def csvstring_to_rows(stream):
+    return lowercase_keys(unicodecsv.DictReader([row for row in StringIO.StringIO(stream)]))
```
jace: This is a bit convoluted. You've already got a stream. Why are you putting it through StringIO again?

funnel/forms/space.py
shreyas-satish: import won't fail in that case. I'll add `(optional)` for optional attributes.

shreyas-satish:
The conversation here triggered changes that extend far beyond the initially mentioned change. Will incorporate changes suggested here into a new PR.

PR #133, opened 2015-09-15
fixed case of receiving multiple geonameids

shreyas-satish:


funnel/util.py
```diff
@@ -7,13 +7,13 @@
...
+        in the string, and returns an array of matched geonameids.
+        Eg: "Bangalore" -> [1277333]
         To detect multiple locations, split them up and pass each location individually
     """
     if 'HASCORE_SERVER' in app.config:
         url = urljoin(app.config['HASCORE_SERVER'], '/1/geo/parse_locations')
         response = requests.get(url, params={'q': text}).json()
         geonameids = [field['geoname']['geonameid'] for field in response['result'] if 'geoname' in field]
-        return geonameids[0] if geonameids else None
+        return geonameids
```
jace: Return a `set` here.

funnel/models/proposal.py
```diff
@@ -189,8 +189,10 @@ def confirmed(self):
         return self.status == PROPOSALSTATUS.CONFIRMED
 
     @cached_property
-    def location_geonameid(self):
-        return geonameid_from_location(self.location)
+    def has_outstation_speaker(self):
+        space_geonameids = set(self.proposal_space.location_geonameid)
+        proposal_geonameids = set(geonameid_from_location(self.location))
```
jace: You won't need to make sets if they are already sets.

PR #134, opened 2015-09-16
Offline support for check in using browser local storage.

vidya-ram:


vidya-ram:
Wrong pull request. Merge to participants not master.

PR #135, opened 2015-09-16
  Offline support for check in using browser local storage.

vidya-ram:


funnel/templates/event.html
```diff
@@ -62,9 +62,9 @@
             </li>
             <li>
             {%- if p.checked_in %}
-              <a href="{{ event.name }}/checkin/{{ p.id }}" class='btn btn-xs btn-danger'>Cancel Check-in</a>
+              <a href="{{ event.name }}/checkin/{{ p.id }}" data-pid="{{ p.id }}" data-tocheckin=false class='btn btn-xs cancel-checkin'>Cancel Check-in</a><img src="{{ url_for('static', filename='img/ajax-loader.gif')}}" class="loader"/>
```
jace: In `data-tocheckin=false`, attribute values should always be quoted as unquoted values are an attack vector. So `data-tocheckin="false"`.

funnel/templates/event.html
```diff
@@ -76,6 +76,37 @@
...
+    }
+    Storage.prototype.remove = function(key) {
+      var participantList = JSON.parse(this.getItem(LocalStorageKey));
+      if(participantList) {
+        delete participantList[key];
+      }
+      return this.setItem(LocalStorageKey, JSON.stringify(participantList))
+    }
+  }
+
```
jace: You need an else condition here to warn the user this page won't work for them.

funnel/templates/event.html
```diff
@@ -161,6 +192,58 @@
...
 
+    if (Modernizr.localstorage) {
+      $('.checkin, .cancel-checkin').on('click', function(event) {
+        event.preventDefault();
+        localStorage.set($(this).data('pid'), $(this).attr('href'));
+        $(this).next('.loader').show();
+      });
+
+      // On page reload, for the all the participants in localStorage['p-id'] array show the ajax-loader image
+      if(localStorage.getItems() && Object.keys(localStorage.getItems()).length > 0) {
```
jace: This line runs `JSON.parse` twice. Can that be avoided?

funnel/views/event.py
```diff
@@ -209,4 +209,6 @@ def event_checkin(profile, space, event, participant):
     a.checked_in = checked_in
     db.session.add(a)
     db.session.commit()
+    if request.method == 'POST':
```
jace: Use `if request.xhr` here. Using jQuery's `$.ajax` sets the appropriate HTTP header and Flask will automatically set `request.xhr` to `True`.

funnel/templates/event.html
vidya-ram: Incase of no localStorage, it falls back to earlier method of reloading the page after updating DB

funnel/views/event.py
karthikb351: that is fine for ajax, but Android's HTTP calls don't have the XHR header. I'll have to add it manually, unless we define a different endpoint for android (which doesn't make sense).

jace: There's actually two different things happening here.

1. `request.xhr` implies (only) that the request came from JS and so a fragment response may be appropriate. It does not indicate whether the response should be a HTML fragment or JSON.
2. The HTTP `Accepts` header is a list of content types that the client is expecting. Browsers send a full list starting with 'text/html' and ending with '*/*'. You can indicate that you'd like a JSON response by sending 'application/json'.

Coaster provides a [`render_with` decorator](https://github.com/hasgeek/coaster/blob/master/coaster/views.py#L388) that will make this easier to write. I'll recommend using that.

@karthikb351 for Android, since client installations have life in the wild that is out of our control, all APIs called from the Android app should be version controlled and maintained independently of the web APIs. We'll need to track usage of those APIs independently to know when it is safe to retire them.

vidya-ram:
@shreyas-satish Please review

vidya-ram:
Earlier we used to reload the page for every check-in, so this ensured multiple check-in counters have the right check-in status for participants.
With this offline check-in feature, we would either have to do a manual page reload or poll to the server to get all the participants check-in status.
Instead of polling to server to get a list of (500 or more) participants check-in status and going through the list and updating the UI, can we do a check for network and do a forced reload (Location.reload()) of page at some fixed interval.

PR #139, opened 2015-10-08
Offline support for check in using browser local storage

vidya-ram:
`getParticipantsData` gets participants data as json from server and uses ractive to display it on the UI

LocalStorage stores only strings.
`addToLocalstorage` stringifies the (check-in / cancel check-in) array and stores in localStorage.
`readFromLocalstorage` reads from localStorage and parses it.

check-in and cancel check-in queue is implemented using arrays.
`enqueue` - adds participant to queue
`peek` - reads participant from queue
`peekAllItems` - reads all participant from queue
`dequeue` - removes participant from queue

If localStorage is available on the browser, event handlers for check-in, cancel-checkin and abort are registered.
`updateQueue` - checks all participants in the localStorage queue. If their status have been updated then removes them from the queue.
`checkQueues` is called every 8 seconds which reads the first item from queue(first from check-in queue and then cancel check-in queue) and calls `postCheckinStatus` which does a ajax post to server and removes from queue on successful post.

Workflow:
1) When a participant is checked-in, the participant id is added to a checkin queue which is stored in localStorage. Loading and abort icons are displayed on UI for this participant.
2) Every 8 seconds first item from check-in queue is read and check-in ajax post request is send to server. The participant id is removed from queue on successful post. Same is done for cancel check-in queue.
3) Every one minute participants data is requested from the server and UI is refreshed by ractive. Participant ids in the queue(stored in localStorage) are checked against their check-in status on the UI, if their status have already been updated (in case of multiple check-in counters) then their are removed from the queue. All participants remaining in the queue will have the loading and abort icons.
4) On aborting a check-in, if the participant id is present in check-in queue then it is removed from the queue. If participant id is not present in the check-in queue it implies check-in has already been posted to the server, so a cancel check-in requires to be send for which the participant is added it to cancel check-in queue.

### DIFF
--- a/funnel/templates/event.html
+++ b/funnel/templates/event.html
@@ -131,8 +131,8 @@
   }
 
   $(document).ready(function() {
-    var checkinUrl = window.location.href + "/checkin/";
-    var getParticipantsUrl = window.location.href + "/participants/json";
+    var checkinUrl = "{{ '{0}event/{1}/checkin/'.format(space.url_for(), event.name) }}";
+    var getParticipantsUrl = "{{ '{0}event/{1}/participants/json'.format(space.url_for(), event.name) }}";
 
     // To disable debug mode
     Ractive.DEBUG = false;
--- a/funnel/templates/event.html
+++ b/funnel/templates/event.html
@@ -90,13 +90,13 @@
 
     //Local storage can only save strings, so value is converted into strings and stored.
     window.Talkfunnel.Store.addToLocalstorage = function(key, value) {
-      return localStorage.setItem(key, JSON.stringify(value))
-    }
+      return localStorage.setItem(key, JSON.stringify(value));
+    };
 
     //Reads from LocalStorage.
     window.Talkfunnel.Store.readFromLocalstorage = function(key) {
       return JSON.parse(localStorage.getItem(key));
-    }
+    };
 
     window.Talkfunnel.Queue = {};
 
@@ -105,9 +105,9 @@
       var participantList = window.Talkfunnel.Store.readFromLocalstorage(queueName) || [];
       if(participantList.indexOf(participant_id) === -1) {
         participantList.push(participant_id);
-        return window.Talkfunnel.Store.addToLocalstorage(queueName, participantList)
+        return window.Talkfunnel.Store.addToLocalstorage(queueName, participantList);
       }
-    }
+    };
 
     //Reads and returns all items from queue
     //Returns undefined when queue is empty or not defined
@@ -116,7 +116,7 @@
       if(participantList && participantList.length) {
         return participantList;
       }
-    }
+    };
 
     //Removes item from queue and returns true
     //Returns undefined when item not present in queue
@@ -129,7 +129,7 @@
         window.Talkfunnel.Store.addToLocalstorage(queueName, participantList);
         return true;
       }
-    }
+    };
   }
 
   $(document).ready(function() {
@@ -149,6 +149,7 @@
       var participantIDs = window.Talkfunnel.Queue.peekAll(queueName);
       if(participantIDs) {
         participantIDs.forEach(function(participantID) {
+          var participant = "[data-pid=" + participantID +"]";
           if(queueName === "checkin-queue") {
             if(participants[participantID].checked_in) {
               //Participant has been checked-in so remove from 'checkin-queue' 
@@ -272,7 +273,7 @@
         if(!navigator.onLine) {
           return 'There is no network!';
         }
-      }
+      };
 
       //Read 'checkin-queue' and 'cancelcheckin-queue' every 8 second and post check-in status to server
       setInterval(processQueues, 8000);
@@ -285,7 +286,7 @@
        if ($('#badge_form select.field-badge_printed').val() === "") {
         return false;
        }
-       return window.confirm("Are you sure? Selected action will apply to all listed participants.")
+       return window.confirm("Are you sure? Selected action will apply to all listed participants.");
     });
 
     var TableSearch = function(tableId){
@@ -301,7 +302,7 @@
       this.tableId = tableId;
       this.rowData = [];
       this.allMatchedIds = [];
-    }
+    };
 
     TableSearch.prototype.getRows = function(){
       return $('#' + this.tableId +' tbody tr');
@@ -321,7 +322,7 @@
 
     TableSearch.prototype.setAllMatchedIds = function(ids) {
       this.allMatchedIds = ids;
-    }
+    };
 
     TableSearch.prototype.searchRows = function(q){
       // Search the rows of the table for a supplied query.
@@ -329,26 +330,26 @@
       // reset data collection on first search or if table has changed
       if (this.rowData.length !== this.getRows().length) {
         this.setRowData();
-      };
+      }
 
       // return cached matched ids if query is blank
       if (q === '' && this.allMatchedIds.length !== 0) {
         return this.allMatchedIds;
-      };
+      }
 
       var matchedIds = [];
       for (var i = this.rowData.length - 1; i >= 0; i--) {
         if (this.rowData[i].text.indexOf(q.toLowerCase()) !== -1) {
           matchedIds.push(this.rowData[i]['rid']);
         }
-      };
+      }
 
       // cache ids if query is blank
       if (q === '') {
         this.setAllMatchedIds(matchedIds);
-      };
+      }
       return matchedIds;
-    }
+    };
 
     var tableSearch = new TableSearch('participants-table');
 
--- a/funnel/templates/event.html
+++ b/funnel/templates/event.html
@@ -46,37 +46,223 @@
         <th>Actions</th>
       </tr>
     </thead>
-    <tbody>
-      {%- for p in participants %}
-        <tr id="p-{{ p.id }}">
-          <td class='js-searchable'>{{ p.fullname }}</td>
-          <td class='js-searchable'>{{ p.email }}</td>
-          <td class='js-searchable'>{{ p.company }}</td>
+    <tbody id="participants">
+      {% raw %}
+      <script id='participant' type='text/ractive'>
+      {{#each participants}}
+        <tr id="p-{{ pid }}">
+          <td class='js-searchable'>{{ fullname }}</td>
+          <td class='js-searchable'>{{ email }}</td>
+          <td class='js-searchable'>{{ company }}</td>
           <td>
             <ul class="list-inline status-btn-list">
             <li>
-              <a href={{ "{0}participant/{1}/badge".format(space.url_for(), p.id) }} class={% if p.badge_printed %} 'btn btn-xs btn-success'  {% else %} 'btn btn-xs btn-info' {% endif %} target="_blank">Show badge</a>
+              <a href="{{ showbadge_url }}" {{#badge_printed}} class="btn btn-xs btn-success" {{/badge_printed}} {{^badge_printed}} class="btn btn-xs btn-info" {{/badge_printed}} target="_blank">Show badge</a>
             </li>
             <li>
-              <a href={{ "{0}participant/{1}/edit".format(space.url_for(), p.id) }} class='btn btn-xs btn-default'>Edit</a>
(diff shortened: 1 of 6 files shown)


funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+      {{/each}}
+      </script>
+      {% endraw %}
     </tbody>
   </table>
 {% endblock %}
 {% block footerscripts %}
 <script>
+  if (Modernizr.localstorage) {
+    window.Store = {};
```
shreyas-satish: Why is this declared on the `window` object? This could just be a local var?

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+
+    //Reads from LocalStorage (key is 'checkin-queue' or 'cancelcheckin-queue')
```
shreyas-satish: `checkin-queue` or `cancelcheckin-queue` are implementation details specific to this particular use case. There are no checks for these specific keys in the code, so these comments make better sense in the caller's section.

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+    window.Store.readFromLocalstorage = function(key) {
```
shreyas-satish: Maybe add a comment on what the purpose of `Store` is? Is it intended to only wrap `localStorage`? If so, just `read` and `write` would suffice as method names?

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+      return JSON.parse(localStorage.getItem(key));
+    }
+    //Adds to LocalStorage (key is 'checkin-queue' or 'cancelcheckin-queue')
+    window.Store.addToLocalstorage = function(key, value) {
+      return localStorage.setItem(key, JSON.stringify(value))
+    }
+
+    window.Queue = {};
```
shreyas-satish: Same as above.

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+
+    //Adds a participant_id to queue
+    window.Queue.enqueue = function(queueName, participant_id) {
```
shreyas-satish: This isn't really a 'Queue' from a computer science perspective, since the `participant_id` is being pushed to the end of the list (queues are FIFO). This seems like a interface that offers list operations for data stored via `Store`. Maybe call it `StoreList`? Again, if I understand correctly, the the inner details in these functions have nothing to do with 'participants' per se? As in these functions can be applied in a different context with a similar problem of saving lists in stores?

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+      var participantList = window.Store.readFromLocalstorage(queueName);
+      if(participantList) {
+        participantList.push(participant_id);
+      }
+      else {
+        participantList = [];
+        participantList.push(participant_id);
+      }
```
shreyas-satish: `participantList.push(participant_id);` seems to be repeated in both `if` and `else`. The conditional isn't really required. This could just be:
```javascript
    var participantList = window.Store.readFromLocalstorage(queueName) || [];
    participantList.push(participant_id);
````

File: funnel/templates/event.html [hunk 7]
```diff
@@ -46,37 +46,251 @@
...
+      return window.Store.addToLocalstorage(queueName, participantList)
+    }
+
+    //Reads first item from queue
+    //Returns undefined when queue is empty or not defined
+    window.Queue.peek = function(queueName) {
```
shreyas-satish: This can be `first` since it returns the first item?

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+      var participantList = window.Store.readFromLocalstorage(queueName);
+      if(participantList && participantList.length) {
+        return participantList.shift();
+      }
+    }
+
+    //Reads and returns all items from queue
+    //Returns undefined when queue is empty or not defined
+    window.Queue.peekAllItems = function(queueName) {
```
shreyas-satish: This can just be `all`?

funnel/templates/event.html
vidya-ram: This does work like a Queue where elements elements are added to the rear and removed from the front.
The first participant from the queue is read, his/her check-in status is posted to the server. On successful post, the participant is removed from queue.

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+    var participantTableContent = new Ractive({
+      el: '#participants-table-content',
+      template: '#participant-row',
+      data: {
+        participants: ''
+      }
+    });
+
+    //updateQueue: If participant in "checkin-queue" has already been checked-in then it is removed from checkin queue
+    function updateQueue(queueName) {
```
shreyas-satish: Checking the DOM to decide what needs to be done next, doesn't seem like the right approach to me. Also, I'm not sure why this function is necessary. Removing checked-in ids from the queue, after sending them to the server, is currently being done in `postCheckinStatus`, right?

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+            }
+            else {
+              $(participant).siblings('.loader').show();
+            }
+          }
+        });
+      }
+    }
+
+    function getParticipantsData() {
```
shreyas-satish: This probably needs a better name, since there are multiple things happening here - getting data, setting the `Ractive` variable, and updating queues. `updateParticipantList` perhaps?

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+        if(!navigator.onLine) {
+          return 'There is no network!';
+        }
+      }
+
+      //Read 'checkin-queue' and 'cancelcheckin-queue' every 8 second
+      setInterval(checkQueues, 8000);
+
+      //Get participants data from server every 1 min
+      setInterval(getParticipantsData, 60000);
```
shreyas-satish: It's better to co-locate this with the first time the call to `getParticipantsData` is made.

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+            window.Queue.dequeue("checkin-queue", data.participant_id);
+          }
+          else {
+            window.Queue.dequeue("cancelcheckin-queue", data.participant_id);
+          }
+        }
+      });
+    }
+
+    function checkQueues() {
```
shreyas-satish: 'Checking' makes it seem like a 'read' operation, but what's actually happening here is, the check-in status of the first participant is being posted to the server. So, it looks like the queue is processed, one participant at a time, every 8 seconds? This can be faster. Just send the entire list and let the server do a bulk check-in?

funnel/templates/event.html
```diff
@@ -46,37 +46,251 @@
...
+
+    //Get the list of participants from server
+    getParticipantsData();
+
+    //If localStorage is available, register event handlers for check-in, cancel-checkin and abort
+    if (Modernizr.localstorage) {
+      $('#participants-table-content').on('click', '.js-checkin', function(event) {
+        event.preventDefault();
+        //Add to participant_id to "checkin-queue"
+        window.Queue.enqueue("checkin-queue", $(this).data('pid'));
```
shreyas-satish: So, what's happening here is, `Queue` has no knowledge about `checkin-queue`, since it's just passing it off to `localStorage`. I would recommend an object-oriented approach here. 
```javascript
   var checkinq = new Queue('checkin');
   checkinq.enqueue(item);
   checkinq.dequeue(item);
````
````

funnel/templates/event.html
vidya-ram: This function helps in avoiding sending redundant requests to server and also prevent from showing loading indicator against incorrect check-in status when UI is refreshed every 1 minute.

Example: When there are two or more check-in counters at the event and two counters does check-in for the same participant. The participant will be added to the check-in queue of both counters. Assume, the request from the first counter reaches the server first and the participant is checked-in.
When UI is refreshed in second counter's browser, this participant status will show Checked-in. Since this participant is in second counter's check-in queue too, the loading and abort indicator will be displayed which is incorrect.
So, this function will check and remove the participant from the check-in queue of second counter.

shreyas-satish: So, you could check the data the server returns, and update the queue in that case, instead of checking the DOM?

funnel/templates/event.html
vidya-ram: Store and Queue are explicitly defined as global to avoid being overwritten. At this point of code, var Store is same as window.Store.

funnel/templates/event.html
vidya-ram: data from the server is used to update the DOM. Once the DOM has been updated, only then is the queue check & update done.

funnel/templates/event.html
vidya-ram: Instance of Queue means handling another copy of check-in and cancel queue. It is straightforward to just store the queue in the local storage.
 When an item has to be added to the queue, read the queue from local storage, update it and write it back.

funnel/templates/event.html
shreyas-satish: I get that, but the DOM is a secondary source of information. For instance, there might be behavior in the future that could muck around with the DOM. Which is why relying on the DOM for data is brittle.

Instead, using the data from the server makes it easier to reason about, since the programmer does not have to worry about multiple sources of data to keep in sync.

funnel/templates/event.html
shreyas-satish: What happens when external JS (through a library or a browser extension) runs on the page and happens to use the `Queue` or `Store` globals? Globals should be avoided as far as possible, and only be used when an alternate implementation is overly complex. A good practice is to have **one** global variable across the app, and use that to scope variables that _really_ need to be global. 

For instance:
````javascript
    window.HasGeekTalkfunnel = {}
    window.HasGeekTalkfunnel.Store = {}
````
This way, the programmer has an immediate handle on all the global variables that are currently in use. See [Hasjob](https://github.com/hasgeek/hasjob/blob/master/hasjob/static/js/app.js) as an example of this pattern.

funnel/templates/event.html
shreyas-satish: > Instance of Queue means handling another copy of check-in and cancel queue.

Why is this a problem? Also, let's consider:

````javascript
    var firstItem = window.Queue.first("cancelcheckin-queue");
    window.Queue.dequeue("cancelcheckin-queue", data.participant_id);
````
versus

````javascript
   var firstItem = cancelCheckinQ.first();
   cancelCheckinQ.dequeue(data.participant_id);
````
As a reader, I find the latter approach more readable, and easier to understand. I have a clear idea about when the queue is being instantiated. Also, I stand to get hints from a [linter](http://jshint.com/) if I try to use a queue that is uninitialized or out of scope. Finally, I don't have to specify the name of localStorage key, every time I need to perform an operation. These points become especially useful when managing multiple queues.

An analogy is the data models on the back-end. For instance:
````python
    class Proposal(BaseScopedIdNameMixin, CoordinatesMixin, db.Model):
        __tablename__ = 'proposal'
       ...
````
The `tablename` is well encapsulated and I only have to worry about the `Proposal` and the instances I create, instead of having to know the name of the table for every time I would need to perform operations on the model.

funnel/templates/event.html
vidya-ram: I agree, DOM is secondary source of information. But in this case, these DOM values are updated only receiving values from server. 
Any future behavior js shouldn't be allowed to alter these DOM values unless data is received from the server. The whole check-in functionality works on information from DOM right from the adding to check-in queue or cancel check-in queue.
$('#participants-table-content').on('click', '.js-checkin', function(event) {
        event.preventDefault();
        window.Queue.enqueue("checkin-queue", $(this).attr('data-pid'));
        .....

funnel/templates/event.html
vidya-ram: This is also readable since queue names are clearly mentioned. 
Why have another copy when it is not required. Less lines of code

funnel/templates/event.html
shreyas-satish: > Any future behavior js shouldn't be allowed to alter these DOM values unless data is received from the server. 

How would you enforce this? A programmer who does not know about this context could easily end up adding or changing behavior that could potentially break the app.

vidya-ram:
Moved to offline_checkin based on latest event_views

PR #143, opened 2015-10-12
Import tickets

shreyas-satish:
1. Added `get` and `upsert` in `Participant` and `SyncTicket`, and a few other helper methods to enable import.
2. Added an interface that imports and massages data from Explara's API.
3. Added `import_from_list` in `TicketClient` to import tickets. Tests in `tests/test_event_models.py`.


funnel/extapi/explara.py
```diff
@@ -0,0 +1,69 @@
...
+        self.access_token = access_token
+        self.headers = {'Authorization': u'Bearer ' + self.access_token}
+        self.base_url = 'https://www.explara.com/api/e/{0}'
+
+    def get_events(self):
+        events = requests.post(self.base_url.format('get-all-events'), headers=self.headers).json()
+        return [{'title': e.get('eventTitle'), 'eventId': e.get('eventId')} for e in events.get('events')]
+
+    def get_ticket_types(self, explara_eventid):
+        return requests.post(self.base_url.format('get-tickets'), headers=self.headers, data={'eventId': explara_eventid}).json()
```
karthikb351: Copy of the data from the API as needed, like in `get_events` instead of passing the raw Explara API response, which might be subject to change.

funnel/extapi/explara.py
```diff
@@ -0,0 +1,69 @@
...
+
+class ExplaraAPI(object):
+    def __init__(self, access_token):
+        self.access_token = access_token
+        self.headers = {'Authorization': u'Bearer ' + self.access_token}
+        self.base_url = 'https://www.explara.com/api/e/{0}'
+
+    def get_events(self):
+        events = requests.post(self.base_url.format('get-all-events'), headers=self.headers).json()
+        return [{'title': e.get('eventTitle'), 'eventId': e.get('eventId')} for e in events.get('events')]
```
karthikb351: `event_id` instead of `eventId` since we follow snake case for keys.

funnel/extapi/explara.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def get_ticket_types(self, explara_eventid):
+        return requests.post(self.base_url.format('get-tickets'), headers=self.headers, data={'eventId': explara_eventid}).json()
+
+    def get_orders(self, explara_eventid):
+        ticket_orders = []
+        completed = False
+        from_record = 0
+        to_record = 50
```
karthikb351: Aren't these better as optional function params with defaults?

shreyas-satish: I wouldn't trust the API to follow a particular order. It seems more consistent to expect the entire data dump, and then massage as required.

jace: Could you add a comment to this effect in the source?

funnel/extapi/explara.py
```diff
@@ -0,0 +1,62 @@
...
+        self.base_url = 'https://www.explara.com/api/e/{0}'
+
+    def get_orders(self, explara_eventid):
+        ticket_orders = []
+        completed = False
+        from_record = 0
+        to_record = 50
+        while not completed:
+            payload = {'eventId': explara_eventid, 'fromRecord': from_record, 'toRecord': to_record}
+            attendee_response = requests.post(self.base_url.format('attendee-list'), headers=self.headers, data=payload).json()
```
jace: Don't use `self.base_url.format('attendee-list')`. Make this a separate function and use `self.url_for('attendee-list')`. That function can simply be:

```python
def url_for(self, endpoint):
    return 'https://www.explara.com/api/e/' + endpoint
```

funnel/extapi/explara.py
```diff
@@ -0,0 +1,62 @@
...
+            if not attendee_response.get('attendee'):
+                completed = True
+            elif isinstance(attendee_response.get('attendee'), list):
+                ticket_orders.extend([order for order in attendee_response.get('attendee')])
```
jace: This could be just `ticket_orders.extend(attendee_response.get('attendee'))`

funnel/extapi/explara.py
```diff
@@ -0,0 +1,62 @@
...
+            # after the first batch, subsequent batches are dicts with batch no. as key.
+            elif isinstance(attendee_response.get('attendee'), dict):
+                ticket_orders.extend([order for order_idx, order in attendee_response.get('attendee').items()])
```
jace: This could be just `ticket_orders.extend(attendee_response.get('attendee').values())`

funnel/extapi/explara.py
```diff
@@ -0,0 +1,62 @@
...
+                    tickets.append({
+                        'fullname': attendee.get('name'),
+                        'email': attendee.get('email'),
+                        'phone': details.get('Phone') or order.get('phoneNo', ''),
+                        'twitter': format_twitter(details.get('Twitter handle', '')),
+                        'job_title': details.get('Job title', ''),
+                        'company': details.get('Company name', ''),
+                        'city': order.get('city', ''),
+                        'ticket_no': attendee.get('ticketNo'),
+                        'ticket_type': attendee.get('ticketName').strip(),
```
jace: You should do `.strip()` for other fields too. For anyone registering from a phone, SwiftKey automatically inserts a trailing space, so unless Explara strips it, you should.

funnel/models/event.py
```diff
@@ -50,6 +50,11 @@ class Event(BaseScopedNameMixin, db.Model):
     participants = db.relationship('Participant', secondary='attendee', backref='events', lazy='dynamic')
     __table_args__ = (db.UniqueConstraint('proposal_space_id', 'name'),)
 
+    @classmethod
+    def get_name(cls, space, title):
+        event = cls.query.filter_by(title=title, proposal_space=space).one_or_none()
+        return event and event.name
```
jace: This pattern can be confusing. Use `return event.name if event else None`

funnel/models/event.py
```diff
@@ -50,6 +50,11 @@ class Event(BaseScopedNameMixin, db.Model):
     participants = db.relationship('Participant', secondary='attendee', backref='events', lazy='dynamic')
     __table_args__ = (db.UniqueConstraint('proposal_space_id', 'name'),)
 
+    @classmethod
+    def get_name(cls, space, title):
```
jace: At the risk of being verbose, I'd call this `get_name_from_title` so it's obvious what it does when you look at the call elsewhere in code.

funnel/models/event.py
```diff
@@ -65,6 +70,11 @@ class TicketType(BaseScopedNameMixin, db.Model):
     events = db.relationship('Event', secondary=event_ticket_type)
     __table_args__ = (db.UniqueConstraint('proposal_space_id', 'name'),)
 
+    @classmethod
+    def get_name(cls, space, title):
+        ticket_type = cls.query.filter_by(title=title, proposal_space=space).one_or_none()
+        return ticket_type and ticket_type.name
```
jace: As above

funnel/models/event.py
```diff
@@ -97,9 +107,35 @@ class Participant(BaseMixin, db.Model):
 
     __table_args__ = (db.UniqueConstraint('proposal_space_id', 'email'),)
 
+    @classmethod
+    def get(cls, space, email):
+        return cls.query.filter_by(proposal_space=space, email=email).one_or_none()
+
+    @classmethod
+    def upsert(cls, **fields):
+        participant = cls.get(fields['proposal_space'], fields['email'])
```
jace: Put these in the mandatory parameters to `upsert`.

funnel/models/event.py
```diff
@@ -97,9 +107,35 @@ class Participant(BaseMixin, db.Model):
...
+        if participant:
+            participant._set_fields(fields)
+        else:
+            participant = cls(**fields)
+            db.session.add(participant)
+        return participant
+
+    def update_events(self, events, cancel=False):
```
jace: Split this into two methods. I had to read the code to understand that `cancel` reverses the operation.

jace: You should also consider using SQLAlchemy's `collection_class=set` in the relationship, to make this easier to perform without special functions.

funnel/models/event.py
```diff
@@ -124,6 +160,29 @@ class TicketClient(BaseMixin, db.Model):
...
         backref=db.backref('ticket_clients', cascade='all, delete-orphan'))
 
+    def import_from_list(self, space, ticket_list, cancel_list=[]):
+        """
+        Batch upserts the tickets and its associated ticket types and participants.
+        Cancels the tickets in cancel_list.
+        """
+        for ticket_dict in ticket_list:
+            ticket_type = TicketType.upsert(space, TicketType.get_name(space, ticket_dict['ticket_type']),
+                            title=ticket_dict['ticket_type'], proposal_space=space)
```
jace: This is an anti-pattern. You shouldn't be calling `TicketType.<method>` twice in the same call. If `title` is so important to this class, change the class to index `title`, and change `get` and `upsert` to accept either `name` or `title` as an identifier for the row.

funnel/models/event.py
```diff
@@ -143,5 +202,32 @@ class SyncTicket(BaseMixin, db.Model):
     ticket_client_id = db.Column(db.Integer, db.ForeignKey('ticket_client.id'), nullable=False)
     ticket_client = db.relationship(TicketClient,
         backref=db.backref('sync_tickets', cascade='all, delete-orphan'))
-
     __table_args__ = (db.UniqueConstraint('proposal_space_id', 'order_no', 'ticket_no'),)
+
+    @classmethod
+    def get(cls, space, order_no, ticket_no):
+        return cls.query.filter_by(proposal_space=space, order_no=order_no, ticket_no=ticket_no).one_or_none()
```
jace: Are `order_no` and `ticket_no` guaranteed unique? Doesn't that depend on `ticket_client` as well? You'll have to accept `ticket_client` instead of `space` here and update the unique constraint above as well.

You may also want to remove `proposal_space` from this model as that is irrelevant if `ticket_client` also links to the space.

funnel/models/event.py
```diff
@@ -143,5 +202,32 @@ class SyncTicket(BaseMixin, db.Model):
...
+                             ticket_no=ticket_no, ticket_type=ticket_type,
+                             participant=participant, ticket_client=ticket_client)
+            db.session.add(ticket)
+        else:
+            # Transfer ticket
+            if ticket.participant is not participant:
+                ticket.participant.update_events(ticket.ticket_type.events, cancel=True)
+                ticket.participant = participant
+
+        ticket.participant.update_events(ticket.ticket_type.events)
```
jace: Upsert on a SyncTicket instance modifies a Participant instance? That's not good. This code doesn't belong here.

funnel/models/event.py
```diff
@@ -143,5 +202,32 @@ class SyncTicket(BaseMixin, db.Model):
...
+
+        return ticket
+
+    @classmethod
+    def exclude(cls, space, ticket_client, ticket_nos):
+        return cls.query.filter_by(proposal_space=space, ticket_client=ticket_client)\
```
jace: Avoid `\`. Push the `)` to the next line instead.

funnel/util.py
```diff
@@ -17,3 +18,11 @@ def geonameid_from_location(text):
...
         return set(geonameids)
     return None
+
+
+def format_twitter(twitter_id):
+    """
+    formats a user given twitter handle
+    Eg: https://twitter.com/shreyas_satish -> shreyas_satish, @shreyas_satish -> shreyas_satish
+    """
+    return urlparse(str(twitter_id)).path.replace('/', '').replace('@', '')
```
jace: Won't this break `https://twitter.com/shreyas_satish/favorites` by returning `shreyas_satishfavorites`?

alembic/versions/2d55f804a4ed_remove_proposal_space_id_from_sync_.py
```diff
@@ -0,0 +1,28 @@
...
+
+def upgrade():
+    op.drop_constraint(u'sync_ticket_proposal_space_id_order_no_ticket_no_key', 'sync_ticket', type_='unique')
+    op.create_unique_constraint(u'sync_ticket_ticket_client_id_order_no_ticket_no_key', 'sync_ticket', ['ticket_client_id', 'order_no', 'ticket_no'])
+    op.drop_constraint(u'sync_ticket_proposal_space_id_fkey', 'sync_ticket', type_='foreignkey')
+    op.drop_column('sync_ticket', 'proposal_space_id')
+
+
+def downgrade():
+    op.add_column('sync_ticket', sa.Column('proposal_space_id', sa.INTEGER(), autoincrement=False, nullable=False))
```
jace: This will fail if you have existing data. You have to reset the value of this column to be `sync_ticket.proposal_space_id`.

alembic/versions/416a2f958279_add_title_index_to_event_and_ticket_type.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '416a2f958279'
+down_revision = '2d55f804a4ed'
+
+from alembic import op
+
+
+def upgrade():
+    op.create_unique_constraint(u'event_proposal_id_title_key', 'event', ['proposal_space_id', 'title'])
```
jace: Constraint should be named `event_proposal_space_id_title_key`

funnel/extapi/explara.py
```diff
@@ -0,0 +1,79 @@
...
+
+import requests
+from ..util import format_twitter
+from ..models import SyncTicket
+
+__all__ = ['ExplaraAPI']
+
+
+def strip_or_empty(val):
+    return val.strip() if val else ''
```
jace: Baseframe has a `StripWhitespace()` validator that does this for you.

funnel/extapi/explara.py
```diff
@@ -0,0 +1,79 @@
...
+                        'company': strip_or_empty(details.get('Company name')),
+                        'city': strip_or_empty(order.get('city')),
+                        'ticket_no': strip_or_empty(attendee.get('ticketNo')),
+                        'ticket_type': strip_or_empty(attendee.get('ticketName')),
+                        'order_no': strip_or_empty(order.get('orderNo')),
+                    })
+        return tickets
+
+    def import_tickets(self, space, ticket_client):
+        ticket_list = self.get_tickets(ticket_client.client_eventid)
```
jace: Three methods that chain each other feels wrong, as there is no context in which you can use them independently. I would split these into three methods that don't call each other, expecting the relevant parameters instead, and a fourth helper method that calls them in a chain. This way you can have a future use case where the helper method is replaced with something that massages data as needed at each step.

funnel/models/event.py
```diff
@@ -48,7 +49,24 @@ class Event(BaseScopedNameMixin, db.Model):
     parent = db.synonym('proposal_space')
     ticket_types = db.relationship('TicketType', secondary=event_ticket_type)
     participants = db.relationship('Participant', secondary='attendee', backref='events', lazy='dynamic')
-    __table_args__ = (db.UniqueConstraint('proposal_space_id', 'name'),)
+    __table_args__ = (db.UniqueConstraint('proposal_space_id', 'name'), db.UniqueConstraint('proposal_space_id', 'title'))
+
+    @classmethod
+    def get(cls, space, name_or_title):
+        return cls.query.filter_by(proposal_space=space)\
+            .filter(or_(cls.name == name_or_title, cls.title == name_or_title)).one_or_none()
```
jace: Don't do this. Let the caller be explicit on whether they are querying by name or title. Take `name` and `title` as parameters and do a XOR test on them (via this non-XOR pattern that can extend to multiple inputs, unlike XOR): `if (not not name) + (not not title) != 1: raise TypeError()`

`not not` is a shorthand for `bool()` that avoids the function call.

funnel/models/event.py
```diff
@@ -48,7 +49,24 @@ class Event(BaseScopedNameMixin, db.Model):
...
+
+    @classmethod
+    def upsert(cls, space, name_or_title, **fields):
```
jace: Separate `name` and `title` here as well, making both parameters have default value `None`. In normal use your `name` is auto-generated, so you can call without a `name` parameter.

funnel/models/event.py
```diff
@@ -63,7 +81,24 @@ class TicketType(BaseScopedNameMixin, db.Model):
         backref=db.backref('ticket_types', cascade='all, delete-orphan'))
     parent = db.synonym('proposal_space')
     events = db.relationship('Event', secondary=event_ticket_type)
-    __table_args__ = (db.UniqueConstraint('proposal_space_id', 'name'),)
+    __table_args__ = (db.UniqueConstraint('proposal_space_id', 'name'), db.UniqueConstraint('proposal_space_id', 'title'))
+
+    @classmethod
+    def get(cls, space, name_or_title):
```
jace: As above.

funnel/models/event.py
```diff
@@ -63,7 +81,24 @@ class TicketType(BaseScopedNameMixin, db.Model):
...
+        return cls.query.filter_by(proposal_space=space)\
+            .filter(or_(cls.name == name_or_title, cls.title == name_or_title)).one_or_none()
+
+    @classmethod
+    def upsert(cls, space, name_or_title, **fields):
```
jace: As above.

funnel/models/event.py
```diff
@@ -97,9 +132,38 @@ class Participant(BaseMixin, db.Model):
...
+        return cls.query.filter_by(proposal_space=space, email=email).one_or_none()
+
+    @classmethod
+    def upsert(cls, space, emailid, **fields):
+        participant = cls.get(space, emailid)
+        if participant:
+            participant._set_fields(fields)
+        else:
+            fields.pop('proposal_space', None)
+            fields.pop('email', None)
```
jace: Remove these two lines and change the parameters above to match the column names.

funnel/models/event.py
```diff
@@ -137,11 +232,39 @@ class SyncTicket(BaseMixin, db.Model):
...
         backref=db.backref('sync_tickets', cascade="all, delete-orphan"))
-    proposal_space_id = db.Column(db.Integer, db.ForeignKey('proposal_space.id'), nullable=False)
-    proposal_space = db.relationship(ProposalSpace,
-        backref=db.backref('sync_tickets', cascade='all, delete-orphan'))
     ticket_client_id = db.Column(db.Integer, db.ForeignKey('ticket_client.id'), nullable=False)
     ticket_client = db.relationship(TicketClient,
         backref=db.backref('sync_tickets', cascade='all, delete-orphan'))
+    __table_args__ = (db.UniqueConstraint('ticket_client_id', 'order_no', 'ticket_no'),)
+
+    previous_participant = None
```
jace: A class-level variable? Every time you set it, you'll be editing the class, not the instance.

funnel/models/event.py
```diff
@@ -137,11 +232,39 @@ class SyncTicket(BaseMixin, db.Model):
...
+    def upsert(cls, ticket_client, order_no, ticket_no, **fields):
+        """
+        Returns a tuple containing the upserted ticket, and the participant the ticket
+        was previously associated with or None if there was no earlier participant.
+        """
+        ticket = cls.get(ticket_client, order_no, ticket_no)
+        if ticket:
+            if ticket.participant is not fields.get('participant'):
+                # Transfer ticket
+                ticket.previous_participant = ticket.participant
```
jace: This will break, as above.

jace: (Actually, it may not, since you're referring to the instance and not the class. Still, seems shady.)

funnel/util.py
```diff
@@ -17,3 +18,16 @@ def geonameid_from_location(text):
...
+    """
+    formats a user given twitter handle
+    Eg: https://twitter.com/shreyas_satish -> shreyas_satish, @shreyas_satish -> shreyas_satish
+    Returns None if handle is Falsy or contains an URL whose domain is not twitter.com
+    """
+    # Return None for invalid cases
+    if not handle or (urlparse(handle).netloc and urlparse(handle).netloc != 'twitter.com'):
+        return None
+
+    return unicode([part for part in urlparse(handle).path.split('/') if part][0]).replace('@', '')
```
jace: You're calling `urlparse` thrice. Put the result in a variable.

funnel/models/event.py
shreyas-satish: Can't do that since `upsert` will complain that it's getting multiple values for the same keyword argument.

funnel/extapi/explara.py
shreyas-satish: But, that's written for a form field? I need a pure function in this case.

alembic/versions/2d55f804a4ed_remove_proposal_space_id_from_sync_.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+
+# revision identifiers, used by Alembic.
+revision = '2d55f804a4ed'
+down_revision = '48ce908329c0'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql import table, column, select
+from funnel.models import (SyncTicket, TicketClient)
```
jace: Never import from the main code in a migration. Migrations are supposed to last forever, so there could be a future in which this import will fail.

funnel/extapi/explara.py
jace: One of the ways to do this data import cleanly is by passing it through a form so the data mangling parts are nicely isolated. That said, the form is possibly not the best from a performance standpoint, so calling it in a loop like here will slow things down.

funnel/models/event.py
```diff
@@ -28,7 +28,30 @@ def make_private_key():
     )
 
 
-class Event(BaseScopedNameMixin, db.Model):
+class ScopedNameTitleMixin(object):
+    # TODO: Move this into coaster?
```
jace: Let's do the move if another use case comes up for this.

funnel/models/event.py
```diff
@@ -28,7 +28,30 @@ def make_private_key():
...
+    @classmethod
+    def get(cls, parent, current_name=None, current_title=None):
+        if current_name:
```
jace: I think you'll need to ensure both aren't provided because of the small possibility of name and title being out of sync in real world data (`make_name` drops some characters when constructing names, so it is technically possible for two titles to have the same generated name). Currently the TypeError is raised only if _neither_ is provided, not _both_.

funnel/models/event.py
```diff
@@ -28,7 +28,30 @@ def make_private_key():
     )
 
 
-class Event(BaseScopedNameMixin, db.Model):
+class ScopedNameTitleMixin(object):
```
jace: Derive this from `BaseScopedNameMixin` so you don't need the class again in the use cases below. It's easy to get confused about method resolution order in base classes and order them the wrong way (L-R is correct), but hard to get confused about parent-child relationships. Since this class replaces methods in `BaseScopedNameMixin`, derive from it.

jace: MRO woes: https://github.com/hasgeek/funnel/commit/a1a81a805f81f7aecc80211c452f42aa6cfa9833

Had to clean up like this across all projects because I hadn't realised it was left to right.

funnel/models/event.py
jace: Don't pass them in `fields` then. :) Or if the dict is coming from elsewhere and guaranteed to have these keys, don't pass them in explicitly.

funnel/models/event.py
shreyas-satish: This shouldn't be a problem for `Event` and `TicketType` because of their uniqueness constraints. But, from a more general perspective, yeah, there should be a case that checks for both `current_title` and `current_name` being passed.

funnel/util.py
```diff
@@ -17,3 +18,25 @@ def geonameid_from_location(text):
...
+
+
+def format_twitter_handle(handle):
+    """
+    formats a user given twitter handle
+    Eg:  -> shreyas_satish, @shreyas_satish -> shreyas_satish
+    Returns None for invalid cases.
+    Twitter restricts the length of handles to 15. 16 is the threshold here, since
+    a user might prefix their handle with an '@', a valid case.
+    Tested in tests/test_util.py
```
jace: Docstrings should be in ReStructuredText. This one will format badly in ReST. (We don't generate documentation from docstrings in Funnel, but only because we've limited it to libraries so far (Coaster and Flask-Lastuser).)

funnel/util.py
```diff
@@ -17,3 +18,25 @@ def geonameid_from_location(text):
...
+    """
+    if not handle:
+        return None
+
+    parsed_handle = urlparse(handle)
+    if (
+            (parsed_handle.netloc and parsed_handle.netloc != 'twitter.com') or
+            (not parsed_handle.netloc and len(handle) > 16)
+    ):
```
jace: Just a formatting thing, but I've usually used:

```python
    if ((parsed_handle.netloc and parsed_handle.netloc != 'twitter.com') or
            (not parsed_handle.netloc and len(handle) > 16)):
```

No strongly held opinion on which is better. Long lines just help seeing more code on screen at once.

funnel/models/event.py
shreyas-satish: There's also the case where the `proposal_space` or the `email` would want to be updated for a particular record. That's why the fields are popped only during an insert.

funnel/util.py
shreyas-satish: Sure. Since this is essentially pattern matching, I felt this style reflects that the best, with the patterns listed out one below the other, un-indented.

jace: Also a personal thing, but I usually indent the closing bracket (PEP8 requires it to be unindented).

funnel/models/event.py
jace: Let's disambiguate by using a `current_` or `new_` prefix then. It's unclear why one is `space` and another `proposal_space` otherwise (or `emailid` and `email`).

jace:
Done with review. Clear apart from the `upsert` parameter name concerns.

PR #147, opened 2015-11-12
Event views

shreyas-satish:
- Manage events, ticket types and participants for a proposal space.
- Added background job to import tickets from external ticket vendors (currently, there's only support for Explara).
- Make HTML badges (with QR Codes) for participants.
- Enable check-in process for participants.
- Serve API for basic contact exchange through [zalebi](https://github.com/hasgeek/zalebi).
- Added rq config.


funnel/forms/participant.py
```diff
@@ -0,0 +1,28 @@
...
+
+class ParticipantForm(forms.Form):
+    fullname = forms.StringField(__("Full Name"), validators=[forms.validators.DataRequired()])
+    email = forms.EmailField(__("Email"), validators=[forms.validators.DataRequired(), forms.validators.Length(max=80)])
+    phone = forms.StringField(__("Phone number"), validators=[forms.validators.Length(max=80)])
+    city = forms.StringField(__("City"), validators=[forms.validators.Length(max=80)])
+    company = forms.StringField(__("Company"), validators=[forms.validators.Length(max=80)])
+    job_title = forms.StringField(__("Job Title"), validators=[forms.validators.Length(max=80)])
+    twitter = forms.StringField(__("Twitter"), validators=[forms.validators.Length(max=15)])
+    events = QuerySelectMultipleField(__("Events"),
```
jace: Doesn't this need a model or query factory?

jace: Because of the concerns in hasgeek/baseframe#112, I'd move this context into this file itself if possible.

funnel/forms/space.py
```diff
@@ -2,13 +2,15 @@
 
 import re
 from coaster.utils import sorted_timezones
+from wtforms.widgets import CheckboxInput, ListWidget
+from wtforms.ext.sqlalchemy.fields import QuerySelectMultipleField
```
jace: I think we should expose the SQLAlchemy fields in Baseframe forms so they don't need specific import here.

funnel/jobs/jobs.py
```diff
@@ -0,0 +1,16 @@
+from ..models import (db, TicketClient, SyncTicket)
+from ..extapi.explara import ExplaraAPI
+from flask.ext.rq import job
+from funnel import app
+
+
+@job('talkfunnel')
```
jace: Make this queue `funnel` since we're still using `funnel` as the package name for this project (and as the user and database name in production).

funnel/models/event.py
```diff
@@ -144,6 +144,18 @@ def remove_events(self, events):
...
+    def checkin_list(cls, event):
+        participant_attendee_join = db.join(Participant, Attendee, Participant.id == Attendee.participant_id)
+        stmt = db.select([Participant.id, Participant.fullname, Participant.email, Participant.company, Participant.twitter, Participant.puk, Participant.key, Attendee.checked_in, Participant.badge_printed]).select_from(participant_attendee_join).where(Attendee.event_id == event.id).order_by(Participant.fullname)
+        return db.session.execute(stmt).fetchall()
+
+    @classmethod
+    def filter_by_ticket_type(cls, ticket_type):
+        participant_sync_ticket_join = db.join(Participant, SyncTicket, Participant.id == SyncTicket.participant_id)
+        stmt = db.select([Participant.id, Participant.fullname, Participant.email, Participant.company, Participant.twitter]).select_from(participant_sync_ticket_join).where(SyncTicket.ticket_type == ticket_type).distinct()
+        return db.session.execute(stmt).fetchall()
```
jace: Curious why you're stepping below the declarative ORM layer with this approach. This could simply be:

```python
return Participant.query.join(SyncTicket).filter(SyncTicket.ticket_type == ticket_type).all()
```

Unless the `DISTINCT` bit is doing something additional.

funnel/models/space.py
```diff
@@ -180,6 +186,13 @@ def permissions(self, user, inherited=None):
                     'edit-schedule',
                     'new-session',
                     'edit-session',
+                    'event-checkin',
+                    'event-view',
+                    'ticket-type-view',
+                    'event-sync',
+                    'participant-edit',
+                    'participant-view',
+                    'new-participant'
```
jace: We'll need to be consistent with `verb-noun` or `noun-verb` order with permissions. Our convention appears to be `verb-noun` for permission names but `noun-verb` for view names (which in itself is a confusing difference).

funnel/models/event.py
shreyas-satish: Fixed.

funnel/models/space.py
```diff
@@ -226,6 +239,22 @@ def url_for(self, action='view', _external=False):
...
+        elif action == 'new-participant':
+            return url_for('new_participant', profile=self.profile.name, space=self.name)
+        elif action == 'new-ticket-type-participant':
+            return url_for('new_ticket_type_participant', profile=self.profile.name, space=self.name)
+        elif action == 'new-event':
+            return url_for('new_event', profile=self.profile.name, space=self.name)
+        elif action == 'new-ticket-type':
+            return url_for('new_ticket_type', profile=self.profile.name, space=self.name)
+        elif action == 'new-ticket-client':
+            return url_for('new_ticket_client', profile=self.profile.name, space=self.name)
```
jace: Hopefully we'll clean this up in hasgeek/coaster#77. :)

funnel/util.py
```diff
@@ -40,8 +43,32 @@ def format_twitter_handle(handle):
...
+
+
+def split_name(fullname):
+    """
+    Splits a given fullname into two parts
+    a first name, and a concanetated last name.
+    Eg: "ABC DEF EFG" -> ("ABC", "DEF EFG")
+    """
+    name_splits = fullname.split()
+    return name_splits[0], " ".join([s for s in name_splits[1:]])
```
jace: Doesn't this need a check for names that don't have a space?

funnel/util.py
```diff
@@ -1,8 +1,11 @@
-from . import app
+import StringIO
```
jace: Usually we do `from StringIO import StringIO` because there's nothing else we need from the class. Also, in some cases this construct helps:

```python
try:
    from cStringIO import StringIO
except ImportError:
    from StringIO import StringIO
```

If you're dealing with potentially large data, cStringIO is faster, but won't work under pypy so you need the fallback.

funnel/views/event.py
```diff
@@ -0,0 +1,177 @@
...
+
+@app.route('/<space>/admin', methods=['GET', 'POST'], subdomain='<profile>')
+@lastuser.requires_login
+@load_models(
+    (Profile, {'name': 'profile'}, 'g.profile'),
+    ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
+    permission='admin')
+def admin(profile, space):
+    attendee_sync_form = forms.Form()
+    if attendee_sync_form.validate_on_submit():
```
jace: Use (a) `if forms.Form().validate_on_submit()` or (b) name it `csrf_form` so the purpose is less ambiguous.

funnel/views/event.py
```diff
@@ -0,0 +1,177 @@
...
+        for ticket_client in space.ticket_clients:
+            if ticket_client and ticket_client.name == u'explara':
+                import_tickets.delay(app.config['ENV'], ticket_client.id)
```
jace: Don't provide `init_for` env to the background job. We use separate mechanisms in RQ to do that when the worker starts, not when the job starts.

funnel/views/event.py
```diff
@@ -0,0 +1,177 @@
...
+    if form.validate_on_submit():
+        event = Event(proposal_space=space)
+        form.populate_obj(event)
+        event.make_name()
+        try:
+            db.session.add(event)
+            db.session.commit()
+        except IntegrityError:
+            db.session.rollback()
+            flash(_(u"This event already exists."), 'info')
```
jace: Switch to `db.session().add_and_commit` here.

funnel/views/event.py
```diff
@@ -0,0 +1,177 @@
...
+    if form.validate_on_submit():
+        ticket_type = TicketType(proposal_space=space)
+        form.populate_obj(ticket_type)
+        ticket_type.make_name()
+        try:
+            db.session.add(ticket_type)
+            db.session.commit()
+        except IntegrityError:
+            db.session.rollback()
+            flash(_(u"This ticket type already exists."), 'info')
```
jace: Use `add_and_commit`

funnel/views/event.py
```diff
@@ -0,0 +1,177 @@
...
+@app.route('/<space>/ticket_type/<name>/edit', methods=['GET', 'POST'], subdomain='<profile>')
+@lastuser.requires_login
+@load_models(
+    (Profile, {'name': 'profile'}, 'g.profile'),
+    ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
+    (TicketType, {'name': 'name', 'proposal_space': 'space'}, 'ticket_type'),
+    permission='event-edit')
+def ticket_type_edit(profile, space, ticket_type):
+    form = TicketTypeForm(obj=ticket_type, model=TicketType)
+    form.events.query = space.events
```
jace: Add an `__init__` method to the form that calls `super`, then sets `self.events.query = self.obj.parent.events`.

funnel/util.py
shreyas-satish: Added check for null case. But, if a name doesn't have a space, this just returns an empty string for last name.

funnel/views/event.py
```diff
@@ -0,0 +1,177 @@
...
+@load_models(
+    (Profile, {'name': 'profile'}, 'g.profile'),
+    ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
+    (Event, {'name': 'name', 'proposal_space': 'space'}, 'event'),
+    permission='event-view')
+def event(profile, space, event):
+    participants = Participant.checkin_list(event)
+    form = ParticipantBadgeForm()
+    if form.validate_on_submit():
+        badge_printed = True if form.data.get('badge_printed') == 't' else False
```
jace: You can use `coaster.utils.getbool` for this.

funnel/views/participant.py
```diff
@@ -0,0 +1,164 @@
...
+    form.events.query = space.events
+    if form.validate_on_submit():
+        participant = Participant(proposal_space=space)
+        form.populate_obj(participant)
+        try:
+            db.session.add(participant)
+            db.session.commit()
+        except IntegrityError:
+            db.session.rollback()
+            flash(_(u"This participant already exists."), 'info')
```
jace: Use `add_and_commit`.

funnel/views/participant.py
```diff
@@ -0,0 +1,164 @@
...
+        try:
+            contact_exchange = ContactExchange(user_id=g.user.id, participant_id=participant.id, proposal_space_id=space.id)
+            db.session.add(contact_exchange)
+            db.session.commit()
+        except IntegrityError:
+            app.logger.warning(u"Contact Exchange already present")
+            db.session.rollback()
+        return jsonp(participant=participant_data(participant, space.id, full=True))
+    else:
+        return jsonp(message=u"Unauthorized", code=401)
```
jace: I'll recommend using `jsonify` instead of `jsonp` as you don't want cross-site access to this endpoint, and maybe use `@render_with(json=True)` for the entire function. You can then `return {'message': "Unauthorized"}, 401`

funnel/views/participant.py
```diff
@@ -0,0 +1,164 @@
...
+    (Participant, {'id': 'participant_id'}, 'participant'),
+    permission='event-checkin')
+def event_checkin(profile, space, event, participant):
+    attendee = Attendee.get(event, participant)
+    form = forms.Form()
+    if form.validate_on_submit():
+        # Toggle check-in status
+        attendee.checked_in = not attendee.checked_in
+        db.session.commit()
+    return redirect(url_for('event', profile=space.profile.name, space=space.name, name=event.name), code=303)
```
jace: This function looks like it should (a) return current status for a JSON request and (b) redirect for a HTML request. (`render_with` can't do this yet).

funnel/views/participant.py
shreyas-satish: But, would that work for the mobile app? @karthikb351

jace: Yeah, mobile isn't affected as the client isn't a browser.

funnel/forms/participant.py
shreyas-satish: The events need to be scoped by the proposal space, which is why they're set in the view.

PR #150, opened 2015-11-19
Offline checkin feature for a smooth registration workflow during internet outage at events.

vidya-ram:
Workflow:
1) When a participant is checked-in, the participant id is added to a checkin queue which is stored in localStorage. Loading and abort indicators are displayed for this participant.
2) Every 8 seconds items from check-in queue and cancel check-in queue are read and ajax post is send to the server. The participant ids are removed from queue on successful post. Same is done for cancel check-in queue.
3) Every one minute, participants data is requested from the server and UI is refreshed by ractive. Check-in status of participant ids in the queue(stored in localStorage) are compared with data received from the server , if their status have already been updated (in case of multiple check-in counters) then their are removed from the queue. All participant ids remaining in the queue will have the loading indicators.
4) On aborting a check-in, if the participant id is present in check-in queue then it is removed from the queue. If participant id is not present in the check-in queue it implies check-in has already been posted to the server, so the participant id is added to the cancel check-in queue.


vidya-ram:
Wrong target branch. It is `offline_checkin` into `event_view.

PR #151, opened 2015-11-20
Offline checkin feature for a smooth registration workflow during internet outage at events.

vidya-ram:
Workflow:
1) When a participant is checked-in, the participant id is added to a checkin queue which is stored in localStorage. Loading and abort indicators are displayed for this participant.
2) Every 8 seconds items from check-in queue and cancel check-in queue are read and ajax post is send to the server. The participant ids are removed from queue on successful post. Same is done for cancel check-in queue.
3) Every one minute, participants data is requested from the server and UI is refreshed by ractive. Check-in status of participant ids in the queue(stored in localStorage) are compared with data received from the server , if their status have already been updated (in case of multiple check-in counters) then their are removed from the queue. All participant ids remaining in the queue will have the loading indicators.
4) On aborting a check-in, if the participant id is present in check-in queue then it is removed from the queue. If participant id is not present in the check-in queue it implies check-in has already been posted to the server, so the participant id is added to the cancel check-in queue.


funnel/static/js/scripts.js
```diff
@@ -1,80 +1,299 @@
...
+    $.ajax({
+      type: 'GET',
+      url: participantTable.Config.participantlistUrl,
+      timeout: 5000,
+      dataType: 'json',
+      success: function(data) {
+        participantTable.participantTableContent.set('participants', data.participants).then(function() {
+          $('.js-loader').hide();
+          $('.js-total').text(data.total_participants);
+          $('.js-totalcheckin').text(data.total_checkedin);
```
shreyas-satish: `js-total` and `js-totalcheckin` can automatcally update if the data is updated, right? Why do we need to reset it with jQuery?

funnel/static/js/scripts.js
```diff
@@ -1,80 +1,299 @@
...
+      $(this).serializeArray().map(function(obj) {
+        formValues[obj.name] = obj.value;
+      });
+      if (formValues["checkin"] === "t") {
+        participantTable.checkinQ.enqueue(formValues["pid"]);
+      }
+      else {
+        participantTable.cancelcheckinQ.enqueue(formValues["pid"]);
+      }
+      $(this).find('.js-loader').show();
```
shreyas-satish: Wouldn't it be simpler to have the `js-loader` show or hide based on some state variable eg: `loading`? That would relieve us from explicitly showing and hiding the loader.

funnel/static/js/scripts.js
```diff
@@ -1,80 +1,299 @@
 
-function radioHighlight(radioName, highlightClass) {
-    var selector = "input[name='" + radioName + "']";
-    $(selector + ":checked").parent().addClass(highlightClass);
-    var handler = function() {
-        $(selector).parent().removeClass(highlightClass);
-        $(selector + ":checked").parent().addClass(highlightClass);
+window.Talkfunnel = {};
+
+if (Modernizr.localstorage) {
```
shreyas-satish: Do we need this check? All the major browsers seem to [support](http://caniuse.com/#search=localstorage) localstorage.

funnel/static/js/scripts.js
```diff
@@ -1,80 +1,299 @@
...
+    }
+  };
+
+  window.Talkfunnel.Queue = function(queueName) {
+    this.queueName = queueName;
+
+    //Adds a participant_id to queue
+    this.enqueue = function(participant_id) {
+      var participantList = window.Talkfunnel.Store.read(this.queueName) || [];
+      if(participantList.indexOf(participant_id) === -1) {
```
shreyas-satish: space after `if`

funnel/static/js/scripts.js
```diff
@@ -1,80 +1,304 @@
...
+  this.enqueue = function(participant_id) {
+    var participantList = window.Talkfunnel.Store.read(this.queueName) || [];
+    if (participantList.indexOf(participant_id) === -1) {
+      participantList.push(participant_id);
+      return window.Talkfunnel.Store.add(this.queueName, participantList);
+    }
+  };
+
+  //Reads and returns all items from queue
+  //Returns undefined when queue is empty or not defined
```
shreyas-satish: When queue is empty, isn't it better to return `[]`?

PR #152, opened 2015-11-23
Added description meta tag.

vidya-ram:

### DIFF
--- a/funnel/templates/proposal.html
+++ b/funnel/templates/proposal.html
@@ -2,7 +2,13 @@
 {% from "baseframe/forms.html" import renderfield, renderform %}
 {% from "comments.html" import commenttree %}
 {% block title %}{{ proposal.title }} &mdash; {{ space.title }}{% endblock %}
-{% block description %}{{ proposal.title }}{% if proposal.speaker %}{% trans %} by {% endtrans%}{{ proposal.speaker.pickername }}{% endif %}, {{ space.title }}{% endblock %}
+{% block description %}
+  {%- if proposal.speaker -%}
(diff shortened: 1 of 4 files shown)


funnel/templates/proposal.html
```diff
@@ -2,6 +2,7 @@
 {% from "baseframe/forms.html" import renderfield, renderform %}
 {% from "comments.html" import commenttree %}
 {% block title %}{{ proposal.title }} &mdash; {{ space.title }}{% endblock %}
+{% block description %}{{ proposal.title }}{% if proposal.speaker %}{% trans %} by {% endtrans%}{{ proposal.speaker.pickername }}{% endif %}, {{ space.title }}{% endblock %}
```
jace: The `{% trans %}` block has to be for the entire string. You can't assume all languages will have an equivalent for "by" that appears before the speaker's name.

However, you can't have an `{% if %}` inside it, so you have to move the `if` block out as well and repeat `{% proposal.title %}` under both `if` and `else` conditions.

funnel/templates/space.html
```diff
@@ -1,6 +1,7 @@
 {% extends "layout.html" %}
 {% from "macros.html" import rsvpform, rsvpscript %}
 {% block title %}{{ space.title }}{% endblock %}
+{% block description %}{{ space.title }} &ndash; {% if g.profile %}{{ g.profile.description|striptags }}{% endif %}{% endblock %}
```
jace: You shouldn't have a trailing dash (in case `g.profile` is false). The dash should be inside the if block.

PR #155, opened 2015-12-15
added ui for adding a parent space

shreyas-satish:


funnel/forms/space.py
```diff
@@ -42,6 +42,7 @@ class ProposalSpaceForm(forms.Form):
     explore_url = forms.URLField(__("Explore tab URL"),
         description=__(u"Page containing the explore tab’s contents, for the mobile app"),
         validators=[forms.validators.Optional()])
+    parent_space = QuerySelectField(u"Parent Space", get_label='title', allow_blank=True)
```
jace: Should be sentence case with gettext, so `__(u"Parent space")`.

PR #156, opened 2015-12-16
Import from csv

shreyas-satish:
Lets an event manager import participants from a CSV file. Incorporates suggestions from https://github.com/hasgeek/funnel/pull/132.

### DIFF
--- a/funnel/forms/participant.py
+++ b/funnel/forms/participant.py
@@ -6,7 +6,7 @@ from baseframe import __
 from wtforms.ext.sqlalchemy.fields import QuerySelectMultipleField
 from wtforms.widgets import CheckboxInput, ListWidget
 
-__all__ = ['ParticipantForm', 'ParticipantBadgeForm']
+__all__ = ['ParticipantForm', 'ParticipantBadgeForm', 'ParticipantImportForm']
 
 
 class ParticipantForm(forms.Form):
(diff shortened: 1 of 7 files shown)


funnel/forms/participant.py
```diff
@@ -26,3 +26,28 @@ class ParticipantForm(forms.Form):
 class ParticipantBadgeForm(forms.Form):
     choices = [('', "Badge printing status"), ('t', "Printed"), ('f', "Not printed")]
     badge_printed = forms.SelectField("", choices=[(val_title[0], val_title[1]) for val_title in choices])
+
+
+# TODO: Move to Baseframe
+class ValidFile(object):
+    def __init__(self, allowed_exts=[], message=None):
```
jace: You need a way to signal that all extensions are allowed. Check for an empty list below.

funnel/forms/participant.py
```diff
@@ -26,3 +26,28 @@ class ParticipantForm(forms.Form):
...
+        ext = os.path.splitext(field.data.filename)[1].strip(".")
+        if ext and ext.lower() not in self.allowed_exts:
+            raise wtforms.validators.StopValidation(self.message)
+
+
+class ParticipantImportForm(forms.Form):
+    error_msg = __(u"Please upload a valid CSV file")
+    participant_list = forms.FileField(__("Participant list"),
+        description=u"Expected headers: fullname*, email*, phone, twitter, company, job_title, city",
+        validators=[forms.validators.DataRequired(error_msg), ValidFile(allowed_exts=['csv'], message=error_msg)])
```
jace: File input fields can take an `accept` attribute with a list of mimetypes (only). Can we do that with FileField? (Chrome supports extensions as well, but Safari doesn't.)

PR #164, opened 2016-02-02
added customizable labels for objective and description

shreyas-satish:


funnel/forms/proposal.py
```diff
@@ -60,6 +60,15 @@ class ProposalForm(forms.Form):
     location = forms.StringField(__("Your location"), validators=[forms.validators.DataRequired(), forms.validators.Length(max=80)],
         description=__("Your location, to help plan for your travel if required"))
 
+    def __init__(self, *args, **kwargs):
+        super(ProposalForm, self).__init__(*args, **kwargs)
+        space = kwargs.get('parent')
+        self.objective.label.text = space.labels['proposal']['partA']['title']
+        self.objective.description = space.labels['proposal']['partA']['hint']
+        self.description.label.text = space.labels['proposal']['partB']['title']
+        self.description.description = space.labels['proposal']['partB']['hint']
```
jace: I'd put all four in defensive `if` statements so that we set the label and description only if they are present in `space.labels`. It's a dict, so it's entirely possible to have a scenario where the values are missing.

jace: Also, since the model has `partA` and `partB` as synonyms, we could change the form to use names instead. `form.populate_obj(…)` will still work fine.

funnel/models/proposal.py
```diff
@@ -93,9 +93,11 @@ class Proposal(BaseScopedIdNameMixin, CoordinatesMixin, db.Model):
     section = db.relationship(ProposalSpaceSection, primaryjoin=section_id == ProposalSpaceSection.id,
         backref="proposals")
     objective = MarkdownColumn('objective', nullable=True)
+    partA = db.synonym('objective')
     session_type = db.Column(db.Unicode(40), nullable=True)
     technical_level = db.Column(db.Unicode(40), nullable=True)
     description = MarkdownColumn('description', nullable=True)
+    partB = db.synonym('description')
```
jace: The capitalization of A and B violates our style guide. This should be `part_a` and `part_b` or `part1` and `part2`.

funnel/models/space.py
```diff
@@ -65,6 +87,7 @@ class ProposalSpace(BaseScopedNameMixin, db.Model):
     parent_space_id = db.Column(None, db.ForeignKey('proposal_space.id', ondelete='SET NULL'), nullable=True)
     parent_space = db.relationship('ProposalSpace', remote_side='ProposalSpace.id', backref='subspaces')
     inherit_sections = db.Column(db.Boolean, default=True, nullable=False)
+    labels = db.Column(JsonDict, nullable=False, server_default=json.dumps(labels_default()))
```
jace: Avoid `server_default` if possible (purely for reason of convention). Let this be a client default. You also don't need `json.dumps` in this case, the column will massage a dictionary into JSON automatically.

alembic/versions/570f4ea99cda_add_labels_to_proposal_space.py
```diff
@@ -0,0 +1,30 @@
+import json
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql import table, column
+from coaster.sqlalchemy import JsonDict
+from funnel.models.space import labels_default
```
jace: Migration must never reference code. This migration may be run on a future version of the Funnel codebase where `labels_default` is no longer defined. It should still work.

Also, `server_default` is not recommended (see comment below), so you can just remove this.

PR #167, opened 2016-04-18
Add missing dependency

braunmagrin:
When I tried to run `python manage.py db create` I found that the `rq` module was missing from the `requirements.txt`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ Flask-Script==0.5.3
 alembic
 git+https://github.com/tobiasandtobias/flask-alembic.git
 qrcode
+rq


jace:
Thanks. Can you confirm flask-rq isn't a distinct dependency as well?

braunmagrin:
@jace Sure, I'll have a look at it tomorrow. But as far as I could see, just by installing `rq` it already worked.

braunmagrin:
Is the problem with the `KeyError: 'LASTUSER_SERVER'` on the CI log normal? I also got it here in my machine actually. I guess there are some configurations missing, right?

jace:
We don't have Funnel working under Travis yet. The current integration is there to remind us it's not done yet.

For the Lastuser settings, you'll need to either make an installation locally (what we do in dev) or use our hosted version at https://auth.hasgeek.com (recommended). Register an app there and add the server URL, app key and secret to Funnel's settings.

PR #176, opened 2016-11-16
Use POST instead of GET for registering votes

shreyas-satish:


funnel/views/commentvote.py
```diff
@@ -1,52 +1,58 @@
...
     ((Proposal, ProposalRedirect), {'url_name': 'proposal', 'proposal_space': 'space'}, 'proposal'),
     permission='vote-proposal', addlperms=lastuser.permissions)
 def proposal_voteup(profile, space, proposal):
+    csrf_form = forms.Form()
+    if not csrf_form.validate_on_submit():
+        abort(401)
     proposal.votes.vote(g.user, votedown=False)
     db.session.commit()
     flash(_("Your vote has been recorded"), 'info')
     return redirect(proposal.url_for())
```
jace: Redirects should be with status 303 given this is a POST. Also, return JSON status for XHR requests.

funnel/views/commentvote.py
```diff
@@ -1,52 +1,58 @@
...
 @lastuser.requires_login
 @load_models(
     (Profile, {'name': 'profile'}, 'g.profile'),
     ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
     ((Proposal, ProposalRedirect), {'url_name': 'proposal', 'proposal_space': 'space'}, 'proposal'),
     permission='vote-proposal', addlperms=lastuser.permissions)
 def proposal_voteup(profile, space, proposal):
+    csrf_form = forms.Form()
+    if not csrf_form.validate_on_submit():
+        abort(401)
```
jace: 403

Issue #177, opened 2017-02-13
Date picker not working in safari


booleanbalaji:
When trying to create a new proposal space on Safari, the date picker doesn't work, and it throws an Internal Server error when trying to manually enter the date in the YYYY-MM-DD format shown in the helper text.

jace:
Closing as this is obsolete.

PR #182, opened 2017-03-10
Changed error message to reflect form field.

harshalbhatia:
The form asks date in DD-MM-YYYY format and error asked for YYYY-MM-DD format. Just a minor thing i noticed.

### DIFF
--- a/funnel/forms/space.py
+++ b/funnel/forms/space.py
@@ -21,9 +21,9 @@ class ProposalSpaceForm(forms.Form):
     title = forms.StringField(__("Title"), validators=[forms.validators.DataRequired()])
     datelocation = forms.StringField(__("Date and Location"), validators=[forms.validators.DataRequired(), forms.validators.Length(max=50)])
     date = forms.DateField(__("Start date (for sorting)"),
-        validators=[forms.validators.DataRequired(__("Enter a valid date in YYYY-MM-DD format"))])
+        validators=[forms.validators.DataRequired(__("Enter a valid date in DD-MM-YYYY format"))])
     date_upto = forms.DateField(__("End date (for sorting)"),
-        validators=[forms.validators.DataRequired(__("Enter a valid date in YYYY-MM-DD format"))])
+        validators=[forms.validators.DataRequired(__("Enter a valid date in DD-MM-YYYY format"))])
     tagline = forms.StringField(__("Tagline"), validators=[forms.validators.DataRequired()],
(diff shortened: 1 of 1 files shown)


jace:
What browser did you test in? DateField renders with `type='date'`, which some browsers recognise and render a custom widget for, using the browser's locale, but will submit back in `YYYY-MM-DD` format. Browsers that don't recognise it (like Safari?) render as plain text.

The error message hasn't been updated to recognise that some browsers do custom rendering these days. However, simply asking for DD-MM-YYYY format is incorrect.

harshalbhatia:
I tested on both Desktop and Mobile chrome.
I was thinking the same thing actually. We can word the error message in a generic way instead of mentioning a date format.

jace:
Since this is a `DataRequired` validator, let's make it `__("This is required")`

jace:
FWIW, http://caniuse.com/#feat=input-datetime

Can we possibly use a polyfill to add support to broken browsers?

harshalbhatia:
In issue #177 Balaji reports that the date picker doesn't work and that he got an error. I tried recreating the same (Safari, iPad, iOS 9) but the date picker worked fine. (Screenshots attached)
Might not be the same on Desktop Safari. I don't have access to a mac, I can't test that.

img_0159
img_0160

jace:
Safari for iOS supports `input="date"` while Safari for macOS doesn't. [Go figure](http://caniuse.com/#feat=input-datetime). I'm guessing it has something to do with disagreements over [Shadow DOM](http://caniuse.com/#feat=shadowdom), which is the technical basis on which desktop browsers implement the widgets. Mobile browsers just use device-native widgets.

funnel/forms/space.py
```diff
@@ -21,9 +21,9 @@ class ProposalSpaceForm(forms.Form):
     title = forms.StringField(__("Title"), validators=[forms.validators.DataRequired()])
     datelocation = forms.StringField(__("Date and Location"), validators=[forms.validators.DataRequired(), forms.validators.Length(max=50)])
     date = forms.DateField(__("Start date (for sorting)"),
-        validators=[forms.validators.DataRequired(__("Enter a valid date in YYYY-MM-DD format"))])
+        validators=[forms.validators.DataRequired(__("This field is required"))])
```
jace: Let's use the same text in both places: "This is required"

harshalbhatia:
Added a new commit after changing to a generic error message. Looking further into a viable polyfill.

PR #185, opened 2017-04-11
Add <space>/criteria

harshalbhatia:
Added a criteria field to spaces. Right now the description is one big blob. As discussed with @karthikb351, it would be good to separate them out into separate fields.

The product decisions can be discussed, but one key technical doubt that i have is how to viably do the database migration while creating a new `nullable=False` field. I can either add the constraint along with a `server_default` value or create the new column first and then add the constraint [1].
For now, I have not added the constraint. Please let me know how you guys generally do it. (cc @shreyas-satish)

[1] http://stackoverflow.com/a/41026374


shreyas-satish:
Closing this for now, since the requirement isn't clear and the `state` property is revamped.

PR #195, opened 2017-07-07
Added length validator to the speaker field of session form

iambibhas:
This adds a maxlength property to the form being generated so the browser doesn't let the user input more than the provided length.

### DIFF
--- a/funnel/models/__init__.py
+++ b/funnel/models/__init__.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 
-from coaster.sqlalchemy import (TimestampMixin, BaseMixin, BaseNameMixin, BaseScopedNameMixin,
-    BaseScopedIdNameMixin, BaseIdNameMixin, MarkdownColumn, JsonDict, CoordinatesMixin)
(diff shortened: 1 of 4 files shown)


funnel/models/__init__.py
```diff
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 
-from coaster.sqlalchemy import (TimestampMixin, BaseMixin, BaseNameMixin, BaseScopedNameMixin,
-    BaseScopedIdNameMixin, BaseIdNameMixin, MarkdownColumn, JsonDict, CoordinatesMixin)
+from coaster.sqlalchemy import (TimestampMixin, BaseMixin, BaseNameMixin,
+    BaseScopedNameMixin, BaseScopedIdNameMixin, BaseIdNameMixin, MarkdownColumn,
+    JsonDict, CoordinatesMixin, make_timestamp_columns)
```
iambibhas: Missing `make_timestamp_columns` from import was making the tests fail as it was still being imported from here inside `label.py` https://github.com/hasgeek/funnel/blob/d2d1763faf83b46e89acdaddd00b16162245d239/funnel/models/label.py#L3

Issue #205, opened 2017-09-10
Broken 'coaster' dependency


ultimatecoder:
* **Description:** The [coaster][coaster] dependency versioned `0.6.dev1` is broken.

* **My Configuration:**
    * Operating System: Ubuntu 16.04 LTS
    * Python Version: Python 2.7.12
    * pip version: 9.0.1

* **Traceback:**
```
$ python manage.py db create
Traceback (most recent call last):
  File "manage.py", line 3, in <module>
    from coaster.manage import init_manager
  File "/home/myuser/.virtualenvs/funnel_dev/local/lib/python2.7/site-packages/coaster/__init__.py", line 6, in <module>
    from .utils import *  # NOQA
ImportError: No module named utils
```

* **Why to not report at [coaster][coaster] and here?**
I believe the development version should not be linked to any project rather a released version should. Perfectly fine if it is the latest version. ~~I might supply a patch linking [coaster][coaster] from [pypi](https://pypi.python.org/) and not the GitHub repository.~~ Created PR #206 

[coaster]: https://github.com/hasgeek/coaster

jace:
Please report coaster issues on coaster, not here. In HasGeek's repositories, the `master` branch is always stable, with development happening on other branches.

PR #206, opened 2017-09-10
[FIX] Fixing issue #205

ultimatecoder:
* **Description:** Deprecating GitHub reference for package coaster to prevent unreleased version glitch. Linking coaster package through [pypi](https://pypi.python.org/pypi/coaster/) to get a released
version. This fixes issue #205

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ Flask-WTF
 pytz
 unicodecsv
 icalendar
-git+https://github.com/hasgeek/coaster.git
+coaster
 git+https://github.com/hasgeek/flask-lastuser.git
 git+https://github.com/hasgeek/baseframe.git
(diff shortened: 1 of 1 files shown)


jace:
Unfortunately, HasGeek apps are always linked to the development versions of our libraries as they are developed in sync. If there is a problem with Coaster's master branch, please report an issue or send a PR there.

jace:
Rationale: Coaster is built on the requirements that come out of apps, so there's usually a pressing need to use the feature immediately in the app. The full use case for any new functionality only manifests as we use it across apps, and we sometimes find we've made mistakes and need to introduce a breaking change. Since we're committed to not breaking API for anything released to PyPI without a considerable notice period, this means the release is withheld until the API is mature.

PR #208, opened 2017-09-14
provide access to contact API until event has concluded

shreyas-satish:


funnel/views/participant.py
```diff
@@ -106,6 +107,8 @@ def participant_edit(profile, space, participant):
     ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
     permission='view')
 def participant(profile, space):
+    if space.date_upto and datetime.combine(space.date_upto, datetime.max.time()) < datetime.utcnow():
```
jace: 1. This uses 23:59:59 of the current day instead of midnight of the next day.
2. This does not account for timezone. Using UTC will make an allowance until 5:30 AM the next day, in India, but will fall short of the day in a Western timezone. Since we do have a `timezone` string column on `space`, use it.

funnel/views/participant.py
```diff
@@ -5,6 +5,8 @@
 from baseframe import forms
 from baseframe.forms import render_form
 from coaster.views import load_models
+from coaster.utils import midnight_to_utc
+from datetime import datetime, date, timedelta
```
jace: `date` import is no longer used.

PR #212, opened 2017-10-31
Added proposal instructions, fixes #211

### LINKED ISSUE #211: Customise explanatory text when submitting a proposal
When submitting a new proposal or editing one, proposers are given a single line explanation:

```html
This form uses <a href="http://daringfireball.net/projects/markdown/">Markdown</a> for formatting.
```

Event managers currently load submission instructions onto the landing page, where the text is unnecessary for those only reading the proposals. This text should be moved into the submission page:

1. [ ] `models.ProposalSpace` currently has a `description` column. It needs an additional `instructions` column (with a migration).
2. [ ] `forms.ProposalSpaceForm` needs a corresponding `instructions` field.
3. [ ] `views.proposal_new` and `views.proposal_edit` need to use the `instructions` text if available, falling back to the default.
4. [ ] Since the text may be long, requiring the proposer to scroll down a fair bit (especially if they are reloading after validation errors), it should collapse by default. `render_form` does not support collapsible text in the message, so there's an upstream dependency.

https://github.com/hasgeek/funnel/blob/42a610f04b15f095bad18f894729787b717f0558/funnel/models/space.py#L38
https://github.com/hasgeek/funnel/blob/42a610f04b15f095bad18f894729787b717f0558/funnel/forms/space.py#L31-L32
https://github.com/hasgeek/funnel/blob/42a610f04b15f095bad18f894729787b717f0558/funnel/views/proposal.py#L127-L130
https://github.com/hasgeek/funnel/blob/42a610f04b15f095bad18f894729787b717f0558/funnel/views/proposal.py#L160-L162

iambibhas:
- Added `instructions` field to `ProposalSpace` model and form
 - Showing it on proposal add and edit page if available, otherwise showing the default message

selection_128

Side note: We should add dividers above and below the instructions


funnel/forms/space.py
```diff
@@ -30,6 +30,8 @@ class ProposalSpaceForm(forms.Form):
         validators=[forms.validators.Optional()])
     description = forms.MarkdownField(__("Description"), validators=[forms.validators.DataRequired()],
         description=__("About Event"))
+    instructions = forms.MarkdownField(__("Instructions"),
+        description=__("Instructions for submitting a Proposal"), default=u'')
```
jace: `"These instructions will be shown to the proposer with their submission form"`

migrations/versions/152442fce4e4_update_null_ticket_client_id.py
```diff
@@ -0,0 +1,39 @@
...
+def upgrade():
+    # updating ticket_client_id=1 as we seemed to have missed some
+    # records when they were created. They are sitting as NULL in the db.
+    # And we need to get rid of them.
+    connection = op.get_bind()
+    sync_ticket = table(
+        'sync_ticket',
+        column(u'id', sa.INTEGER()),
+        column(u'ticket_client_id', sa.INTEGER())
+    )
```
jace: Move the table outside the function.

migrations/versions/152442fce4e4_update_null_ticket_client_id.py
```diff
@@ -0,0 +1,39 @@
...
+    connection.execute(
+        sync_ticket.update().where(
+            sync_ticket.c.ticket_client_id == None
+        ).values({
+            'ticket_client_id': 1
```
jace: What is this? Where do the `None` and `1` values come from?

jace: @shreyas-satish seems to be something you may know about.

migrations/versions/d576f55f9eba_proposal_instructions.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'd576f55f9eba'
+down_revision = 'f80ab5d0e11e'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
jace: Remove these autogenerated Alembic comments when you add the migration code.

migrations/versions/152442fce4e4_update_null_ticket_client_id.py
iambibhas: @jace Yes. @shreyas-satish has the context for it.

funnel/forms/space.py
```diff
@@ -30,6 +30,8 @@ class ProposalSpaceForm(forms.Form):
         validators=[forms.validators.Optional()])
     description = forms.MarkdownField(__("Description"), validators=[forms.validators.DataRequired()],
         description=__("About Event"))
+    instructions = forms.MarkdownField(__("Instructions"),
+        description=__("These instructions will be shown to the proposer with their submission form"), default=u'')
```
iambibhas: @shreyas-satish I did provide a default value in the form.

PR #214, opened 2017-11-28
Using the new StateManager

iambibhas:


funnel/models/proposal.py
```diff
@@ -8,6 +8,7 @@
 from .commentvote import CommentSpace, VoteSpace, SPACETYPE
 from coaster.utils import LabeledEnum
 from coaster.sqlalchemy import SqlSplitIdComparator
+from coaster.sqlalchemy.statemanager import StateManager
```
jace: Import `StateManager` from `coaster.sqlalchemy`. Add it to the previous line's imports.

funnel/models/proposal.py
```diff
@@ -108,7 +109,10 @@ class Proposal(BaseScopedIdNameMixin, CoordinatesMixin, db.Model):
     slides = db.Column(db.Unicode(250), nullable=True)
     preview_video = db.Column(db.Unicode(250), default=u'', nullable=True)
     links = db.Column(db.Text, default=u'', nullable=True)
-    status = db.Column(db.Integer, default=PROPOSALSTATUS.SUBMITTED, nullable=False)
+
+    _status = db.Column('status', db.Integer, StateManager.check_constraint('status', PROPOSALSTATUS),
+        default=PROPOSALSTATUS.SUBMITTED, nullable=False)
+    status = StateManager('_status', PROPOSALSTATUS, doc="The proposal's status")
```
jace: Change this to `state`. We'll keep the db column name `status` for now -- or throw in a migration.

funnel/models/proposal.py
```diff
@@ -194,7 +198,7 @@ def status_title(self):
 
     @hybrid_property
     def confirmed(self):
-        return self.status == PROPOSALSTATUS.CONFIRMED
+        return self.status.value == PROPOSALSTATUS.CONFIRMED
```
jace: Remove this method entirely. Change all references to it from `proposal.confirmed` to `proposal.state.CONFIRMED`.

funnel/models/space.py
```diff
@@ -42,7 +45,10 @@ class ProposalSpace(BaseScopedNameMixin, db.Model):
     date_upto = db.Column(db.Date, nullable=True)
     website = db.Column(db.Unicode(250), nullable=True)
     timezone = db.Column(db.Unicode(40), nullable=False, default=u'UTC')
-    status = db.Column(db.Integer, default=SPACESTATUS.DRAFT, nullable=False)
+
+    _status = db.Column('status', db.Integer, StateManager.check_constraint('status', SPACESTATUS),
+        default=SPACESTATUS.DRAFT, nullable=False)
+    status = StateManager('_status', SPACESTATUS, doc="This Proposal Space's status")
```
jace: Make this `state` as well.

funnel/views/login.py
```diff
@@ -11,7 +11,7 @@
 @app.route('/login')
 @lastuser.login_handler
 def login():
-    return {'scope': 'id email phone organizations teams'}
+    return {'scope': 'id email phone organizations teams local.imgee.two:image'}
```
iambibhas: oh dang, this needs to go. did it by mistake.

funnel/models/space.py
```diff
@@ -115,7 +121,7 @@ def proposals_by_status(self):
             basequery = Proposal.query.filter(Proposal.proposal_space_id.in_([self.id] + [s.id for s in self.subspaces]))
         else:
             basequery = Proposal.query.filter_by(proposal_space=self)
-        return dict((status, basequery.filter_by(status=status).order_by(db.desc('created_at')).all()) for (status, title) in PROPOSALSTATUS.items() if status != PROPOSALSTATUS.DRAFT)
+        return dict((status, basequery.filter_by(_status=status).order_by(db.desc('created_at')).all()) for (status, title) in PROPOSALSTATUS.items() if status != PROPOSALSTATUS.DRAFT)
```
jace: This multiple query thing is a bad idea. It needs to go away.

1. In `PROPOSALSTATUS` add a group for all values that are not `DRAFT`.
2. Use [SQLAlchemy's `group_by`](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.group_by) on the state column.
3. Turn the groups into a dictionary.

funnel/models/space.py
```diff
@@ -125,8 +131,8 @@ def proposals_by_confirmation(self):
         else:
             basequery = Proposal.query.filter_by(proposal_space=self)
         response = dict(
-            confirmed=basequery.filter_by(status=PROPOSALSTATUS.CONFIRMED).order_by(db.desc('created_at')).all(),
-            unconfirmed=basequery.filter(Proposal.status != PROPOSALSTATUS.CONFIRMED, Proposal.status != PROPOSALSTATUS.DRAFT).order_by(db.desc('created_at')).all())
+            confirmed=basequery.filter_by(_status=PROPOSALSTATUS.CONFIRMED).order_by(db.desc('created_at')).all(),
+            unconfirmed=basequery.filter(~Proposal._status.in_([PROPOSALSTATUS.CONFIRMED, PROPOSALSTATUS.DRAFT])).order_by(db.desc('created_at')).all())
```
jace: Never touch `_status` or `_state` again. Add `UNCONFIRMED` as a group that includes everything that is not `DRAFT` or `CONFIRMED`. Use `group_by` here to get everything in one query.

funnel/models/space.py
```diff
@@ -178,7 +184,7 @@ def permissions(self, user, inherited=None):
         perms = super(ProposalSpace, self).permissions(user, inherited)
         perms.add('view')
         if user is not None:
-            if self.status == SPACESTATUS.SUBMISSIONS:
+            if self.status.value == SPACESTATUS.SUBMISSIONS:
```
jace: Use `if self.state.SUBMISSIONS:`

funnel/models/space.py
```diff
@@ -303,7 +309,7 @@ def all(cls):
         """
         Return currently active events, sorted by date.
         """
-        return cls.query.filter(cls.status >= 1).filter(cls.status <= 4).order_by(cls.date.desc()).all()
+        return cls.query.filter(cls._status >= 1).filter(cls._status <= 4).order_by(cls.date.desc()).all()
```
jace: Replace with a group named `CURRENTLY_LISTED`. Change this query to:
```python
return cls.query.filter(cls.state.CURRENTLY_LISTED).order_by(cls.date.desc()).all()
```

funnel/templates/proposal.html.jinja2
```diff
@@ -121,9 +121,9 @@
           <iframe class="facebooklike" src="//www.facebook.com/plugins/like.php?app_id=114496105304651&amp;href={{ request.url }}&amp;send=false&amp;layout=button_count&amp;width=150&amp;show_faces=false&amp;action=recommend&amp;colorscheme=light&amp;font&amp;height=21" scrolling="no" frameborder="0" style="border:none; overflow:hidden; width:150px; height:21px;" allowTransparency="true"></iframe>
         </div>
         <div class="col-md-3 col-lg-3 col-sm-12 col-xs-12">
-          Status: 
+          Status:
           {%- if 'view-status' in g.permissions %}
-            {{ PROPOSALSTATUS[proposal.status] }}
+            {{ PROPOSALSTATUS[proposal.status.value] }}
```
jace: You should no longer have `PROPOSALSTATUS` available in templates. Remove it from wherever this template is rendered. Change this to `{{ proposal.state.label }}`.

funnel/templates/space.html.jinja2
```diff
@@ -112,19 +112,19 @@
...
-    {%- elif space.status == 4 -%}
+    {%- elif space.status.value == 4 -%}
       {% trans %}Open for post-event feedback{% endtrans %}
-    {%- elif space.status == 5 -%}
+    {%- elif space.status.value == 5 -%}
       {% trans %}Closed{% endtrans %}
-    {%- elif space.status == 6 -%}
+    {%- elif space.status.value == 6 -%}
       {% trans %}Withdrawn{% endtrans %}
     {%- endif -%}
```
jace: Replace this entire `if` block with a simple `{{ space.state.label }}`.

funnel/templates/space.html.jinja2
```diff
@@ -177,7 +177,7 @@
     {% endfor %}
   </ul>
 {%- endif %}
-{% if space.status == 1 -%}
+{% if space.status.value == 1 -%}
```
jace: `{% if space.state.OPEN %}` (check if that is the state name)

funnel/views/index.py
```diff
@@ -11,7 +11,7 @@
 def index():
     g.profile = None
     g.permissions = []
-    spaces = ProposalSpace.query.filter_by(parent_space=None).filter(ProposalSpace.profile != None).filter(ProposalSpace.status >= 1).filter(ProposalSpace.status <= 4).order_by(ProposalSpace.date.desc()).all()  # NOQA
+    spaces = ProposalSpace.query.filter_by(parent_space=None).filter(ProposalSpace.profile != None).filter(ProposalSpace._status >= 1).filter(ProposalSpace._status <= 4).order_by(ProposalSpace.date.desc()).all()  # NOQA
```
jace: Replace with `.filter(ProposalSpace.state.CURRENTLY_LISTED)`

funnel/views/index.py
```diff
@@ -42,7 +42,7 @@ def spaces_json(profile):
 @load_model(Profile, {'name': 'profile'}, 'g.profile', permission='view')
 def profile_view(profile):
     spaces = ProposalSpace.query.filter(ProposalSpace.profile == profile, ProposalSpace.parent_space == None).filter(
-        ProposalSpace.status >= 1).filter(ProposalSpace.status <= 4).order_by(ProposalSpace.date.desc()).all()
+        ProposalSpace._status >= 1).filter(ProposalSpace._status <= 4).order_by(ProposalSpace.date.desc()).all()
```
jace: Likewise.

iambibhas:
Just noticed that there is also `COMMENTSTATUS` for comments that needs to be ported to statemanager. I'll add it.

funnel/views/proposal.py
```diff
@@ -80,7 +80,7 @@ def proposal_data(proposal):
             ('votes_count', proposal.votes_count()),
             ('votes_groups', proposal.votes_by_group()),
             ('votes_bydate', proposal.votes_by_date()),
-            ('status', proposal.status),
+            ('status', proposal.status.value),
```
jace: Use `proposal.status.label`. The value is now private internal detail. (This may break the Android app.)

We should change the `LabeledEnum` to use (value, name, title) tuples and return `proposal.status.label.name` here so that `name` can be a consistent non-localized string. However, all other references to `.label` in the UI will need to become `.label.title`.

funnel/views/proposal.py
```diff
@@ -89,7 +89,7 @@ def proposal_data_flat(proposal, groups=[]):
     cols = [data.get(header) for header in proposal_headers if header not in ('votes_groups', 'votes_bydate')]
     for name in groups:
         cols.append(data['votes_groups'][name])
-    cols.append(PROPOSALSTATUS[proposal.status])
+    cols.append(PROPOSALSTATUS[proposal.status.value])
```
jace: `cols.append(proposal.status.label.name)`

jace: PROPOSALSTATUS should no longer imported anywhere.

iambibhas:
and `RSVP_STATUS` too.

funnel/views/proposal.py
```diff
@@ -184,9 +184,9 @@ def proposal_edit(profile, space, proposal):
 def proposal_status(profile, space, proposal):
     form = ProposalStatusForm()
     if form.validate_on_submit():
-        proposal.status = form.status.data
+        proposal._status = form.status.data
         db.session.commit()
-        flash(_("The proposal has been ") + PROPOSALSTATUS[proposal.status].lower(), 'success')
+        flash(_("The proposal has been ") + PROPOSALSTATUS[proposal.status.value].lower(), 'success')
```
jace: `flash(_("The proposal's status has changed to ") + proposal.state.label.title, 'success')`

However, this entire view needs to go away and be replaced with transitions (separate PR).

funnel/views/proposal.py
```diff
@@ -291,8 +291,8 @@ def proposal_view(profile, space, proposal):
                 flash(_("No such comment"), 'error')
             return redirect(proposal.url_for(), code=303)
     links = [Markup(linkify(unicode(escape(l)))) for l in proposal.links.replace('\r\n', '\n').split('\n') if l]
-    if proposal.status != PROPOSALSTATUS.DRAFT:
-        statusform = ProposalStatusForm(status=proposal.status)
+    if proposal.status.value != PROPOSALSTATUS.DRAFT:
```
jace: `if not proposal.state.DRAFT:`

funnel/views/space.py
```diff
@@ -26,7 +26,7 @@ def space_data(space):
         'timezone': space.timezone,
         'start': space.date.isoformat() if space.date else None,
         'end': space.date_upto.isoformat() if space.date_upto else None,
-        'status': space.status,
+        'status': space.status.value,
```
jace: `'status': space.state.label.name`

jace: For API compatibility, maybe we should include both:
```python
'status': space.status.value,
'state': space.state.label.name,
```
And remove `status` once the Android app is updated. (Also applies to the other example above.)

funnel/models/space.py
iambibhas: I was going to ask you about this. that multiple query was bothering me as well.

funnel/views/proposal.py
iambibhas: so it'll be like `SUBMITTED = (1, 'submitted', __("Submitted"))`?

funnel/models/space.py
iambibhas: @jace `group_by` is used for aggregations, right? that doesn't really apply here.

funnel/models/commentvote.py
```diff
@@ -95,7 +98,9 @@ class Comment(BaseMixin, db.Model):
 
     message = MarkdownColumn('message', nullable=False)
 
-    status = db.Column(db.Integer, default=0, nullable=False)
+    _state = db.Column('status', db.Integer, StateManager.check_constraint('status', COMMENTSTATUS),
```
jace: If you're adding a CHECK constraint, you need a corresponding migration.

funnel/models/commentvote.py
```diff
@@ -111,7 +116,7 @@ def delete(self):
         Delete this comment.
         """
         if len(self.children) > 0:
-            self.status = COMMENTSTATUS.DELETED
+            self._state = COMMENTSTATUS.DELETED
```
jace: Never mutate `_state`. Add a `soft_delete` transition instead, and call it.
```python
@state.transition(None, state.DELETED)
def soft_delete(self):
    self.user = None
    self.message = ''
```

funnel/models/proposal.py
```diff
@@ -24,18 +24,25 @@
...
+    UNDER_EVALUATION = (8, 'under_evaluation', __("Under evaluation"))
+    SHORTLISTED_FOR_REHEARSAL = (9, 'shortlisted_for_rehearsal', __("Shortlisted for rehearsal"))
+    REHEARSAL = (10, 'rehearsal', __("Rehearsal ongoing"))
+
+    # everything except draft and confirmed
+    UNCONFIRMED = {SUBMITTED, WAITLISTED, SHORTLISTED, REJECTED, CANCELLED, AWAITING_DETAILS,
+                   UNDER_EVALUATION, SHORTLISTED_FOR_REHEARSAL, REHEARSAL}
+    # everything except draft
+    NOT_DRAFT = {SUBMITTED, CONFIRMED, WAITLISTED, SHORTLISTED, REJECTED, CANCELLED,
+                 AWAITING_DETAILS, UNDER_EVALUATION, SHORTLISTED_FOR_REHEARSAL, REHEARSAL}
```
jace: You don't need `NOT_DRAFT`. In any place you're using this, just use `not state.DRAFT`.

funnel/models/proposal.py
```diff
@@ -108,7 +115,10 @@ class Proposal(BaseScopedIdNameMixin, CoordinatesMixin, db.Model):
     slides = db.Column(db.Unicode(250), nullable=True)
     preview_video = db.Column(db.Unicode(250), default=u'', nullable=True)
     links = db.Column(db.Text, default=u'', nullable=True)
-    status = db.Column(db.Integer, default=PROPOSALSTATUS.SUBMITTED, nullable=False)
+
+    _state = db.Column('status', db.Integer, StateManager.check_constraint('status', PROPOSALSTATUS),
```
jace: CHECK constraint needs a migration.

funnel/models/rsvp.py
```diff
@@ -27,9 +28,9 @@ class Rsvp(TimestampMixin, db.Model):
     user_id = db.Column(None, db.ForeignKey('user.id'), nullable=False, primary_key=True)
     user = db.relationship(User)
 
-    status = db.Column(db.CHAR(1),
-        db.CheckConstraint('status IN (%s)' % ', '.join(["'%s'" % value for value in RSVP_STATUS.keys()])),
+    _state = db.Column('status', db.CHAR(1), StateManager.check_constraint('status', RSVP_STATUS),
```
jace: Please confirm the CHECK constraint hasn't changed. Migration required if it has.

funnel/models/rsvp.py
```diff
@@ -46,12 +47,12 @@ def _space_rsvp_for(self, user, create=False):
...
 def _space_rsvps_with(self, status):
-    return self.rsvps.filter_by(status=status)
+    return self.rsvps.filter_by(_state=status)
 
 
 def _space_rsvp_counts(self):
-    return {value: count for value, count in db.session.query(Rsvp.status, db.func.count(Rsvp.status)).filter(
-        Rsvp.proposal_space == self).group_by(Rsvp.status)}
+    return dict(db.session.query(Rsvp._state, db.func.count(Rsvp._state)).filter(
+        Rsvp.proposal_space == self).group_by(Rsvp._state).all())
```
jace: This use of `_state` doesn't feel right, but I'll have to see what we can do about it.

funnel/models/space.py
```diff
@@ -42,7 +49,10 @@ class ProposalSpace(BaseScopedNameMixin, db.Model):
     date_upto = db.Column(db.Date, nullable=True)
     website = db.Column(db.Unicode(250), nullable=True)
     timezone = db.Column(db.Unicode(40), nullable=False, default=u'UTC')
-    status = db.Column(db.Integer, default=SPACESTATUS.DRAFT, nullable=False)
+
+    _state = db.Column('status', db.Integer, StateManager.check_constraint('status', SPACESTATUS),
```
jace: CHECK migration required.

funnel/models/space.py
```diff
@@ -110,23 +120,29 @@ def proposals_all(self):
...
     @property
     def proposals_by_status(self):
-        from .proposal import Proposal, PROPOSALSTATUS
+        from .proposal import Proposal
         if self.subspaces:
             basequery = Proposal.query.filter(Proposal.proposal_space_id.in_([self.id] + [s.id for s in self.subspaces]))
         else:
             basequery = Proposal.query.filter_by(proposal_space=self)
-        return dict((status, basequery.filter_by(status=status).order_by(db.desc('created_at')).all()) for (status, title) in PROPOSALSTATUS.items() if status != PROPOSALSTATUS.DRAFT)
+        all_proposals = basequery.filter(Proposal.state.NOT_DRAFT).order_by(db.desc('created_at')).all()
```
jace: What happens if you use `basequery.filter(not Proposal.state.DRAFT)`? We haven't tested this in Coaster. We should.

funnel/models/space.py
```diff
@@ -110,23 +120,29 @@ def proposals_all(self):
...
+        from .proposal import Proposal
         if self.subspaces:
             basequery = Proposal.query.filter(Proposal.proposal_space_id.in_([self.id] + [s.id for s in self.subspaces]))
         else:
             basequery = Proposal.query.filter_by(proposal_space=self)
         response = dict(
-            confirmed=basequery.filter_by(status=PROPOSALSTATUS.CONFIRMED).order_by(db.desc('created_at')).all(),
-            unconfirmed=basequery.filter(Proposal.status != PROPOSALSTATUS.CONFIRMED, Proposal.status != PROPOSALSTATUS.DRAFT).order_by(db.desc('created_at')).all())
+            confirmed=basequery.filter(Proposal.state.CONFIRMED).order_by(db.desc('created_at')).all(),
+            unconfirmed=basequery.filter(Proposal.state.UNCONFIRMED).order_by(db.desc('created_at')).all())
```
jace: Likewise, `not Proposal.state.CONFIRMED`

funnel/templates/proposal.html.jinja2
```diff
@@ -121,12 +121,12 @@
           <iframe class="facebooklike" src="//www.facebook.com/plugins/like.php?app_id=114496105304651&amp;href={{ request.url }}&amp;send=false&amp;layout=button_count&amp;width=150&amp;show_faces=false&amp;action=recommend&amp;colorscheme=light&amp;font&amp;height=21" scrolling="no" frameborder="0" style="border:none; overflow:hidden; width:150px; height:21px;" allowTransparency="true"></iframe>
         </div>
         <div class="col-md-3 col-lg-3 col-sm-12 col-xs-12">
-          Status: 
+          Status:
           {%- if 'view-status' in g.permissions %}
-            {{ PROPOSALSTATUS[proposal.status] }}
-            {% if proposal.session and proposal.confirmed %} &amp; Scheduled{% endif %}
+            {{ proposal.state.label }}
```
jace: This should be `{{ proposal.state.label.title }}` since you've switched to NameTitle tuples.

funnel/models/space.py
```diff
@@ -110,23 +120,29 @@ def proposals_all(self):
...
         else:
             basequery = Proposal.query.filter_by(proposal_space=self)
-        return dict((status, basequery.filter_by(status=status).order_by(db.desc('created_at')).all()) for (status, title) in PROPOSALSTATUS.items() if status != PROPOSALSTATUS.DRAFT)
+        all_proposals = basequery.filter(Proposal.state.NOT_DRAFT).order_by(db.desc('created_at')).all()
+        all_by_status = defaultdict(list)
+        for p in all_proposals:
+            # returning title instead of value because this is only used in space.html.jnja2,
+            # and that template requires the status title, not the value
+            all_by_status[p.state.label.title].append(p)
+        return all_by_status
```
jace: Titles should not be used as dictionary keys. Don't forget that they are localized strings and could cause you grief in a different language.

funnel/templates/space.html.jinja2
```diff
@@ -189,7 +173,7 @@
 #}
       {%- if 'view-status' in g.permissions %}
         {%- for status, proposal_group in space.proposals_by_status.iteritems() if proposal_group -%}
-            <h3>{{ PROPOSALSTATUS[status] }} {% trans %}proposals{% endtrans %}</h3>
+            <h3>{{ status }} {% trans %}proposals{% endtrans %}</h3>
```
jace: `status` will no longer be a title.

funnel/views/proposal.py
```diff
@@ -184,9 +185,10 @@ def proposal_edit(profile, space, proposal):
 def proposal_status(profile, space, proposal):
     form = ProposalStatusForm()
     if form.validate_on_submit():
-        proposal.status = form.status.data
+        proposal._status = form.status.data
```
jace: Put a `TODO` marker here. This view needs to go. State changes should require transitions.

funnel/views/space.py
```diff
@@ -108,7 +109,7 @@ def space_view_json(profile, space):
         'sections': [section_data(s) for s in sections],
         'venues': [venue_data(venue) for venue in space.venues],
         'rooms': [room_data(room) for room in space.rooms],
-        'proposals': [proposal_data(proposal) for proposal in proposals],
+        'proposals': map(proposal_data, proposals),
```
jace: List comprehensions are preferred over `map`.

funnel/models/commentvote.py
iambibhas: `db migrate` doesn't detect anything. does it have to be manually added?

funnel/models/commentvote.py
iambibhas: Why are we deleting user and message? doesn't that cause loss of data in case if we want to un-delete or see history?

funnel/models/proposal.py
iambibhas: Can't use that in sqlalchemy query though, `.filter(not Proposal.state.DRAFT)` doesn't work.

funnel/models/proposal.py
iambibhas: `db migrate` doesn't detect anything.

funnel/models/rsvp.py
iambibhas: `db migrate` doesn't detect anything.

funnel/models/space.py
iambibhas: This doesn't work. it becomes `WHERE true` as `not Proposal.state.DRAFT` becomes `True`

funnel/models/commentvote.py
iambibhas: Oh, never mind. It's an existing logic. Let me figure that out.

funnel/models/space.py
iambibhas: This will mean importing `PROPOSALSTATUS` into the template then.

funnel/models/commentvote.py
jace: Yes. Alembic's auto-detection isn't comprehensive.

funnel/models/proposal.py
jace: Use `.filter(~Proposal.state.DRAFT)`

funnel/models/space.py
jace: That only means something else is being done wrong. 😃

iambibhas: Can `Proposal.state` itself expose a method for the lookup?

iambibhas: As the dict has already been provided to it in definition.

jace: Let me think about this. You don't need a lookup as much as you need grouping by state.

funnel/models/commentvote.py
```diff
@@ -111,14 +111,17 @@ def __init__(self, **kwargs):
...
         """
         Delete this comment.
         """
         if len(self.children) > 0:
-            self._state = COMMENTSTATUS.DELETED
-            self.user = None
-            self.message = ''
+            self.soft_delete()
         else:
             if self.parent and self.parent.is_deleted:
```
jace: The `is_deleted` property should be replaced with `self.parent.state.DELETED`.

jace: Also, we could merge `soft_delete` and `delete` into a single transition. It always transitions to `DELETED` state, always erases user and message, and will optionally remove itself from the database (using `parent.children.remove(self)`) if conditions are right.

funnel/models/proposal.py
```diff
@@ -167,6 +163,10 @@ def _validate_proposal_space(self, key, value):
                 redirect.proposal = self
         return value
 
+    @state.transition(None, state.DELETED)
+    def soft_delete(self):
```
jace: This too could be named `delete` since there will no longer be hard deletes.

funnel/models/space.py
jace: There's now `Proposal.state.group(items)`.

funnel/templates/space.html.jinja2
```diff
@@ -173,7 +173,7 @@
 #}
       {%- if 'view-status' in g.permissions %}
         {%- for status, proposal_group in space.proposals_by_status.iteritems() if proposal_group -%}
-            <h3>{{ status }} {% trans %}proposals{% endtrans %}</h3>
+            <h3>{{ status.label.title }} {% trans %}proposals{% endtrans %}</h3>
```
jace: Change the iterator variable name to `state`.

funnel/views/index.py
```diff
@@ -1,17 +1,19 @@
...
 
 
 @app.route('/')
 def index():
     g.profile = None
     g.permissions = []
-    spaces = ProposalSpace.query.filter_by(parent_space=None).filter(ProposalSpace.profile != None).filter(ProposalSpace.state.CURRENTLY_LISTED).order_by(ProposalSpace.date.desc()).all()  # NOQA
+    spaces = ProposalSpace.query.filter_by(parent_space=None).filter(
+        ProposalSpace.profile != None
+    ).filter(ProposalSpace.state.CURRENTLY_LISTED).order_by(ProposalSpace.date.desc()).all()  # NOQA
```
jace: We have to change the sorting from date descending to absolute number of days ±today, future events first. That means this query needs to be a union: future events with date ascending union with past events, date descending. Union queries: http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.union

jace: This also applies to other ProposalSpace queries below.

funnel/models/space.py
```diff
@@ -303,7 +314,21 @@ def all(cls):
         """
         Return currently active events, sorted by date.
         """
-        return cls.query.filter(cls.status >= 1).filter(cls.status <= 4).order_by(cls.date.desc()).all()
+        return cls.query.filter(cls.state.CURRENTLY_LISTED).order_by(cls.date.desc()).all()
+
+    @classmethod
+    def fetch_sorted(cls, profile=None, any_profile=True):
```
shreyas-satish: The `profile` and `any_profile` kwargs don't seem to be used?

iambibhas: Oh right. they are not needed. deleting.

funnel/models/space.py
```diff
@@ -109,24 +120,26 @@ def proposals_all(self):
...
-        from .proposal import Proposal, PROPOSALSTATUS
+    def proposals_by_state(self):
+        from .proposal import Proposal
         if self.subspaces:
             basequery = Proposal.query.filter(Proposal.proposal_space_id.in_([self.id] + [s.id for s in self.subspaces]))
         else:
             basequery = Proposal.query.filter_by(proposal_space=self)
-        return dict((status, basequery.filter_by(status=status).order_by(db.desc('created_at')).all()) for (status, title) in PROPOSALSTATUS.items() if status != PROPOSALSTATUS.DRAFT)
+        all_proposals = basequery.filter(~Proposal.state.DRAFT).order_by(db.desc('created_at'))
+        all_by_state = Proposal.state.group(all_proposals)
```
shreyas-satish: There's perhaps no need for `all_proposals` and `all_by_state` here, you can return the result of the query directly.

funnel/models/space.py
```diff
@@ -109,24 +120,26 @@ def proposals_all(self):
             return self.proposals
 
     @property
-    def proposals_by_status(self):
-        from .proposal import Proposal, PROPOSALSTATUS
+    def proposals_by_state(self):
+        from .proposal import Proposal
```
shreyas-satish: Is it perhaps to better to move this method to the `Proposal` model so you don't have to import inside the function? Especially since it's proposals that are being queried for anyway.

funnel/models/space.py
```diff
@@ -109,24 +120,26 @@ def proposals_all(self):
...
 
     @property
     def proposals_by_confirmation(self):
-        from .proposal import Proposal, PROPOSALSTATUS
+        from .proposal import Proposal
         if self.subspaces:
             basequery = Proposal.query.filter(Proposal.proposal_space_id.in_([self.id] + [s.id for s in self.subspaces]))
         else:
             basequery = Proposal.query.filter_by(proposal_space=self)
         response = dict(
```
shreyas-satish: `response` is a slightly confusing variable name in this context. Like above you don't need this variable, you can return `dict(...)` directly.

funnel/models/space.py
```diff
@@ -303,7 +314,21 @@ def all(cls):
...
+        # sorts the spaces so that both new and old spaces are sorted from closest to farthest
+        now = db.func.utcnow()
+        currently_listed_spaces = cls.query.filter_by(parent_space=None).filter(
+            cls.state.CURRENTLY_LISTED
+            )
+        upcoming = currently_listed_spaces.filter(cls.date >= now).order_by(cls.date.asc())
+        past = currently_listed_spaces.filter(cls.date < now).order_by(cls.date.desc())
+
+        # union_all() because union() doesn't respect the orders mentioned in subqueries
+        all_spaces = upcoming.union_all(past)
```
shreyas-satish: Like above you can `return upcoming.union_all(past)`.

funnel/models/space.py
iambibhas: I think I like assigning variables for more readability. Is there any harm of using them?

jace: Assigning variables is _slightly_ slower because it has to allocate memory for that variable in the local namespace, and then reference it again in the next statement. There is also the potential for developer error when a name accidentally clobbers another variable with that name, or if there are two similarly named variables and the developer accidentally refers to the wrong one. I find myself making this error all the time (usually caught by failing tests).

As a matter of habit, I avoid giving names and instead leave a comment.

That said, names are useful when debugging because it helps examine each line separately, so I do this during development and remove the names later.

iambibhas: Got it. Cool. I'll update the code.

funnel/models/space.py
iambibhas: There are 3 methods - `proposals_all`, `proposals_by_state` and `proposals_by_confirmation`. I can move all 3 to the Proposal model if needed. At first I thought it makes sense to have them here as proposals have backreferences to the proposal spaces, but we don't seem to be using backreferences for these queries. in this case we can move them to Proposal model. But should I do that in this PR or create a new PR for that? Because it's going to be another level of refactoring unrelated to this PR.

PR #219, opened 2018-01-25
Change users field in UserGroupForm from textarea widget to user select widget

vidya-ram:


jace:
Don't you also need to change the view? That is where it iterates through the textarea to find users.

PR #220, opened 2018-01-29
Fixed state transition issue

iambibhas:


funnel/views/proposal.py
```diff
@@ -186,7 +186,7 @@ def proposal_edit(profile, space, proposal):
 def proposal_status(profile, space, proposal):
     form = ProposalStatusForm()
     if form.validate_on_submit():
-        proposal._status = form.status.data
+        proposal._state = form.status.data
```
shreyas-satish: Is this how it's supposed to be updated? Shouldn't it use a method like shown in https://github.com/hasgeek/coaster/blob/89995e227070fadb6d192933735a3ef8e9c46f89/coaster/sqlalchemy/statemanager.py#L151? `_state` seems to be indicated as a private variable, so perhaps not a good idea to update outside of the model?

iambibhas:
This is where I missed it - https://github.com/hasgeek/funnel/pull/214/files#r157972220

funnel/models/proposal.py
```diff
@@ -164,6 +164,50 @@ def _validate_proposal_space(self, key, value):
                 redirect.proposal = self
         return value
 
+    @state.transition(None, state.DRAFT)
```
jace: Do not use `None` for from state in these transitions.

funnel/forms/proposal.py
```diff
@@ -74,10 +74,15 @@ def __init__(self, *args, **kwargs):
...
-        choices=[(status, label.title) for (status, label) in PROPOSALSTATUS.items() if status != PROPOSALSTATUS.DRAFT])
+class ProposalTransitionForm(forms.Form):
+    transition = forms.SelectField(__("Status"), validators=[forms.validators.DataRequired()])
+
+    def populate_transitions(self, proposal):
+        # value: transition method name
+        # label: method name in title case, after removing underscores
+        self.transition.choices = sorted([
+            (tname, tname.replace("_", " ").title()) for (tname, t) in proposal.state.transitions.iteritems() if t.is_available
+        ])
```
jace: Transitions are actions, not choices. A `SelectField` is the wrong UI element. Remove this form entirely and create a page with buttons. Process available transitions in the view.

funnel/templates/proposal.html.jinja2
```diff
@@ -213,7 +213,7 @@
               {% endfor %}
             </div>
 
-            <p>{{ renderform(statusform, 'proposal-status', action=proposal.url_for('status'), submit='Save') }}</p>
+            <p>{{ renderform(transitionform, 'proposal-status-transition', action=proposal.url_for('transition'), submit='Save') }}</p>
```
jace: Remove this form and insert transition buttons here.

jace:
This is what the transition UI should look like (this screenshot is from [Kharcha](/hasgeek/kharcha), which used the older Docflow instead of StateManager):

Transitions in Kharcha

iambibhas:
This is what it looks like now - 
selection_245

The UI looks like a mess. :sweat: I need some help with CSS.

iambibhas:
Moved the status down with the transitions.
selection_247

funnel/templates/proposal.html.jinja2
```diff
@@ -120,24 +120,9 @@
...
-            {% endif %}
-          {% endif %}
-          {% if proposal.session %}
-            <a class="btn btn-xs btn-info" href="{{ proposal.session.url_for() }}">View session in schedule</a>
-          {% endif %}
-          {%- if 'move-proposal' in g.permissions %}
-            <br>
-            {{ renderform(proposal_move_form, 'proposal-move', action=proposal.url_for('move-to'), submit='Move', style=None) }}
+        <div class="col-md-12 col-lg-12 col-sm-12 col-xs-12">
+          {%- if 'editor' in proposal.roles_for(g.user) %}
```
iambibhas: @jace does this look right?

jace: No, roles shouldn't be in templates. Find a way to move this to the view.

iambibhas:
selection_248

funnel/views/proposal.py
```diff
@@ -294,19 +287,24 @@ def proposal_view(profile, space, proposal):
...
     if not proposal.state.DRAFT:
-        statusform = ProposalStatusForm(status=proposal.state.value)
-    else:
-        statusform = None
+        transitionform = ProposalTransitionForm()
+        # fill up the form choices with the given proposal's transitions
+        transitionform.populate_transitions(proposal)
+
+    # considering the user an admin of the proposal if they have either one of the roles
+    user_is_admin = 'admin' in proposal.roles_for(g.user)
```
shreyas-satish: Define this as a method on `Proposal`? So that it's `proposal.is_admin(g.user)`.

funnel/models/proposal.py
```diff
@@ -164,7 +165,82 @@ def _validate_proposal_space(self, key, value):
                 redirect.proposal = self
         return value
 
-    @state.transition(None, state.DELETED)
+    @with_roles(call={'admin'})
+    # XXX: Added `type` metadata just to decide color on the front end button. Needed?
+    @state.transition(state.AWAITING_DETAILS, state.DRAFT, title='Draft it', type='danger')
```
shreyas-satish: I think just `Draft` should suffice for the `title`.

funnel/models/proposal.py
```diff
@@ -273,6 +349,21 @@ def permissions(self, user, inherited=None):
                     perms.add('decline-proposal')  # Decline speaking
         return perms
 
+    def roles_for(self, user=None, token=None):
+        if not user and not token:
+            return set()
+        roles = super(Proposal, self).roles_for(user, token)
+        if self.speaker and self.speaker == user:
+            roles.add('author')
```
shreyas-satish: Use `speaker` since it's more obvious?

funnel/models/proposal.py
```diff
@@ -273,6 +349,21 @@ def permissions(self, user, inherited=None):
...
+        if self.user == user:
+            roles.add('creator')
```
shreyas-satish: Use `proposer`?

funnel/models/proposal.py
```diff
@@ -164,7 +165,82 @@ def _validate_proposal_space(self, key, value):
...
+    @state.transition(state.SHORLISTABLE, state.SHORTLISTED_FOR_REHEARSAL, title='Shortlist for rehearsal', type='success')
+    def shortlist_for_rehearsal(self):
+        pass
+
+    @with_roles(call={'admin'})
+    @state.transition(state.SHORTLISTED_FOR_REHEARSAL, state.REHEARSAL, title='Let\'s rehearse!', type='success')
+    def rehearsal(self):
+        pass
+
+    state.add_state_group('DELETABLE', state.DRAFT, state.SUBMITTED)
```
shreyas-satish: Can you move these `state.add_state_group` calls to the top (beneath the column names) and place them together, so it's easy to scan for which groups are available?

funnel/models/proposal.py
```diff
@@ -164,7 +165,82 @@ def _validate_proposal_space(self, key, value):
...
+    @with_roles(call={'admin'})
+    @state.transition(state.CONFIRMED, state.SUBMITTED, type='danger')
+    def unconfirm(self):
+        pass
+
+    state.add_state_group('WAITLISTABLE', state.SUBMITTED, state.CONFIRMED,
+                          state.SHORTLISTED, state.REJECTED, state.CANCELLED)
+
+    @with_roles(call={'admin'})
+    @state.transition(state.WAITLISTABLE, state.WAITLISTED, title='Put in waitlist', type='primary')
```
shreyas-satish: Just `Waitlist` for the `title`.

funnel/models/proposal.py
```diff
@@ -164,7 +165,82 @@ def _validate_proposal_space(self, key, value):
...
+    def reject(self):
+        pass
+
+    @with_roles(call={'admin'})
+    @state.transition(state.SUBMITTED, state.CANCELLED, type='danger')
+    def cancel(self):
+        pass
+
+    @with_roles(call={'admin'})
+    @state.transition(state.SUBMITTED, state.AWAITING_DETAILS, title='Wait for details', type='primary')
```
shreyas-satish: `Awaiting details` for the `title`.

funnel/models/proposal.py
```diff
@@ -164,7 +165,82 @@ def _validate_proposal_space(self, key, value):
...
+    def awaiting_details(self):
+        pass
+
+    state.add_state_group('EVALUATEABLE', state.SUBMITTED, state.AWAITING_DETAILS)
+
+    @with_roles(call={'admin'})
+    @state.transition(state.EVALUATEABLE, state.UNDER_EVALUATION, title='Evaluate', type='success')
```
shreyas-satish: `Under evaluation` for the `title`.

funnel/models/proposal.py
```diff
@@ -164,7 +165,82 @@ def _validate_proposal_space(self, key, value):
...
+    state.add_state_group('SHORLISTABLE', state.SUBMITTED,
+                          state.AWAITING_DETAILS, state.UNDER_EVALUATION)
+
+    @with_roles(call={'admin'})
+    @state.transition(state.SHORLISTABLE, state.SHORTLISTED_FOR_REHEARSAL, title='Shortlist for rehearsal', type='success')
+    def shortlist_for_rehearsal(self):
+        pass
+
+    @with_roles(call={'admin'})
+    @state.transition(state.SHORTLISTED_FOR_REHEARSAL, state.REHEARSAL, title='Let\'s rehearse!', type='success')
```
shreyas-satish: `Rehearsal` for the `title`.

iambibhas: Sure. These are supposed to be actions. Hence I tried to use them as verb.

funnel/forms/proposal.py
```diff
@@ -74,10 +74,15 @@ def __init__(self, *args, **kwargs):
...
-        __("Status"), coerce=int,
-        choices=[(status, label.title) for (status, label) in PROPOSALSTATUS.items() if status != PROPOSALSTATUS.DRAFT])
+class ProposalTransitionForm(forms.Form):
+    transition = forms.SelectField(__("Status"), validators=[forms.validators.DataRequired()])
+
+    def populate_transitions(self, proposal):
+        # value: transition method name
+        # label: method name in title case, after removing underscores
+        self.transition.choices = sorted([
+            (tname, tname.replace("_", " ").title()) for (tname, t) in proposal.state.transitions.iteritems()
```
shreyas-satish: Use `iterkeys` since you're only using the key?

funnel/forms/proposal.py
```diff
@@ -74,10 +74,15 @@ def __init__(self, *args, **kwargs):
             self.description.description = space.proposal_part_b.get('hint')
 
 
-class ProposalStatusForm(forms.Form):
-    status = forms.SelectField(
-        __("Status"), coerce=int,
-        choices=[(status, label.title) for (status, label) in PROPOSALSTATUS.items() if status != PROPOSALSTATUS.DRAFT])
+class ProposalTransitionForm(forms.Form):
+    transition = forms.SelectField(__("Status"), validators=[forms.validators.DataRequired()])
```
shreyas-satish: Add a validation to check if the supplied choice is a valid transition.

shreyas-satish: The `getattr` call below will fail if it isn't.

funnel/forms/proposal.py
iambibhas: @shreyas-satish Need a key and value for choices. And then wtforms uses this to validate the form.

funnel/forms/proposal.py
iambibhas: @shreyas-satish the validation is done with the choices. It's already happening because I'm filling up the choices.

funnel/forms/proposal.py
shreyas-satish: `(tname, tname.replace("_", " ").title()) for tname in proposal.state.transitions.iterkeys()` won't work?

funnel/forms/proposal.py
shreyas-satish: That's assuming a user submits the form through the user interface. In this context this will likely not be a problem, but generally it's safer to not trust user input.

funnel/forms/proposal.py
iambibhas: Oh right. yeah, that'll work.

funnel/forms/proposal.py
shreyas-satish: > It's already happening because I'm filling up the choices.

That isn't a validation. That's just pre-populating an input element with choices. A user can still supply a value through JavaScript that isn't listed in the choices.

funnel/models/proposal.py
```diff
@@ -273,6 +343,26 @@ def permissions(self, user, inherited=None):
...
+        roles = super(Proposal, self).roles_for(user, token)
+        if self.speaker and self.speaker == user:
+            roles.add('speaker')
+        if self.user == user:
+            roles.add('proposer')
+        if self.proposal_space.admin_team in user.teams or self.proposal_space.profile.admin_team in user.teams:
+            roles.add('admin')
+        if self.proposal_space.review_team in user.teams:
+            roles.add('reviewer')
+            roles.add('admin')
```
shreyas-satish: Are you sure you want to add the `admin` role here? There won't be any difference between a user in the `review_team` and a user in the `admin_team` in that case.

funnel/templates/proposal.html.jinja2
```diff
@@ -211,9 +206,25 @@
...
-            <p>{{ renderform(statusform, 'proposal-status', action=proposal.url_for('status'), submit='Save') }}</p>
+              <div class="col-md-6 col-lg-6 col-sm-12 col-xs-12 pull-right">
+                <h3>Status:
+                    {{ proposal.state.label.title }}
+                    {% if proposal.session and proposal.state.CONFIRMED %} &amp; <mark><a class="" href="{{ proposal.session.url_for() }}">Scheduled</a></mark>{% endif %}
+                </h3>
+                <ul class="list-inline">
+                {% for name, transition in proposal.state.transitions.iteritems() %}
+                  <li>
+                    <form action="{{ proposal.url_for('transition') }}" method="post" class="form-inline">
```
shreyas-satish: We aren't using `render_form` because of styling constraints?

iambibhas: @shreyas-satish renderform renders SelectField as dropdown. We need buttons on the front end as they are actions.

funnel/models/proposal.py
iambibhas: @shreyas-satish right. my bad. Question: Should a reviewer be able to change state of proposals? Right now they can.

shreyas-satish: > Should a reviewer be able to change state of proposals?
Yes.

funnel/views/proposal.py
```diff
@@ -163,33 +163,25 @@ def proposal_edit(profile, space, proposal):
...
+    transitionform = ProposalTransitionForm()
+    # fill up the form choices with the given proposal's transitions for validation
+    transitionform.populate_transitions(proposal)
+
+    if transitionform.validate_on_submit():  # check if the provided transition is valid
+        getattr(proposal, transitionform.transition.data)()  # call the transition
         db.session.commit()
-        flash(_("The proposal's status has changed to ") +
-              proposal.state.label.title, 'success')
+        flash(_("The proposal's status has changed to ") + proposal.state.label.title, 'success')
```
jace: We will now describe transitions rather than states. For all transitions, add a parameter `message='…'`. Over here, we can flash `getattr(proposal.access_for(current_auth.user), transitionform.transition.data).message`

Since we're accessing the transition object twice, might as well do `transition = getattr(…)` right after form validation.

funnel/views/proposal.py
```diff
@@ -294,19 +287,24 @@ def proposal_view(profile, space, proposal):
...
+        transitionform = ProposalTransitionForm()
+        # fill up the form choices with the given proposal's transitions
+        transitionform.populate_transitions(proposal)
+
+    # considering the user an admin of the proposal if they have either one of the roles
+    user_is_admin = proposal.is_admin(g.user) or proposal.is_reviewer(g.user)
 
     proposal_move_form = None
-    if 'move-proposal' in space.permissions(g.user, g.permissions):
+    if user_is_admin:
```
jace: This is not wise. Don't assume permissions are granted because of a blanket check.

funnel/views/proposal.py
```diff
@@ -294,19 +287,24 @@ def proposal_view(profile, space, proposal):
...
         proposal_move_form = ProposalMoveForm()
 
     return render_template('proposal.html.jinja2', space=space, proposal=proposal,
         comments=comments, commentform=commentform, delcommentform=delcommentform,
-        votes_groups=proposal.votes_by_group(),
-        links=links, statusform=statusform, proposal_move_form=proposal_move_form,
+        votes_groups=proposal.votes_by_group(), user_is_admin=user_is_admin,
```
jace: Reviewers are not admins.

funnel/models/proposal.py
```diff
@@ -164,7 +165,76 @@ def _validate_proposal_space(self, key, value):
...
+        pass
+
+    @with_roles(call={'admin'})
+    @state.transition(state.SHORLISTABLE, state.SHORTLISTED_FOR_REHEARSAL, title='Shortlist for rehearsal', type='success')
+    def shortlist_for_rehearsal(self):
+        pass
+
+    @with_roles(call={'admin'})
+    @state.transition(state.SHORTLISTED_FOR_REHEARSAL, state.REHEARSAL, title='Rehearsal', type='success')
+    def rehearsal(self):
```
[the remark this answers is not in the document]
iambibhas: From the state list, `rehearsal` means `Rehearsal ongoing`.

funnel/templates/proposal.html.jinja2
```diff
@@ -211,9 +201,22 @@
                   {% endfor %}
                 {% endif %}
               {% endfor %}
-            </div>
 
-            <p>{{ renderform(statusform, 'proposal-status', action=proposal.url_for('status'), submit='Save') }}</p>
+              <div class="col-md-6 col-lg-6 col-sm-12 col-xs-12 pull-right">
+                <h3>Status:
+                    {{ proposal.state.label.title }}
+                    {% if proposal.state.SCHEDULED %} &amp; <mark><a class="" href="{{ proposal.session.url_for() }}">Scheduled</a></mark>{% endif %}
```
jace: This also needs a similar fix as above (to be added in Coaster).

funnel/templates/proposal.html.jinja2
```diff
@@ -211,9 +201,22 @@
...
+                </h3>
+                <form action="{{ proposal.url_for('transition') }}" method="post" class="form-inline">
+                  {{ csrf_form.hidden_tag() }}
+                  <p>
+                  {% for name, transition in proposal.state.transitions.iteritems() %}
```
jace: Avoid `iteritems` since it's removed in Python 3. Use the regular `items`.

funnel/views/proposal.py
```diff
@@ -12,10 +12,10 @@
 from baseframe import _
 from baseframe.forms import render_form, render_delete_sqla, Form
 
-from .. import app, mail, lastuser
+from .. import app, mail, lastuser, current_auth
```
jace: Import `current_auth` from `coaster.auth`.

funnel/views/proposal.py
```diff
@@ -294,19 +285,19 @@ def proposal_view(profile, space, proposal):
...
+    transitionform = None
     if not proposal.state.DRAFT:
-        statusform = ProposalStatusForm(status=proposal.state.value)
-    else:
-        statusform = None
+        transitionform = ProposalTransitionForm(obj=proposal)
 
     proposal_move_form = None
-    if 'move-proposal' in space.permissions(g.user, g.permissions):
+    if proposal.is_admin(current_auth.user):
```
jace: `if proposal.current_roles.admin:`

funnel/views/proposal.py
```diff
@@ -294,19 +285,19 @@ def proposal_view(profile, space, proposal):
                 flash(_("No such comment"), 'error')
             return redirect(proposal.url_for(), code=303)
     links = [Markup(linkify(unicode(escape(l)))) for l in proposal.links.replace('\r\n', '\n').split('\n') if l]
+
+    transitionform = None
     if not proposal.state.DRAFT:
```
jace: Remove. Even drafts get a submit transition.

funnel/forms/proposal.py
```diff
@@ -74,10 +74,13 @@ def __init__(self, *args, **kwargs):
...
+
+    def set_queries(self):
+        # value: transition method name
+        # label: method name in title case, after removing underscores
+        self.transition.choices = sorted(self.edit_obj.state.transitions.items())
```
jace: Don't you need this?
```python
self.transition.choices = sorted(
    [(name, transition.data['title']) for name, transition in self.edit_obj.state.transitions.items()])
```

funnel/models/proposal.py
```diff
@@ -164,7 +171,71 @@ def _validate_proposal_space(self, key, value):
                 redirect.proposal = self
         return value
 
-    @state.transition(None, state.DELETED)
+    # State transitions
+    state.add_conditional_state('SCHEDULED', state.CONFIRMED, lambda proposal: proposal.session is not None)
```
jace: After hasgeek/coaster#165, you'll need to add `label=('scheduled', __("Confirmed & Scheduled"))`

funnel/templates/proposal.html.jinja2
```diff
@@ -121,26 +121,16 @@
...
-            {% endif %}
-          {% endif %}
-          {% if proposal.session %}
-            <a class="btn btn-xs btn-info" href="{{ proposal.session.url_for() }}">View session in schedule</a>
-          {% endif %}
-          {%- if 'move-proposal' in g.permissions %}
-            <br>
-            {{ renderform(proposal_move_form, 'proposal-move', action=proposal.url_for('move-to'), submit='Move', style=None) }}
-          {%- endif %}
+          Status: {% if proposal.state.SCHEDULED %}Confirmed and scheduled{% elif proposal.state.CONFIRMED %}Confirmed{% else %}Submitted{% endif %}
```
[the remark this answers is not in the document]
jace: Coming in hasgeek/coaster#165.

funnel/templates/proposal.html.jinja2
jace: Coming in hasgeek/coaster#165.

funnel/views/proposal.py
```diff
@@ -294,19 +286,17 @@ def proposal_view(profile, space, proposal):
...
-    if not proposal.state.DRAFT:
-        statusform = ProposalStatusForm(status=proposal.state.value)
-    else:
-        statusform = None
+
+    transitionform = ProposalTransitionForm(obj=proposal)
 
     proposal_move_form = None
-    if 'move-proposal' in space.permissions(g.user, g.permissions):
+    if proposal.current_roles.admin:
```
jace: This thing is bothering me a bit. Neither the old permission model nor the role model adequately describes what is needed. Temporary fix (still incomplete):

1. Add a `move_to` method to the object. It takes a new parent and moves to it, but does not validate if the user has permission to do this.
2. Decorate this with `@with_roles(call={'admin'})`
3. On this line here, check `if 'move_to' in proposal.current_access():`
4. After form validation, call `proposal.current_access().move_to(form.new_parent.data)`
5. TODO: Find a way to declaratively associate `ProposalMoveForm` with this method, mapping form fields to method parameters.

migrations/versions/3c47ba103724_proposal_status.py
```diff
@@ -13,16 +13,16 @@
 from alembic import op
 import sqlalchemy as sa
 from sqlalchemy.sql import table, column
-from funnel.models import PROPOSALSTATUS
+from funnel.models import PROPOSAL_STATE
```
jace: Why is a migration importing from code? Make a copy of `PROPOSALSTATUS` here (and retain the old name, since it's an old migration).

iambibhas: The migration needed the states dict I guess. Really old migration.  I'll do that.

funnel/views/proposal.py
iambibhas: Should I do this in this PR? or a separate one?

jace: We can't merge this PR until we review transitions with editorial, so you might as well incorporate these.

funnel/models/proposal.py
```diff
@@ -273,6 +345,17 @@ def permissions(self, user, inherited=None):
                     perms.add('decline-proposal')  # Decline speaking
         return perms
 
+    # Roles
+
+    def roles_for(self, actor=None, anchors=()):
+        roles = super(Proposal, self).roles_for(actor, anchors)
+        if self.speaker and self.speaker == actor:
+            roles.add('speaker')
+        if self.user == actor:
```
jace: Make this `if self.user and self.user == actor`. While there is no condition under which `user` will be None _currently_, don't be caught by surprise when it comes up in future.

We made this mistake in Lastuser some years ago: https://github.com/hasgeek/lastuser/commit/17aaa61ff0b06531dc1bd17b7726ad0b680cbcad

funnel/models/space.py
```diff
@@ -329,6 +329,16 @@ def fetch_sorted(cls):
...
 
+    def roles_for(self, actor=None, anchors=()):
+        roles = super(ProposalSpace, self).roles_for(actor, anchors)
+        if actor is not None:
+            if self.admin_team in actor.teams:
+                roles.add('admin')
+            if self.review_team in actor.teams:
+                roles.add('reviewer')
+        roles.update(self.profile.roles_for(actor, anchors))
+        return roles
```
jace: Please add a `reader` role here (available to all; [see notes](http://coaster.readthedocs.io/en/latest/sqlalchemy/roles.html)). In future we'll have private events where you must explicitly be granted a `reader` role.

funnel/templates/proposal.html.jinja2
```diff
@@ -121,26 +121,16 @@
...
-            {% endif %}
-          {% endif %}
-          {% if proposal.session %}
-            <a class="btn btn-xs btn-info" href="{{ proposal.session.url_for() }}">View session in schedule</a>
-          {% endif %}
-          {%- if 'move-proposal' in g.permissions %}
-            <br>
-            {{ renderform(proposal_move_form, 'proposal-move', action=proposal.url_for('move-to'), submit='Move', style=None) }}
-          {%- endif %}
+          Status: {% if proposal.state.SCHEDULED %}{{ proposal.state.SCHEDULED.label }}{% elif proposal.state.CONFIRMED %}{{ proposal.state.CONFIRMED.label }}{% else %}Submitted{% endif %}
```
jace: This can be reduced to `{% if proposal.state.CONFIRMED %}{{ proposal.state.label }}{% else %}{% trans %}Submitted{% endtrans %}{% endif %}`

funnel/models/space.py
jace: Also, in `Proposal.roles_for`, explicitly revoke the `reader` role if the state is DRAFT.

funnel/views/proposal.py
```diff
@@ -398,16 +388,12 @@ def proposal_prev(profile, space, proposal):
     permission='move-proposal', addlperms=lastuser.permissions)
 def proposal_moveto(profile, space, proposal):
     proposal_move_form = ProposalMoveForm()
-    if not proposal_move_form.validate_on_submit():
+    if proposal_move_form.validate_on_submit():
+        target_space = proposal_move_form.target.data
+        if target_space != proposal.proposal_space:
+            proposal.current_access().move_to(target_space)
```
jace: Hang on, isn't there a ProposalRedirect model that must be created to manage redirects from the old URL?

iambibhas: @jace ummm. I'm not sure. First time I'm hearing about it. Previous code didn't have any mention of that.

jace: So there is no ProposalRedirect model in the code? We need to add it then. Separate PR.

iambibhas: It's mentioned only at the decorator of this view. I need to read up if that does what it's supposed to.

jace: It does. It's added in the old location and points to the proposal, so the browser can.be redirected. Question is, where is the code to add it?

iambibhas: @jace I think here it is - https://github.com/hasgeek/funnel/pull/220/files#diff-7a3dcb6289f76819c08604a8f03c0e72L154

funnel/forms/proposal.py
```diff
@@ -74,10 +74,18 @@ def __init__(self, *args, **kwargs):
...
+    transition = forms.SelectField(__("Status"), validators=[forms.validators.DataRequired()])
+
+    def set_queries(self):
+        """
+        value: transition method name
+        label: transition object itself
+        We need the whole object to get the additional metadata in templates
+        """
+        self.transition.choices = sorted([
+            (name, transition) for name, transition in self.edit_obj.state.transitions.items() if hasattr(self.edit_obj.current_access(), name)
```
iambibhas: @jace I dont know if it's an issue or design, but `edit_obj.state.transitions.items()` is showing transitions the current user role doesn't have access to. E.g. `awaiting_details` is available to only `reviewer`, but `admin` can still see it in items(). But of course `current_access()` doesn't have it.

jace: This is a design issue. RoleMixin and StateManager aren't aware of each other, so they don't offer any way to obtain just the intersection. My guess is StateManager should offer an additional `transitions_for` method.

iambibhas: Ok. So can we go with this for now? And update it once the changes in statemanager is made?

funnel/templates/proposal.html.jinja2
```diff
@@ -211,9 +201,25 @@
...
+                </h3>
+                <form action="{{ proposal.url_for('transition') }}" method="post" class="form-inline">
+                  {{ transitionform.hidden_tag() }}
+                  <p>
+                  {% for name, transition in transitionform.transition.choices %}
+                    <button name="transition" value="{{ name }}" class="transition control-group btn btn-{{ transition.data['type'] }}">{{ transition.data['title'] }}</button>
+                  {% endfor %}
+                  </p>
+                </form>
+            </div>
```
shreyas-satish: Fix indent.

PR #222, opened 2018-02-21
Using tuple for label for conditional state

iambibhas:


funnel/models/proposal.py
```diff
@@ -172,7 +172,7 @@ def _validate_proposal_space(self, key, value):
         return value
 
     # State transitions
-    state.add_conditional_state('SCHEDULED', state.CONFIRMED, lambda proposal: proposal.session is not None, label=__("Confirmed & Scheduled"))
+    state.add_conditional_state('SCHEDULED', state.CONFIRMED, lambda proposal: proposal.session is not None, label=('confirmed_scheduled', __("Confirmed & Scheduled")))
```
jace: The name can be just `scheduled`, since `confirmed` is implicit.

PR #223, opened 2018-02-22
Specifically showing title of the state label in template

iambibhas:
For SCHEDULED state(and maybe others as well, dont have admin to check right now) the template is showing the namedtuple repr string.

image


jace:
Good to merge. For minor fixes like this, a PR is not necessary.

PR #224, opened 2018-02-22
Transitions for ProposalSpace and modified transitions for proposals

iambibhas:


funnel/models/proposal.py
```diff
@@ -123,11 +127,11 @@ class Proposal(BaseScopedIdNameMixin, CoordinatesMixin, db.Model):
     state = StateManager('_state', PROPOSAL_STATE, doc="Current state of the proposal")
 
     votes_id = db.Column(db.Integer, db.ForeignKey('votespace.id'), nullable=False)
-    votes = db.relationship(VoteSpace, uselist=False,
+    votes = db.relationship(VoteSpace, uselist=False, lazy='joined',
                             cascade='all, delete-orphan', single_parent=True)
 
     comments_id = db.Column(db.Integer, db.ForeignKey('commentspace.id'), nullable=False)
-    comments = db.relationship(CommentSpace, uselist=False,
+    comments = db.relationship(CommentSpace, uselist=False, lazy='joined',
```
iambibhas: These 3 lazies are to improve funnel loading time.

iambibhas:
Added transitions to proposal space. 

selection_262

funnel/forms/space.py
```diff
@@ -77,6 +69,21 @@ def validate_bg_color(self, field):
...
+                                   forms.validators.DataRequired()])
+
+    def set_queries(self):
+        """
+        value: transition method name
+        label: transition object itself
+        We need the whole object to get the additional metadata in templates
+        """
+        self.transition.choices = sorted([
+            (name, transition) for name, transition in self.edit_obj.state.transitions().items()
```
jace: Why can't this be just `self.transition.choices = sorted(self.edit_obj.state.transitions().items())`

funnel/models/proposal.py
```diff
@@ -93,7 +97,7 @@ class Proposal(BaseScopedIdNameMixin, CoordinatesMixin, db.Model):
         backref=db.backref('proposals', cascade="all, delete-orphan"))
 
     speaker_id = db.Column(db.Integer, db.ForeignKey('user.id'), nullable=True)
-    speaker = db.relationship(User, primaryjoin=speaker_id == User.id,
+    speaker = db.relationship(User, primaryjoin=speaker_id == User.id, lazy='joined',
```
jace: Under what conditions do we _not_ want to load a speaker?

funnel/models/proposal.py
jace: Agree on votes and comments, [but you need `lazy='select'` since `'joined'` does an eager load](http://docs.sqlalchemy.org/en/latest/orm/relationship_api.html#sqlalchemy.orm.relationship.params.lazy).

funnel/models/proposal.py
```diff
@@ -200,45 +209,45 @@ def waitlist(self):
...
-    def rehearsal_ongoing(self):
-        pass
-
-    @with_roles(call={'admin', 'speaker', 'proposer'})
+    @with_roles(call={'speaker', 'proposer'})
     @state.transition(state.DELETABLE, state.DELETED, title=__("Delete"), message=__("This proposal has been deleted"), type='danger')
     def delete(self):
         pass
 
+    # @with_roles(call={'admin'})
```
jace: Leave the same note here about why these states have been removed. Ideally, issue a migration to move any proposals having this state into a more relevant state, and just remove all this code (separate PR).

funnel/models/proposal.py
iambibhas: @jace `select` seems like the default behavior, where it emits a select query when needed, but for the same reason, when loading funnel front page, for rootconf locally, it's making >300 queries right now. maybe I need to move these lazies and try to do joinedload while loading the front page?

PR #226, opened 2018-04-02
Copying proposal if the existing proposal already has a session

iambibhas:


funnel/views/proposal.py
```diff
@@ -390,7 +390,10 @@ def proposal_moveto(profile, space, proposal):
     if proposal_move_form.validate_on_submit():
         target_space = proposal_move_form.target.data
         if target_space != proposal.proposal_space:
-            proposal.current_access().move_to(target_space)
+            if proposal.state.MOVABLE:
+                proposal.current_access().move_to(target_space)
+            else:
+                proposal = proposal.current_access().copy_to(target_space)
```
jace: This has to be in the UI as well: don't show "Move to" if you can only make a copy.

jace (changes requested):

jace:
@iambibhas When you make a copy, "user" must be the person who made the copy. "speaker" can remain as at present. We have to record the person who requested the copy operation.

jace:
Can you clean this up for merger?

iambibhas:
Pushed some fixes. But UI has some serious issue. /cc @vidya-ram 
funnel-issue

jace:
What should we do with this?

jace:
As per #327, proposals are no longer attached to projects. There is a separate `ProposalSubmission` instead.

jace:
@iambibhas this is the issue that broke for Zainab last week. Let's close this?

jace (changes requested):
This has a few issues.

jace:
This is now severely out of date, so we'll have to redo this as a new PR.

PR #253, opened 2018-06-30
Supporting contact exchange creation using POST request + bug fix

iambibhas:


funnel/forms/participant.py
```diff
@@ -27,3 +28,20 @@ class ParticipantForm(forms.Form):
...
+
+
+class ParticipantContactExchangeForm(forms.Form):
+    puk = forms.StringField(__("Public key"), validators=[forms.validators.DataRequired()])
+    key = forms.StringField(__("Private key"), validators=[forms.validators.DataRequired()])
+
+    class Meta:
+        # Disable CSRF as this form will most likely be used over API
+        # XXX: Do we need to disable CSRF for this?
+        csrf = False
```
jace: Don't do this. Instantiate the form with CSRF disabled.

jace: http://flask-wtf.readthedocs.io/en/latest/form.html#module-flask_wtf

funnel/forms/participant.py
```diff
@@ -27,3 +28,20 @@ class ParticipantForm(forms.Form):
...
+
+    def validate_key(form, field):
+        form.edit_obj = Participant.query.filter_by(puk=form.puk.data, proposal_space=form.edit_parent).first()
```
jace: Make this `form.participant` and declare it in `__returns__` (pending ticket: https://github.com/hasgeek/baseframe/issues/112)

funnel/views/participant.py
```diff
@@ -115,7 +115,7 @@ def participant_edit(profile, space, participant):
     return render_form(form=form, title=_(u"Edit Participant"), submit=_(u"Save changes"))
 
 
-@app.route('/<space>/participant', subdomain='<profile>')
+@app.route('/<space>/participant', methods=['POST'], subdomain='<profile>')
```
jace: This will break current Android apps. Keep both GET and POST and leave a comment that GET is deprecated. We'll need to set a reminder to remove it in a few months.

funnel/views/participant.py
```diff
@@ -125,10 +125,9 @@ def participant(profile, space):
     if space.date_upto:
         if midnight_to_utc(space.date_upto + timedelta(days=1), space.timezone, naive=True) < datetime.utcnow():
             return jsonify(message=u"This event has concluded", code=401)
-    participant = Participant.query.filter_by(puk=request.args.get('puk')).first()
-    if not participant:
-        return jsonify(message=u"Not found", code=404)
-    elif participant.key == request.args.get('key'):
+    form = ParticipantContactExchangeForm(parent=space)
```
jace: If you keep support for `GET`, you can't use a form. Since the form doesn't actually do anything more than a query, use `@requestargs` on this view instead, and just put the query in the view. Form is unnecessary.

funnel/views/participant.py
```diff
@@ -115,20 +115,26 @@ def participant_edit(profile, space, participant):
...
     (Profile, {'name': 'profile'}, 'g.profile'),
     ((ProposalSpace, ProposalSpaceRedirect), {'name': 'space', 'profile': 'profile'}, 'space'),
     permission='view')
-def participant(profile, space):
+@requestargs('puk', 'key')
+def participant(profile, space, puk, key):
+    """
+    Endpoint for contact exchange.
+
+    Notice: The GET method to this endpoint is deprecated and will be removed by 1st September, 2018
```
jace: Use `TODO` so it can be found later. Or use a Sphinx deprecated tag. We'll forget about this notice.

iambibhas: Doing that. I anyway added a reminder for myself at end of august just in case.

PR #254, opened 2018-07-11
Second funnel app

### LINKED ISSUE #230: Add a second Flask app
To support the transition from Talkfunnel.com to HasGeek.com, a second Flask app should be added to Funnel, to support a different set of routes for the HasGeek website:

1. [x] Rename `app` to `funnelapp` since Talkfunnel will be the legacy site
2. [x] Add `runfunnel.py` that imports `funnelapp` as `application` instead of `app`
3. [x] Likewise, `funnelwebsite.py` for WSGI
4. [x] All modules importing app should import `app, funnelapp`
5. [x] All _current_ instances of `@app.route` should be replaced with `@funnelapp.route`
6. [x] A new set of `@app.route` rules should be added to the same views, with `<profile>.website.tld/<space>` becoming `website.tld/<profile>/<space>` (originally introduced in #102)
7. [x] Disable the `/new` route for `funnelapp`
8. [x] All run-time references to `app` should change to `current_app`
9. [x] Audit for extensions that can't handle multiple apps (Flask-Lastuser can't handle multiple Lastuser key/secret pairs, and Lastuser will be fussy about the website URL, so a ticket to fix that is necessary, maybe by accepting config as `init_app` parameters instead of retrieving from `app.config`)
10. [x] The layout template should render a different header (HasGeek or Talkfunnel logo) depending on which app is rendering, so this flag should be inspectable somewhere
11. [x] Profiles should be linked to just one of the apps via a flag (existing profiles go to Talkfunnel, new profiles go to HasGeek, and existing HasGeek profiles are migrated manually). This requires a migration
12. [x] If a URL is accessed from the wrong app, it should be redirected to the corresponding endpoint at the other app, so a before_request handler is required to intercept and examine (but can it return responses?)

iambibhas:
First part of completing #230.


funnel/__init__.py
```diff
@@ -36,13 +37,47 @@
...
 db.app = app
+
+migrate = Migrate(funnelapp, db)
 migrate = Migrate(app, db)
+
+mail.init_app(funnelapp)
 mail.init_app(app)
+
+lastuser.init_app(funnelapp)
 lastuser.init_app(app)
```
iambibhas: @jace these lines above seems wrong. specially the `db.app = <>` part. How to handle these?

jace: `db.app` assignment is only required if there's a database operation outside a request context, to let the db object know which app to operate on. This can happen during init, if we're loading from db, or doing a background job. Workaround: enclose the database operation in `with app.app_context():` (we've previously used `with app.test_request_context():`, a somewhat heavier operation since it also constructs a dummy request).

iambibhas: Ok. And what about the others? Double assigning `migrate` and calling `init_app()` multiple times on `db`, `mail` and `lastuser` doesn't seem right.

funnel/views/profile.py
```diff
@@ -6,10 +6,11 @@
 from baseframe.forms import render_message, render_redirect, render_form
 from ..models import db, Profile, Team
 from ..forms import NewProfileForm, EditProfileForm
-from .. import app, lastuser
+from .. import funnelapp, app, lastuser
 
 
 @app.route('/new', methods=['GET', 'POST'])
+@funnelapp.route('/new', methods=['GET', 'POST'])
```
jace: Remove `@funnelapp.route`. We're removing support for adding new profiles in Talkfunnel.

funnel/views/proposal.py
```diff
@@ -379,7 +388,8 @@ def proposal_prev(profile, space, proposal):
         return redirect(space.url_for())
 
 
-@app.route('/<space>/<proposal>/move', methods=['POST',], subdomain='<profile>')
+@app.route('/<profile>/<space>/<proposal>/move', methods=['POST',])
+@funnelapp.route('/<space>/<proposal>/move', methods=['POST',], subdomain='<profile>')
```
jace: `methods=['POST']` in both lines (for style consistency).

funnel/views/usergroup.py
```diff
@@ -31,12 +33,12 @@ def usergroup_view(profile, space, usergroup):
...
-    endpoint='usergroup_new',
-    methods=['GET', 'POST'],
-    subdomain='<profile>')
-@app.route('/<space>/users/<group>/edit', methods=['GET', 'POST'], subdomain='<profile>')
+@app.route('/<profile>/<space>/users/new', defaults={'group': None}, endpoint='usergroup_new',
+    methods=['GET', 'POST'])
+@funnelapp.route('/<space>/users/new', defaults={'group': None}, endpoint='usergroup_new',
+    methods=['GET', 'POST'], subdomain='<profile>')
+@app.route('/<profile>/<space>/users/<group>/edit', methods=['GET', 'POST'])
+@funnelapp.route('/<space>/users/<group>/edit', methods=['GET', 'POST'], subdomain='<profile>')
```
jace: Order them `app`, `app`, `funnelapp`, `funnelapp`.

manage.py
```diff
@@ -2,7 +2,7 @@
 
 from coaster.manage import init_manager
 
-from funnel import app, models
+from funnel import funnelapp, app, models
```
jace: `funnelapp` is not required here.

rqfunnelinit.py
```diff
@@ -0,0 +1,14 @@
+from urlparse import urlparse
```
jace: This file is not required. The same RQ backend will power both apps.

runserver.py
```diff
@@ -6,5 +6,5 @@
 try:
     port = int(sys.argv[1])
 except (IndexError, ValueError):
-    port = 3000
+    port = 3001
```
jace: Keep the port here, change it for Funnel.

funnel/jobs/jobs.py
```diff
@@ -1,11 +1,11 @@
+from flask import current_app
 from ..models import (db, TicketClient)
 from ..extapi.explara import ExplaraAPI
 from ..extapi.boxoffice import Boxoffice
-from funnel import app
 
 
 def import_tickets(ticket_client_id):
-    with app.test_request_context():
+    with current_app.test_request_context():
```
jace: Change this back to `app`. If you need a `test_request_context`, then `current_app` will be `None`, so this construct will fail. Do this everywhere that `test_request_context` or `app_context` is used.

funnel/models/profile.py
```diff
@@ -16,6 +17,22 @@ class Profile(ProfileBase, db.Model):
...
+            },
+        }
+
+    @cached_property
+    def parent_spaces(self):
+        from .space import ProposalSpace
+        spaces_all = ProposalSpace.fetch_sorted().filter(
+            ProposalSpace.profile == self, ProposalSpace.parent_space == None
+        ).all()
+        return spaces_all
```
jace: This should be a relationship. Why is it a property?

funnel/views/space.py
```diff
@@ -38,7 +38,8 @@ def space_data(space):
 
 
 # Test endpoint
-@app.route('/form', methods=['GET', 'POST'], subdomain='<profile>')
+@app.route('/<profile>/form', methods=['GET', 'POST'])
+@funnelapp.route('/form', methods=['GET', 'POST'], subdomain='<profile>')
```
jace: Remove this endpoint. We added it to test Baseframe's form generator. It's not required anymore.

funnel/models/profile.py
iambibhas: Not sure why it was defined as a property, I just organized it. Anyway, now when I try to define it as a relationship, there is a circular import issue. `ProposalSpace` model requires `Profile` for a relationship and vice versa.

funnel/views/index.py
```diff
@@ -1,22 +1,33 @@
...
+
+
+@funnelapp.route('/', endpoint='index')
+@render_with({'text/html': 'funnelindex.html.jinja2'}, json=True)
+def funnelindex():
+    g.profile = None
+    g.permissions = []
+    spaces = ProposalSpace.fetch_sorted().filter(ProposalSpace.profile != None).all()
+    # XXX: Can we just return index() ?
+    return {'spaces': [d for d in [dict(s.current_access()) for s in spaces] if d]}
```
iambibhas: @jace ^

funnel/models/profile.py
```diff
@@ -16,6 +17,27 @@ class Profile(ProfileBase, db.Model):
 
     description = MarkdownColumn('description', default=u'', nullable=False)
 
+    parent_spaces_test = db.relationship('ProposalSpace',
+        lazy='dynamic',
+        primaryjoin="and_(ProposalSpace.profile_id == Profile.id, ProposalSpace.parent_space_id == None)"
+        )
```
iambibhas: @jace need some help with this one, `parent_spaces` below is returned by the `fetch_sorted()` classmethod which, along with sorting the results, [does a state filter](https://github.com/hasgeek/funnel/blob/f29f04004f348de3f8309a9f6b41927de7aa95d6/funnel/models/space.py#L379-L389), which I can't do where without importing `ProposalSpace`, which I can do as I removed the circular import issue from space.py, but want to check if it's possible without importing `ProposalSpace`

funnel/views/index.py
```diff
@@ -1,22 +1,33 @@
...
     spaces = ProposalSpace.fetch_sorted().filter(ProposalSpace.profile != None).all()
-    return render_template('index.html.jinja2', spaces=spaces)
+    return {'spaces': [d for d in [dict(s.current_access()) for s in spaces] if d]}
+
+
+@funnelapp.route('/', endpoint='index')
+@render_with({'text/html': 'funnelindex.html.jinja2'}, json=True)
+def funnelindex():
+    g.profile = None
+    g.permissions = []
```
iambibhas: @jace  `g.permissions` is widely used in this repo. I'm not too sure about the changes required to change them back to `current_auth.permissions`.  Can we handle this in a separate PR?

funnel/views/index.py
```diff
@@ -1,22 +1,37 @@
...
+
+
+@funnelapp.route('/', endpoint='index')
+@render_with({'text/html': 'funnelindex.html.jinja2', 'application/json': index_jsonify})
+def funnelindex():
+    g.profile = None
+    g.permissions = []
+    spaces = ProposalSpace.fetch_sorted().filter(ProposalSpace.profile != None).all()
+    # XXX: Can we just return index() ?
+    return {'spaces': spaces}
```
iambibhas: @jace can we just return index()?

funnel/views/index.py
```diff
@@ -1,22 +1,41 @@
...
     spaces = ProposalSpace.fetch_sorted().filter(ProposalSpace.profile != None).all()
-    return render_template('index.html.jinja2', spaces=spaces)
+    return dict(spaces=spaces)
+
+
+@funnelapp.route('/', endpoint='index')
+@render_with({'text/html': 'funnelindex.html.jinja2', 'application/json': index_jsonify})
+def funnelindex():
+    g.profile = None
+    g.permissions = []
```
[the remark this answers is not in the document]
iambibhas: `g.permissions` is used in a lot of places on this repo. I'm not completely sure about the effects of change to `current_auth.permissions`. can we do this part of another PR? or maybe the next PR.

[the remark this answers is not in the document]
jace: `current_auth.permissions` is guaranteed to be an `InspectableSet`. `g.permissions` only promises to support the `in` operator. It could be a list, tuple or set. This difference is crucial for the transition.

funnel/models/profile.py
jace: You can put the name in quotes. SQLAlchemy will do an `eval` using the model namespace to pick the correct model. That means any valid Python expression will work inside the quotes.

funnel/models/profile.py
jace: That entire function could be a SQL expression, which means it could be a relationship.

jace: Leave this aside for now. I'll look at rewriting it as a relationship later.

funnel/views/index.py
jace: You'll get the output of `render_with`, unless you pass it `_render=False`. https://coaster.readthedocs.io/en/latest/views/decorators.html#coaster.views.decorators.render_with

PR #264, opened 2018-09-25
UI fixes and changes to improve load time & performance

vidya-ram:
1. Add spacing between proposal status change buttons. 
2. Move prev/next arrows to top of the proposal page in smaller screens. 
3. Add support to collapse proposal categories(proposal space page) in admin view.
4. Optimise banner images. Change banner images to responsive images using `srcset`
5. Remove `Codemirror.js` and `Ractive.js` from `ext_requires`
5. Add `formlayout.html.jinja2` and include js_codemirrormarkdown and css_codemirrormarkdown
6. Include `js_ractive` in `event.html.jinja2` and `scan_badge.html.jinja` pages

<span>Collapsible categories in admin view</span>


funnel/__init__.py
```diff
@@ -60,8 +60,8 @@
 lastuser.init_usermanager(UserManager(db, models.User, models.Team))
 
 baseframe.init_app(app, requires=['funnel'], ext_requires=[
-    ('codemirror-markdown', 'pygments'), 'toastr', 'baseframe-mui', 'fontawesome>=4.0.0',
```
jace: pygments is required on proposal view pages. It renders syntax highlighting in code samples. Pygments and CodeMirror will not be used on the same page as they represent separate read and edit views.

vidya-ram: Moved Pygments back inside `ext_requires`

PR #277, opened 2018-11-12
Migrate db to be in sync with models

jace:


funnel/models/commentvote.py
```diff
@@ -64,10 +64,10 @@ def getvote(self, user):
 
 class Vote(BaseMixin, db.Model):
     __tablename__ = 'vote'
-    user_id = db.Column(db.Integer, db.ForeignKey('user.id'), nullable=False)
+    user_id = db.Column(None, db.ForeignKey('user.id'), nullable=False)
```
iambibhas: What does `None` do here? What will be the column type after migration?

iambibhas: What does `None` do here?

jace: It picks up the column type from the foreign key. Using `None` avoids errors where the types are mismatched, and so is part of our code style.

PR #279, opened 2018-11-13
Rename ProposalSpace to Project

jace:
Also cleans up various inconsistencies in models. I have reviewed every line to ensure there are no unintended changes, but I'll appreciate more eyes on this.

There is a CSS style named `.space` that I haven't touched. I don't know if it refers to a ProposalSpace or just some spacing in a div.


funnel/models/event.py
```diff
@@ -66,14 +66,14 @@ class Event(ScopedNameTitleMixin, db.Model):
     """
     __tablename__ = 'event'
 
-    proposal_space_id = db.Column(None, db.ForeignKey('proposal_space.id'), nullable=False)
-    proposal_space = db.relationship(ProposalSpace,
+    project_id = db.Column(None, db.ForeignKey('project.id'), nullable=False)
+    project = db.relationship(Project,
```
iambibhas: It's possible to define `relationship("Project")` without the need of importing Project in all models file. Should we make that a part of our dev guideline?

jace: It is possible, but SQLAlchemy uses `eval` to process it. It is safer to use the direct object reference.

funnel/models/profile.py
```diff
@@ -26,10 +26,10 @@ class Profile(ProfileBase, db.Model):
         }
 
     @cached_property
-    def parent_spaces(self):
-        from .space import ProposalSpace
-        spaces_all = ProposalSpace.fetch_sorted().filter(
-            ProposalSpace.profile == self, ProposalSpace.parent_space == None  # NOQA
+    def parent_projects(self):
+        from .project import Project
+        spaces_all = Project.fetch_sorted().filter(
```
iambibhas: Rename the variable to `projects_all`?

funnel/views/index.py
```diff
@@ -41,51 +41,51 @@ def whoami():
...
-    spaces = ProposalSpace.fetch_sorted().filter(ProposalSpace.profile != None).all()
-    return jsonp(spaces=map(space_data, spaces))
+    projects = Project.fetch_sorted().filter(Project.profile != None).all()
+    return jsonp(projects=map(project_data, projects))
 
 
 @app.route('/<profile>/json')
 @funnelapp.route('/json', subdomain='<profile>')
 @load_model(Profile, {'name': 'profile'}, 'g.profile', permission='view')
 def spaces_json(profile):
```
iambibhas: `projects_json`?

PR #297, opened 2018-11-19
Update UI of project page

vidya-ram:
Add banner, ticketing and related job sections to project page.


funnel/templates/project.html.jinja2
```diff
@@ -67,51 +67,89 @@
...
 
+{% block contentwrapper %}
   <div class="grid details-section">
-    <div class="grid__col--bleed-y">
-      <hr class="mui-divider">
-    </div>
     <div class="grid__col-xs-12 grid__col-md-5 grid__col-lg-4">
-      <div class="sticky-admin-panel">
-        {%- if project.current_roles.admin or'checkin-event' in g.permissions %}
+      {%- if project.current_roles.admin or'checkin-event' in g.permissions %}
```
jace: Formatting issue: should be `or 'checkin-event'` (space missing). Also, use `current_auth.permissions.checkin-event` instead of `'checkin-event' in g.permissions`.

funnel/views/helpers.py
```diff
@@ -20,3 +22,9 @@ def localize_date(date, from_tz=utc, to_tz=utc):
             to_tz = pytz_timezone(to_tz)
         return from_tz.localize(date).astimezone(to_tz).replace(tzinfo=None)
     return date
```
jace: These methods should be in Coaster. Where are they being used?

funnel/templates/project.html.jinja2
jace: Just realised `current_auth.permissions.checkin-event` won't work because of the hyphen. We have to update all permission names in the app to use underscores instead of hyphens.

funnel/views/helpers.py
vidya-ram: @iambibhas Could you please look into this. It's used in schedule and session views.

iambibhas: Along the 3, only `localize_date()` is used throughout funnel. Rest 2 are not used anywhere. I can add it to coaster.

PR #300, opened 2018-11-21
Add banner image url to Session.

### LINKED ISSUE #291: Sessions should have a cover image field
Session cards may need customisation with cover images to attract attention. The model should include a cover image URL field and the UI should use it.

Also, possibly a colour accent field (depending on opinion from @StephanieBr).

vidya-ram:
Fixes #291 
Add banner image url to Session. 
Add UI for featured talks


migrations/versions/07ebe99161d5_add_banner_image_url_to_sessio.py
```diff
@@ -0,0 +1,23 @@
+"""add banner_image_url field to session
+
+Revision ID: 07ebe99161d5
+Revises: d6b1904bea0e
+Create Date: 2018-11-21 19:06:35.140390
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '07ebe99161d5'
```
iambibhas: I already have a PR with a migration ready to merge. This needs to be updated if I merge mine first. Or if this goes first then I'll update my PR.

migrations/versions/07ebe99161d5_add_banner_image_url_to_sessio.py
```diff
@@ -0,0 +1,23 @@
...
+down_revision = 'd6b1904bea0e'
```
iambibhas: I already have a PR with a migration ready to merge. This needs to be updated if I merge mine first. Or if this goes first then I'll update my PR.

funnel/__init__.py
```diff
@@ -30,6 +30,7 @@
 assets['schedules.css'][version] = 'css/schedules.css'
 assets['schedules.js'][version] = 'js/schedules.js'
 assets['screens.css'][version] = 'css/screens.css'
+assets['truncate.js'][version] = 'js/libs/jquery-succinct.js'
```
jace: The file should be named `truncate` as well.

funnel/models/project.py
```diff
@@ -87,6 +87,9 @@ class Project(BaseScopedNameMixin, db.Model):
     inherit_sections = db.Column(db.Boolean, default=True, nullable=False)
     labels = db.Column(JsonDict, nullable=False, server_default='{}')
 
+    featured_sessions = db.relationship("Session",
+        primaryjoin="and_(Session.project_id==Project.id, Session.featured==True)", back_populates="project")
```
jace: Should be `and_(Session.project_id == id, Session.featured == True)`

funnel/__init__.py
jace: Or if the plugin is named `succint`, so should the asset.

funnel/views/session.py
```diff
@@ -13,7 +13,7 @@
 
 def rooms_list(project):
     return [(u"", _("Select Room"))] + [
-        (room.id, "{venue} - {room}".format(venue=room.venue.title, room=room.title)) for room in project.rooms]
+        (room.id, u"{venue} - {room}".format(venue=room.venue.title, room=room.title)) for room in project.rooms]
```
jace: The hyphen here should be an n-dash.

funnel/models/project.py
iambibhas: @jace using only `id` instead of `Project.id` throws error - 
> ArgumentError: Could not locate any relevant foreign key columns for primary join condition 'session.project_id = :project_id_1 AND session.featured = true' on relationship Project.featured_sessions. Ensure that referencing columns are associated with a ForeignKey or ForeignKeyConstraint, or are annotated in the join condition with the foreign() annotation.

jace: If it points to a column in the model, it'll return featured sessions from other projects. Test and confirm.

iambibhas: @jace tested across 2 projects. Each one shows the respective featured sessions. Not from each other.

iambibhas: I took the usage reference from here (the example) https://docs.sqlalchemy.org/en/latest/orm/join_conditions.html#specifying-alternate-join-conditions

jace: Okay, this is good then. I guess SQLAlchemy is being intelligent about value substitution.

funnel/__init__.py
vidya-ram: Changed it to `succinct`

migrations/versions/07ebe99161d5_add_banner_image_url_to_sessio.py
vidya-ram: Updated the `down_revision`

funnel/views/session.py
vidya-ram: changed it to `{venue}–{room}`

PR #301, opened 2018-11-22
Add logo url field to profile form

vidya-ram:


funnel/forms/profile.py
```diff
@@ -23,10 +23,12 @@ class EditProfileForm(forms.Form):
     """
     Edit a profile.
     """
-    description = forms.MarkdownField(u"Welcome message",
+    description = forms.MarkdownField(__("Welcome message"),
         validators=[forms.validators.DataRequired(_(u"Please write a message for the landing page"))],
-        description=_(u"This welcome message will be shown on the landing page."))
+        description=__("This welcome message will be shown on the landing page."))
+    logo_url = forms.URLField(__("Logo URL"), description=__("Profile logo"),
+        validators=[forms.validators.Optional(), forms.validators.Length(max=2000)])
```
jace: Use the `valid_url` validator here. We should add it to all image URL fields.

PR #302, opened 2018-11-22
Unscheduled session

### LINKED ISSUE #290: Unscheduled sessions should be possible
Sessions represent the event owner's copy of a Proposal, allowing for customisation without changing the original. However, when scheduling events, @zainabbawa frequently finds herself moving proposals into and out of the schedule to find a correct fit. Since sessions require a mandatory start/end time, this means session objects are created and deleted on such moves, losing all customisation.

Instead, sessions should support an unscheduled mode and should not be deleted. This is necessary to preserve customisations and to support featuring (#289) before the schedule is finalised.

iambibhas:
Fixes #290. 

Retains session objects in database when sessions are deleted from schedule, and reuse them whenever a session for the same proposal is created again.

### DIFF
--- a/funnel/models/project.py
+++ b/funnel/models/project.py
@@ -88,7 +88,9 @@ class Project(BaseScopedNameMixin, db.Model):
     labels = db.Column(JsonDict, nullable=False, server_default='{}')
 
     featured_sessions = db.relationship("Session",
-        primaryjoin="and_(Session.project_id == Project.id, Session.featured == True)", back_populates="project")
+        primaryjoin="and_(Session.project_id == Project.id, Session.start != None, Session.end != None, Session.featured == True)", back_populates="project")
+    scheduled_sessions = db.relationship("Session",
+        primaryjoin="and_(Session.project_id == Project.id, Session.start != None, Session.end != None)", back_populates="project")
 
     #: Redirect URLs from Funnel to Talkfunnel
     legacy_name = db.Column(db.Unicode(250), nullable=True, unique=True)
--- a/funnel/models/project.py
+++ b/funnel/models/project.py
@@ -88,9 +88,11 @@ class Project(BaseScopedNameMixin, db.Model):
     labels = db.Column(JsonDict, nullable=False, server_default='{}')
 
     featured_sessions = db.relationship("Session",
-        primaryjoin="and_(Session.project_id == Project.id, Session.start != None, Session.end != None, Session.featured == True)", back_populates="project")
+        primaryjoin="and_(Session.project_id == Project.id, Session.featured == True)")
     scheduled_sessions = db.relationship("Session",
-        primaryjoin="and_(Session.project_id == Project.id, Session.start != None, Session.end != None)", back_populates="project")
+        primaryjoin="and_(Session.project_id == Project.id, Session.start != None, Session.end != None)")
+    unscheduled_sessions = db.relationship("Session",
+        primaryjoin="and_(Session.project_id == Project.id, Session.start == None, Session.end == None)")
 
     #: Redirect URLs from Funnel to Talkfunnel
     legacy_name = db.Column(db.Unicode(250), nullable=True, unique=True)
--- a/funnel/models/project.py
+++ b/funnel/models/project.py
@@ -87,12 +87,12 @@ class Project(BaseScopedNameMixin, db.Model):
     inherit_sections = db.Column(db.Boolean, default=True, nullable=False)
     labels = db.Column(JsonDict, nullable=False, server_default='{}')
 
-    featured_sessions = db.relationship("Session",
-        primaryjoin="and_(Session.project_id == Project.id, Session.featured == True)")
-    scheduled_sessions = db.relationship("Session",
-        primaryjoin="and_(Session.project_id == Project.id, Session.start != None, Session.end != None)")
-    unscheduled_sessions = db.relationship("Session",
-        primaryjoin="and_(Session.project_id == Project.id, Session.start == None, Session.end == None)")
+    featured_sessions = db.relationship('Session',
+        primaryjoin='and_(Session.project_id == Project.id, Session.featured == True)')
+    scheduled_sessions = db.relationship('Session',
+        primaryjoin='and_(Session.project_id == Project.id, Session.scheduled)')
+    unscheduled_sessions = db.relationship('Session',
+        primaryjoin='and_(Session.project_id == Project.id, Session.scheduled != True)')
 
     #: Redirect URLs from Funnel to Talkfunnel
     legacy_name = db.Column(db.Unicode(250), nullable=True, unique=True)
--- a/funnel/models/session.py
+++ b/funnel/models/session.py
@@ -44,6 +44,12 @@ class Session(UuidMixin, BaseScopedIdNameMixin, db.Model):
         # A session is scheduled only when both start and end fields have value
         return (self.start != None) & (self.end != None)
 
+    def make_unscheduled(self):
+        # Session is not deleted, but we remove start and end time,
+        # so it becomes an unscheduled session.
+        self.start = None
+        self.end = None
+
     def url_for(self, action='view', _external=False):
         if action == 'view':
             return self.project.url_for('schedule', _external=_external) + u'#' + self.url_name
--- a/funnel/models/session.py
+++ b/funnel/models/session.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from flask import url_for
+from sqlalchemy.ext.hybrid import hybrid_property
 from . import db, UuidMixin, BaseScopedIdNameMixin, MarkdownColumn
 from .project import Project
 from .proposal import Proposal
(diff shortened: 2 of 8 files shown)


funnel/models/project.py
```diff
@@ -88,7 +88,9 @@ class Project(BaseScopedNameMixin, db.Model):
     labels = db.Column(JsonDict, nullable=False, server_default='{}')
 
     featured_sessions = db.relationship("Session",
-        primaryjoin="and_(Session.project_id == Project.id, Session.featured == True)", back_populates="project")
+        primaryjoin="and_(Session.project_id == Project.id, Session.start != None, Session.end != None, Session.featured == True)", back_populates="project")
```
jace: Unscheduled sessions can be featured. That was the main motivation for having unscheduled sessions.

funnel/models/project.py
```diff
@@ -88,7 +88,9 @@ class Project(BaseScopedNameMixin, db.Model):
...
+    scheduled_sessions = db.relationship("Session",
+        primaryjoin="and_(Session.project_id == Project.id, Session.start != None, Session.end != None)", back_populates="project")
```
jace: You'll also need one for unscheduled sessions.

funnel/models/session.py
```diff
@@ -23,8 +23,8 @@ class Session(UuidMixin, BaseScopedIdNameMixin, db.Model):
     proposal = db.relationship(Proposal,
         backref=db.backref('session', uselist=False, cascade='all, delete-orphan'))
     speaker = db.Column(db.Unicode(200), default=None, nullable=True)
-    start = db.Column(db.DateTime, nullable=False)
-    end = db.Column(db.DateTime, nullable=False)
+    start = db.Column(db.DateTime, nullable=True)
+    end = db.Column(db.DateTime, nullable=True)
```
jace: Add a check constraint to ensure both or neither are set.

funnel/models/venue.py
```diff
@@ -44,6 +44,9 @@ class VenueRoom(BaseScopedNameMixin, db.Model):
     description = MarkdownColumn('description', default=u'', nullable=False)
     bgcolor = db.Column(db.Unicode(6), nullable=False, default=u'229922')
 
+    scheduled_sessions = db.relationship("Session",
+        primaryjoin="and_(Session.venue_room_id == VenueRoom.id, Session.start != None, Session.end != None)", back_populates="venue_room")
```
jace: Did you really mean to use `back_populates`? https://docs.sqlalchemy.org/en/latest/orm/relationship_api.html#sqlalchemy.orm.relationship.params.back_populates

funnel/templates/session_view_popup.html.jinja2
```diff
@@ -8,12 +8,18 @@
...
+    {% if session.scheduled %}
+      <p class="mui--text-subhead zero-bottom-margin">
+        <small class="mui--text-light">at</small>
+        {{ localize_date(session.start, to_tz=timezone).strftime('%I:%M %p') }} - {{ localize_date(session.end, to_tz=timezone).strftime('%I:%M %p') }}
+        <small class="mui--text-light">on</small>
+        {{ localize_date(session.start, to_tz=timezone).strftime('%d %B, %Y') }}
+      </p>
+    {% else %}
+      <p class="mui--text-subhead zero-bottom-margin">
+        This session has not been scheduled yet.
```
jace: Add `{% trans %}`

funnel/views/schedule.py
```diff
@@ -81,6 +81,12 @@ def schedule_data(project):
 
 
 def session_ical(session):
+    # This function is only called with scheduled sessions.
+    # If for some reason it is used somewhere else and called with an unscheduled session,
+    # this function should fail.
+    if not session.scheduled:
+        raise Exception(u"{} is not scheduled".format(session))
```
jace: Use `{0!r}` instead of `{}`.

jace: https://docs.python.org/2.7/library/string.html#format-string-syntax

funnel/views/schedule.py
```diff
@@ -259,8 +265,10 @@ def schedule_edit(profile, project):
         'scheduled': session_data(project.sessions, timezone=project.timezone, with_modal_url='edit', with_delete_url=True)
         }
     # Set the proper range for the calendar to allow for date changes
-    first_session = Session.query.filter_by(project=project).order_by("created_at asc").first()
-    last_session = Session.query.filter_by(project=project).order_by("created_at desc").first()
+    first_session = Session.query.filter(
+        Session.start.isnot(None), Session.end.isnot(None)).filter_by(project=project).order_by("created_at asc").first()
+    last_session = Session.query.filter(
+        Session.start.isnot(None), Session.end.isnot(None)).filter_by(project=project).order_by("created_at desc").first()
```
jace: Make `scheduled` a `hybrid_property` and you can rewrite this as `Session.query.filter(Session.scheduled)…`

funnel/views/session.py
```diff
@@ -17,15 +17,19 @@ def rooms_list(project):
 
 
 def session_form(project, proposal=None, session=None):
-    if session:
+    # Look for any existing unscheduled session
+    if proposal and not session:
+        session = Session.query.filter_by(project=project, proposal=proposal).first()
```
jace: Add a classmethod to Session and make this call `Session.for_proposal(proposal, create=True)`

funnel/views/session.py
```diff
@@ -111,6 +115,9 @@ def session_edit(profile, project, session):
     permission='edit-session')
 def session_delete(profile, project, session):
     modal_url = session.proposal.url_for('schedule') if session.proposal else None
-    db.session.delete(session)
+    # Session is not deleted, but we remove start and end time,
+    # so it becomes an unscheduled session.
+    session.start = None
+    session.end = None
```
jace: Make this a method: `Session.remove_schedule()`

jace: Or `Session.make_unscheduled()`

funnel/models/venue.py
iambibhas: Let me test couple of things. The complementing property already exists for us.

funnel/models/session.py
```diff
@@ -23,15 +24,42 @@ class Session(UuidMixin, BaseScopedIdNameMixin, db.Model):
...
     is_break = db.Column(db.Boolean, default=False, nullable=False)
     featured = db.Column(db.Boolean, default=False, nullable=False)
     banner_image_url = db.Column(db.Unicode(2000), nullable=True)
 
-    __table_args__ = (db.UniqueConstraint('project_id', 'url_id'),)
+    __table_args__ = (
+        db.UniqueConstraint('project_id', 'url_id'),
+        db.CheckConstraint(
+            u'("start" IS NULL AND "end" IS NULL) OR ("start" IS NOT NULL AND "end" IS NOT NULL)',
+            u'ck_session_start_end_nullable')
```
jace: Coding convention is `_check` suffix, so `session_start_end_check`.

funnel/models/session.py
```diff
@@ -23,15 +24,42 @@ class Session(UuidMixin, BaseScopedIdNameMixin, db.Model):
...
+        )
+
+    @hybrid_property
+    def scheduled(self):
+        # A session is scheduled only when both start and end fields have value
+        return (self.start != None) & (self.end != None)
```
jace: You'll need separate implementations for Python and SQLAlchemy. This should be the SQL version defined with `@scheduled.expression`

funnel/models/session.py
```diff
@@ -23,15 +24,42 @@ class Session(UuidMixin, BaseScopedIdNameMixin, db.Model):
...
+
+    @classmethod
+    def for_proposal(cls, proposal, create=False):
+        session_obj = cls.query.filter_by(proposal=proposal).first()
+        if session_obj is None and create:
+            session_obj = cls(title=proposal.title, description=proposal.description,
+                speaker_bio=proposal.bio, project=proposal.project, proposal=proposal   )
+            session_obj.make_id()
+            session_obj.make_name()
```
jace: This is now automatic.

funnel/views/schedule.py
```diff
@@ -259,8 +265,8 @@ def schedule_edit(profile, project):
         'scheduled': session_data(project.sessions, timezone=project.timezone, with_modal_url='edit', with_delete_url=True)
         }
     # Set the proper range for the calendar to allow for date changes
-    first_session = Session.query.filter_by(project=project).order_by("created_at asc").first()
-    last_session = Session.query.filter_by(project=project).order_by("created_at desc").first()
+    first_session = Session.query.filter(Session.scheduled).filter_by(project=project).order_by("created_at asc").first()
+    last_session = Session.query.filter(Session.scheduled).filter_by(project=project).order_by("created_at desc").first()
```
jace: order_by values should be in single quotes.

funnel/views/session.py
```diff
@@ -17,19 +17,20 @@ def rooms_list(project):
 
 
 def session_form(project, proposal=None, session=None):
+    # Look for any existing unscheduled session
+    if proposal and not session:
+        session = Session.for_proposal(proposal, create=True)
+        db.session.commit()
```
jace: If this is a GET method, we can't create and commit here.

migrations/versions/d62b392b0f25_session_start_end_nullable.py
```diff
@@ -0,0 +1,28 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+def upgrade():
+    op.alter_column('session', 'end', existing_type=postgresql.TIMESTAMP(), nullable=True)
+    op.alter_column('session', 'start', existing_type=postgresql.TIMESTAMP(), nullable=True)
+    op.create_check_constraint(u'ck_session_start_end_nullable', 'session',
```
jace: Name as above.

PR #311, opened 2018-11-29
Added ModelView for Section views

iambibhas:


funnel/views/section.py
```diff
@@ -59,39 +47,49 @@ def section_new(profile, project):
...
-def section_delete(profile, project, section):
-    return render_delete_sqla(section, db, title=_(u"Confirm delete"),
-        message=_(u"Do you really wish to delete section ‘{title}’?").format(title=section.title),
-        success=_("Your section has been deleted"),
-        next=project.url_for('sections'),
-        cancel_url=project.url_for('sections'))
+    @route('', methods=['GET'])
+    @render_with('section.html.jinja2', json=True)
+    def view(self, **kwargs):
+        return {'project': self.obj.project, 'section': self.obj.current_access()}
```
jace: `project` should also be wrapped in `current_access()`

funnel/views/section.py
```diff
@@ -59,39 +47,49 @@ def section_new(profile, project):
...
+            cancel_url=self.obj.project.url_for('sections'))
+
+
+@route('/<project>/sections/<section>', subdomain='<profile>')
+class FunnelSectionView(SectionView):
+    pass
+
+
+SectionView.init_app(app)
+FunnelSectionView.init_app(funnelapp)
```
jace: Have you confirmed this works as expected? With different sets of URLs for each? Asking because I haven't.

iambibhas: Yeah. Tested on both funnel and hasgeekapp. works.

PR #315, opened 2018-12-05
ModelView updates for coaster

iambibhas:


funnel/views/section.py
```diff
@@ -57,7 +57,7 @@ class SectionView(UrlForView, ModelView):
     route_model_map = {'profile': 'project.profile.name', 'project': 'project.name', 'section': 'name'}
     __decorators__ = [lastuser.requires_login]
 
-    def loader(self, kwargs):
+    def loader(self, **kwargs):
```
jace: Don't use `**kwargs`. Name the parameters explicitly.

jace:
This depends on hasgeek/coaster#206.

PR #316, opened 2018-12-06
Using ModelView for Project views

iambibhas:


funnel/models/project.py
```diff
@@ -166,6 +153,19 @@ def relist(self):
...
+        return self.url_for(_external=True)
+
+    @property
+    def url_json(self):
+        return self.url_for('json', _external=True)
+
+    def __init__(self, **kwargs):
+        super(Project, self).__init__(**kwargs)
+        self.voteset = Voteset(type=SET_TYPE.PROJECT)
+        self.commentset = Commentset(type=SET_TYPE.PROJECT)
```
iambibhas: This will go back up. Moved by mistake.

funnel/views/project.py
```diff
@@ -62,134 +62,139 @@ def project_new(profile):
...
+            'space': project_data(self.obj),  # FIXME: Remove when the native app switches over
+            'sections': [section_data(s) for s in sections],
+            'venues': [venue_data(venue) for venue in self.obj.venues],
+            'rooms': [room_data(room) for room in self.obj.rooms],
+            'proposals': [proposal_data(proposal) for proposal in proposals],
+            'schedule': schedule_data(self.obj),
+            })
+
+    @route('csv')
+    @render_with(json=True)
```
jace: CSV is not JSON. Remove this.

funnel/views/project.py
```diff
@@ -62,134 +62,139 @@ def project_new(profile):
...
+            usergroups = []
+        proposals = Proposal.query.filter_by(project=self.obj).order_by(db.desc('created_at')).all()
+        outfile = StringIO()
+        out = unicodecsv.writer(outfile, encoding='utf-8')
+        out.writerow(proposal_headers + ['votes_' + group for group in usergroups] + ['status'])
+        for proposal in proposals:
+            out.writerow(proposal_data_flat(proposal, usergroups))
+        outfile.seek(0)
+        return Response(unicode(outfile.getvalue(), 'utf-8'), content_type='text/csv',
+            headers=[('Content-Disposition', 'attachment;filename="{project}.csv"'.format(project=self.obj.title))])
```
jace: We need to escape `self.obj.title`. If it has quotes (valid user input), it will break the header. We'll have to lookup valid characters for the `Content-Disposition` header since HTML escaping may not be appropriate here.

funnel/views/project.py
```diff
@@ -62,134 +62,139 @@ def project_new(profile):
...
+
+    @route('edit', methods=['GET', 'POST'])
+    @lastuser.requires_login
+    @requires_permission('edit-project')
+    def edit(self):
+        if self.obj.parent:
+            form = ProposalSubprojectForm(obj=self.obj, model=Project)
```
jace: What is a `PropsalSubprojectForm`?

funnel/views/project.py
```diff
@@ -62,134 +62,139 @@ def project_new(profile):
...
+    def rsvp_list(self):
+        return dict(project=self.obj, statuses=RSVP_STATUS)
+
+    @route('admin')
+    @render_with('admin.html.jinja2')
+    @lastuser.requires_login
+    @requires_permission('admin')
+    def admin(self):
+        csrf_form = forms.Form()
+        if csrf_form.validate_on_submit():
```
jace: This will only work on a POST request, but the `@route` doesn't mention it.

funnel/views/proposal.py
```diff
@@ -82,7 +82,7 @@ def proposal_data(proposal):
             ('votes_bydate', proposal.votes_by_date()),
             ('status', proposal.state.value),
             ('state', proposal.state.label.name),
-        ] if 'view-contactinfo' in g.permissions else []))
+        ] if 'view-contactinfo' in proposal.current_permissions else []))
```
jace: Replace `g.permissions` with `current_auth.permissions`

funnel/models/profile.py
```diff
@@ -28,6 +28,11 @@ class Profile(ProfileBase, db.Model):
             },
         }
 
+    @cached_property
+    def teams(self):
+        from .user import Team
+        return Team.query.filter_by(orgid=self.userid).order_by(Team.title)
```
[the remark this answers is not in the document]
iambibhas: I tried this. Doesn't work. SQLAlchemy complains that there is no FK relationship between them.

funnel/views/project.py
iambibhas: Should we replace `title` with `name`?

iambibhas: I checked [rfc6266](https://tools.ietf.org/html/rfc6266) and it mentions - 

> Avoid using non-ASCII characters in the filename parameter.
      Although most existing implementations will decode them as
      ISO-8859-1, some will apply heuristics to detect UTF-8, and thus
      might fail on certain names.

iambibhas: So we can encode the title with ascii just to be safe.

funnel/views/participant.py
```diff
@@ -67,38 +67,48 @@ def participant_checkin_data(participant, project, event):
...
-    return render_form(form=form, title=_(u"New Participant"), submit=_(u"Add Participant"))
+@route('/<profile>/<project>/participants')
+class ProjectParticipantView(UrlForView, ModelView):
+    model = Project
+    route_model_map = {'profile': 'profile.name', 'project': 'name'}
+
+    def loader(self, profile, project):
+        return self.model.query.join(Profile).filter(
+                Project.name == project, Profile.name == profile
+            ).first_or_404()
```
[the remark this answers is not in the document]
iambibhas: @jace if I query passing both profile and project, `.profile` will never come back as `None`, would it?

funnel/models/profile.py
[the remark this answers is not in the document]
iambibhas: Fixed it. As there is no real FK relationship present, had to use [`foreign()`](https://docs.sqlalchemy.org/en/latest/orm/relationship_api.html#sqlalchemy.orm.foreign).

funnel/templates/admin.html.jinja2
```diff
@@ -26,9 +26,9 @@
       </div>
       <div class="mui-divider"></div>
       <div class="card__footer">
-        <a href={{ project.url_for('new-event') }} title="{% trans %}New event {% endtrans %}" class="mui-btn mui-btn--small mui-btn--flat mui-btn--primary">New event</a>
+        <a href={{ project.url_for('events_new') }} title="{% trans %}New event {% endtrans %}" class="mui-btn mui-btn--small mui-btn--flat mui-btn--primary">New event</a>
         {%- if project.events %}
-          <a href={{ project.url_for('new-participant') }} title="{% trans %}New participant {% endtrans %}" class="mui-btn mui-btn--small mui-btn--flat mui-btn--accent">New participant</a>
+          <a href={{ project.url_for('participants_new') }} title="{% trans %}New participant {% endtrans %}" class="mui-btn mui-btn--small mui-btn--flat mui-btn--accent">New participant</a>
```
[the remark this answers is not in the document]
iambibhas: Updated all the views to use `verb_subnoun`.

funnel/views/decorators.py
```diff
@@ -0,0 +1,17 @@
+from functools import wraps
+from flask import g, request, redirect, current_app
+from funnel import app, funnelapp
+
+
+def legacy_redirect(f):
+    @wraps(f)
+    def decorated_function(clview, **kwargs):
+        if g.profile:
+            if g.profile.legacy and current_app == app:
```
jace: `current_app._get_current_object() is app`

funnel/views/decorators.py
```diff
@@ -0,0 +1,17 @@
...
+                with funnelapp.app_context(), funnelapp.test_request_context():
+                    return redirect(clview.obj.url_for(_external=True), code=303)
+            elif not g.profile.legacy and current_app == funnelapp:
```
jace: Same here.

funnel/views/decorators.py
```diff
@@ -0,0 +1,17 @@
...
+from flask import g, request, redirect, current_app
+from funnel import app, funnelapp
+
+
+def legacy_redirect(f):
+    @wraps(f)
+    def decorated_function(clview, **kwargs):
+        if g.profile:
+            if g.profile.legacy and current_app == app:
+                with funnelapp.app_context(), funnelapp.test_request_context():
```
jace: `test_request_context` should not be necessary.

funnel/views/decorators.py
```diff
@@ -0,0 +1,17 @@
+from functools import wraps
+from flask import g, request, redirect, current_app
+from funnel import app, funnelapp
+
+
+def legacy_redirect(f):
+    @wraps(f)
+    def decorated_function(clview, **kwargs):
+        if g.profile:
```
jace: `if g.profile and request.method == 'GET'` (don't redirect on POST!)

funnel/views/decorators.py
```diff
@@ -0,0 +1,17 @@
+from functools import wraps
+from flask import g, request, redirect, current_app
+from funnel import app, funnelapp
+
+
+def legacy_redirect(f):
+    @wraps(f)
+    def decorated_function(clview, **kwargs):
```
jace: I'd call this parameter `classview` to avoid confusion.

funnel/views/project.py
```diff
@@ -74,6 +75,8 @@ class FunnelProfileProjectView(ProfileProjectView):
 
 @route('/<profile>/<project>/')
 class ProjectView(ProjectViewMixin, UrlForView, ModelView):
+    __decorators__ = [legacy_redirect]
+
```
jace: You need this on all classviews that handle a profile, not just ProjectView.

iambibhas: Yeah, i'll do that. Was just testing with this one.

iambibhas: Added it to all modelviews.

funnel/views/decorators.py
iambibhas: url_for doesn't work with only app_context. It needs request context. It's kind of a hack. Not an official feature of flask it seems. https://github.com/pallets/flask/issues/1412

PR #318, opened 2018-12-06
UUID Columns. Resolves #305

### LINKED ISSUE #305: More models need UUID columns
We previously added UUID columns to Proposal and Session in #276. Turns out more models need UUIDs because they need public-facing keys:

* [x] Project
* [x] Profile (already has `userid`, a UUID in Base64 form)
* [x] User (likewise)
* [x] Team (likewise)

jace:
UUID columns for Project, Profile, User, Team

### DIFF
--- a/migrations/versions/b34aa62af7fc_uuid_columns_for_project_profile_user_team.py
+++ b/migrations/versions/b34aa62af7fc_uuid_columns_for_project_profile_user_team.py
@@ -1,0 +1,121 @@
+"""UUID columns for Project, Profile, User, Team
+
+Revision ID: b34aa62af7fc
+Revises: 19a1f7f2a365
+Create Date: 2018-12-06 20:38:40.092866
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'b34aa62af7fc'
+down_revision = '19a1f7f2a365'
+
+from uuid import uuid4
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql import column, table
+from sqlalchemy_utils import UUIDType
+from progressbar import ProgressBar
+import progressbar.widgets
+from coaster.utils import buid2uuid
+
+project = table('project',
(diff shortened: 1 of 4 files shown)


migrations/versions/b34aa62af7fc_uuid_columns_for_project_profile_user_team.py
```diff
@@ -0,0 +1,121 @@
...
+    op.create_unique_constraint('team_uuid_key', 'team', ['uuid'])
+
+    op.add_column('user', sa.Column('uuid', UUIDType(binary=False), nullable=True))
+    count = conn.scalar(
+        sa.select([sa.func.count('*')]).select_from(user))
+    progress = get_progressbar("Users", count)
+    progress.start()
+    items = conn.execute(sa.select([user.c.id, user.c.userid]))
+    for counter, item in enumerate(items):
+        conn.execute(sa.update(user).where(user.c.id == item.id).values(uuid=buid2uuid(item.userid)))
```
iambibhas: Why `buid2uuid` for these 3 cases?

jace: `userid` columns use buid renderings of UUIDs, so we migrate existing values. This is required to ensure the values match what is in Lastuser.

jace:
The unusual definition of `userid` attributes in `Profile`, `User` and `Team` will raise this warning from SQLAlchemy: `SAWarning: Unmanaged access of declarative attribute uuid from non-mapped class UuidMixin`. We can work around this by copying the code for the `buid` hybrid property from `UuidMixin` class, but the warning seems harmless. In any case, this will be resolved when (if?) Funnel stops depending on Flask-Lastuser since there's a proposal to move the `User` and `Organization` tables into Funnel, while `Team` is scheduled to be deprecated from Lastuser.

PR #324, opened 2018-12-10
Fixed comment posting, editing and deleting

iambibhas:


funnel/templates/comments.html.jinja2
```diff
@@ -69,11 +69,9 @@
           {%- endif %}
           <div data-id="{{ comment.id }}" class="comment--footer">
             {% if not comment.state.DELETED %}
-              {% if g.user %}<a title="{% trans %}Reply{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-reply" href="#c{{ comment.id }}">[reply]</a>{% endif %}
-              {%- with cperms = comment.permissions(g.user, g.permissions) %}
-                {% if 'edit-comment' in cperms %}<a title="{% trans %}Edit{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-edit" href="#c{{ comment.id }}">[{% trans %}edit{% endtrans %}]</a>{% endif %}
-                {% if 'delete-comment' in cperms %}<a title="{% trans %}Delete{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-delete" href="#c{{ comment.id }}">[{% trans %}delete{% endtrans %}]</a>{% endif %}
-              {%- endwith %}
+              {% if current_auth.user %}<a title="{% trans %}Reply{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-reply" href="#c{{ comment.id }}">[reply]</a>{% endif %}
```
jace: `{% if current_auth %}`. The issue about anonymous users should be fixed in Coaster separately.

funnel/templates/comments.html.jinja2
```diff
@@ -69,11 +69,9 @@
...
+              {% if comment.current_permissions.edit_comment %}<a title="{% trans %}Edit{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-edit" href="#c{{ comment.id }}">[{% trans %}edit{% endtrans %}]</a>{% endif %}
+              {% if comment.current_permissions.delete_comment %}<a title="{% trans %}Delete{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-delete" href="#c{{ comment.id }}">[{% trans %}delete{% endtrans %}]</a>{% endif %}
```
jace: Let's change the labels to `Edit` and `Delete` (capitalised). Lowercase display should be done in CSS, not in text. @vidya-ram this will require your attention.

funnel/templates/comments.html.jinja2
```diff
@@ -69,11 +69,9 @@
...
             {%- endif %}
             <a title="Permalink" class="mui--text-menu mui--text-accent comment--permalink" href="#c{{ comment.id }}">[{% trans %}link{% endtrans %}]</a>
             {% if comment.parent %}<a title="{% trans %}Parent{% endtrans %}" class="mui--text-menu mui--text-accent comment--parent" href="#c{{ comment.parent.id }}">[{% trans %}parent{% endtrans %}]</a>{% endif %}
```
jace: Also with these two.

funnel/views/commentvote.py
```diff
@@ -1,13 +1,17 @@
...
 from baseframe import _, forms
 
 from .. import app, funnelapp, lastuser
-from ..models import db
-from .mixins import ProposalViewMixin, CommentViewMixin
+from ..forms import CommentForm, DeleteCommentForm
+from ..models import db, Comment
 from .decorators import legacy_redirect
+from .helpers import send_mail
+from .mixins import ProposalViewMixin, CommentViewMixin
```
jace: If mixins are not being reused across models, they should be in the respective file for each model. `CommentViewMixin` should be in `views/comments.py`, etc.

funnel/views/commentvote.py
```diff
@@ -46,19 +50,98 @@ def votedown(self):
...
+        # the browser Back button
+        return redirect(to_redirect, code=303)
+
+    @route('comments/delete', methods=['POST'])
+    @lastuser.requires_login
+    @requires_permission('new_comment')
+    @requestargs('comment_id')
+    def delete_comment(self, comment_id):
+        comment = Comment.query.filter_by(id=comment_id).first_or_404()
+        if not comment.current_permissions.delete_comment:
```
jace: IIRC, this will break because it will only grant delete permission to the owner of a comment, whereas previously even admins had the right to delete comments. You have to use `current_auth.permissions.delete_comment` here _and_ change `after_loader` to load permissions using inheritance (`comment.permissions(current_auth.user, proposal.permissions(current_auth.user))`). The `after_loader` bit should be fixed in Coaster with the `MultiModelView` issue (ticket is pending to be filed there).

funnel/views/commentvote.py
```diff
@@ -46,19 +50,98 @@ def votedown(self):
...
+                if item.get('to'):
+                    # Sender is set to None to prevent revealing email.
+                    send_mail(sender=None, body=email_body, **item)
+        # Redirect despite this being the same page because HTTP 303 is required to not break
+        # the browser Back button
+        return redirect(to_redirect, code=303)
+
+    @route('comments/delete', methods=['POST'])
+    @lastuser.requires_login
+    @requires_permission('new_comment')
```
jace: Wrong permission.

funnel/views/commentvote.py
jace: Wait, why is the comment not being loaded in `loader`? All the background processing goes for a toss when the loading moves into the view.

iambibhas: @jace this is not a CommentView. It's a ProposalView. The frontend code makes it difficult to POST to the `comments/<id>/delete` URL. The handling is very weird there. Hence the request is being made to `<proposal_id>/comments/delete` with the comment_id as POST payload.

funnel/views/commentvote.py
iambibhas: `delete_comment` is not available here for the reason posted above. The right permission is being checked inside the view for same reason. The comments edit/delete code in the front end needs to change to use right permission.

funnel/models/commentvote.py
```diff
@@ -80,12 +82,14 @@ class Commentset(BaseMixin, db.Model):
     type = db.Column(db.Integer, nullable=True)
     count = db.Column(db.Integer, default=0, nullable=False)
 
+    proposal = db.relationship("Proposal", back_populates="commentset", uselist=False)
```
iambibhas: Now I'm thinking this might not be right. Was there a reason for making a single commentset/voteset being able to relate with multiple proposals?

jace: Commentsets can be linked to multiple models, not just proposals.

funnel/views/commentvote.py
jace: I assume this is fixed now? Didn't notice it when reading through the code.

iambibhas: Yeah. This is fixed.

funnel/views/commentvote.py
```diff
@@ -46,19 +52,80 @@ def votedown(self):
...
 
+    @route('comments/new', methods=['POST'])
+    @lastuser.requires_login
+    @requires_permission('new_comment')
+    def new_comment(self):
+        to_redirect = self.obj.url_for(_external=True)
+        commentform = CommentForm(model=Comment)
+        if commentform.validate_on_submit():
+            send_mail_info = []
+            if commentform.comment_edit_id.data:
```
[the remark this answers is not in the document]
iambibhas: The frontend is a bit messed up for this. I need to sit with vidya to fix this. also, found out today that a part of current comment UI is broken right now which will get fixed with the schedule PR that's on review right now. So I'd prefer to fix this once that's fixed. Can't test it properly without that fix.

funnel/templates/comments.html.jinja2
```diff
@@ -67,21 +67,19 @@
...
-                {% if 'edit-comment' in cperms %}<a title="{% trans %}Edit{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-edit" href="#c{{ comment.id }}">[{% trans %}edit{% endtrans %}]</a>{% endif %}
-                {% if 'delete-comment' in cperms %}<a title="{% trans %}Delete{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-delete" href="#c{{ comment.id }}">[{% trans %}delete{% endtrans %}]</a>{% endif %}
-              {%- endwith %}
+              {% if current_auth %}<a title="{% trans %}Reply{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-reply" href="#c{{ comment.id }}">[Reply]</a>{% endif %}
+              {% if comment.current_permissions.edit_comment %}<a title="{% trans %}Edit{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-edit" href="#c{{ comment.id }}">[{% trans %}Edit{% endtrans %}]</a>{% endif %}
+              {% if comment.current_permissions.delete_comment %}<a title="{% trans %}Delete{% endtrans %}" class="mui--text-menu mui--text-accent js-comment-delete" href="#c{{ comment.id }}">[{% trans %}Delete{% endtrans %}]</a>{% endif %}
```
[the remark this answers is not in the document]
iambibhas: these buttons are comment specific. and only need to show up if that particular comment has that permission. Wouldnt using `current_auth.permissions` make these button show up on all comments? Right now `comment.current_permissions.edit_comment` is not available at this point as this is loaded by the proposal view and those permissions dont get assigned at this point.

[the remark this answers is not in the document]
jace: Then use what the old code was using: `comment.permissions(current_auth.user, current_auth.permissions)`.

PR #326, opened 2018-12-11
Using flask-rq2

iambibhas:
This replaces existing dormant Flask-RQ with the more maintained [Flask-RQ2](https://flask-rq2.readthedocs.io/en/latest/).


jace:
Please do this on all projects linked to hasgeek/hgapp#11. Might as well get this migration done with.

PR #329, opened 2018-12-14
Schedule UI fixes

vidya-ram:
1. Add macro for header and navbar section to be shown on project rel…
2. Add header and navbar to schedule page.
3. UI fix: show full description text in session box and fade out at the bottom. Clicking on session box opens the modal and update browser history using pushState
4. Add permalinks to items in the schedule.

ezgif com-video-to-gif 1


funnel/static/css/app.css
```diff
@@ -845,7 +850,7 @@
 .schedule .schedule__row .schedule__row__column {
   padding: 5px 10px;
   position: relative;
-  background: url('../img/schedule-box-background.png?1544501585');
+  background: url('../img/schedule-box-background.png?1544526769');
```
iambibhas: What is this ID?

vidya-ram: compass auto generates the url for caching purpose.
http://compass-style.org/reference/compass/helpers/urls/#image-url

jace: The id number changes when Compass detects that the referred file has changed. However, since this change detection is based on the filesystem (inode number, I think), if two developers work on the same stylesheet file, Compass will assume all the image files have changed as well. We will only solve this problem if we move build to an automated subsystem, or if Compass switches to an advanced change detection system, such as depending on Git to identify changed files.

funnel/views/session.py
```diff
@@ -86,7 +86,7 @@ class SessionView(SessionViewMixin, UrlForView, ModelView):
 
     @route('viewsession')
     def view(self):
-        return redirect(self.obj.project.url_for('schedule') + u"#" + self.obj.url_name, code=303)
+        return redirect(self.obj.project.url_for('view_session_on_schedule', session=self.obj.url_name_suuid), code=303)
```
jace: This should be `self.obj.url_for()`. The view should not be in ProjectView.

funnel/views/session.py
```diff
@@ -80,13 +83,18 @@ class FunnelProjectSessionView(ProjectSessionView):
...
 
 
-@route('/<profile>/<project>/<session>')
+@route('/<profile>/<project>/schedule/<session>')
 class SessionView(SessionViewMixin, UrlForView, ModelView):
     __decorators__ = [legacy_redirect]
 
-    @route('viewsession')
+    @route('')
+    @render_with('schedule.html.jinja2')
```
jace: Add `json=True`

PR #347, opened 2019-01-28
Add touch swipe support to move between tabs on the project page

vidya-ram:
1. Add touch swipe support to move between tabs on the project page
2. Add support to show active section on navbar
3. Refactor JS to ES6 and add webpack, babel to build and transpile.
4. Add support to lazy load images  (load images when they are scrolled into the visible viewport)


Makefile
```diff
@@ -0,0 +1,5 @@
+assets:
+	cd funnel/assets && npm install && npm run build
+
+assetsonly:
+	cd funnel/assets && npm run build
```
jace: Add an `all:` target. This will get called when `make` is executed without any parameters. It should be used for the default action. Also, the commands can be on separate lines without using `&&` as Make will abort if any fails (the equivalent of `set -e` in a bash script).

funnel/assets/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+      'ObjectExpression': "always",
+      'ObjectPattern': { "multiline": true }
+    }],
+    'new-cap': WARN,
+    'no-param-reassign' : [ERROR, { "props": false }],
+    'max-len' : WARN,
+    // allow debugger during development
+    'no-debugger': process.env.NODE_ENV === 'production' ? ERROR : OFF
+  }
+}
```
jace: Missing newline.

funnel/assets/js/app.js
```diff
@@ -0,0 +1,37 @@
...
+        intersectionObserverComponents();
+      };
+      document.head.appendChild(polyfill);
+    }
+    else {
+      intersectionObserverComponents();
+    }
+  }
+
+  // Send click events to Google analytics
```
jace: If swiping moves to a different content area, shouldn't that also be logged to GA?

funnel/assets/js/project.js
```diff
@@ -0,0 +1,142 @@
+import { Utils, TableSearch } from './util';
+import L from "leaflet";
+
+const defaultLatitude = '12.961443';
+
+const defaultLongitude = '77.64435000000003';
```
jace: Where does this point, and why is it the default? Defaults can have interesting consequences:

https://splinternews.com/how-an-internet-mapping-glitch-turned-a-random-kansas-f-1793856052
https://gizmodo.com/how-cartographers-for-the-u-s-military-inadvertently-c-1830758394

funnel/templates/layout.html.jinja2
```diff
@@ -7,7 +7,7 @@
   <meta property="og:title" content="{{ self.title()|e }}" />
   <link rel="preconnect" crossorigin href="https://fonts.googleapis.com/">
   <link rel="preconnect" crossorigin href="https://ajax.googleapis.com/">
-  <link rel="preconnect" crossorigin href="https://images.hasgeek.com/">
+  <link rel="preconnect" crossorigin href="https://imgee.s3.amazonaws.com/">
```
jace: I think we'll need both. Sometimes we're using Imgee app URLs to access images.

funnel/assets/js/app.js
vidya-ram: We are sending these events to GA.
On swipe, an click event to prev/next sibling of active nav item(anchor tag) is triggered. Every click on a button or anchor tags are currently send to GA.

PR #352, opened 2019-02-05
Draft model and autosave on Project edit form

iambibhas:
This PR introduces Draft model as proposed in https://github.com/hasgeek/baseframe/issues/207 in `Project` edit form.

Depends on https://github.com/hasgeek/baseframe/pull/208


funnel/models/__init__.py
```diff
@@ -3,19 +3,20 @@
...
 from .project import *
-from .section import *
-from .usergroup import *
 from .proposal import *
-from .feedback import *
+from .rsvp import *
+from .section import *
 from .session import *
+from .user import *
+from .usergroup import *
```
iambibhas: Sorted them alphabetically.

funnel/views/project.py
```diff
@@ -129,23 +131,77 @@ def csv(self):
...
+                    elif client_revision is not None and draft.revision == client_revision:
+                        # revision ID sent my client matches, save updated draft data and update revision ID
+                        draft.body = {'form': request.form}
+                        draft.revision = uuid4()
+                else:
+                    # no draft exists, create one
+                    draft = Draft(table=Project.__tablename__, table_row_id=self.obj.uuid, body={'form': request.form}, revision=uuid4())
+                db.session.add(draft)
+                db.session.commit()
+                return {'draft': draft.body['form'], 'revision': draft.revision}
```
jace: Why are we sending the form back? Only revision is required.

funnel/views/project.py
```diff
@@ -129,23 +131,77 @@ def csv(self):
...
+            else:
+                if self.obj.parent_project:
+                    form = SubprojectForm(obj=self.obj, model=Project)
+                else:
+                    form = ProjectForm(obj=self.obj, parent=self.obj.profile, model=Project)
+                form.parent_project.query = Project.query.filter(Project.profile == self.obj.profile, Project.id != self.obj.id, Project.parent_project == None)  # NOQA
```
jace: This line should be moved to `set_queries` in the form.

funnel/views/project.py
```diff
@@ -129,23 +131,77 @@ def csv(self):
...
+                if form.validate_on_submit():
+                    form.populate_obj(self.obj)
+                    db.session.commit()
+                    flash(_("Your changes have been saved"), 'info')
+                    tag_locations.queue(self.obj.id)
+
+                    # find and delete drafts
+                    Draft.query.filter_by(table=Project.__tablename__, table_row_id=self.obj.uuid).delete()
```
jace: Once the columns are part of a primary key, this can change to `Draft.query.get((Project.__tablename__, self.obj.uuid)).delete()`.

funnel/templates/formlayout.html.jinja2
```diff
@@ -17,3 +28,91 @@
...
+                  errorMsg = 'Unable to connect. Please reload and try again.';
+                }
+                $('#autosave-msg').text(errorMsg);
+                window.toastr.error(errorMsg);
+              },
+            });
+          }
+
+          function haveDirtyFields() {
+            var latestFormData = $('#{{ ref_id }}').find(':input:not(:hidden)').serialize();
```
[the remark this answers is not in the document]
iambibhas: @vidya-ram ^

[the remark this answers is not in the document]
vidya-ram: There is event to detect user typing into input fields and event to detect form changes for fields like select, radio. Incase user selects browser's autocomplete, both these events are triggered and we end up sending two ajax events. This was added to avoid this.

We are actually comparing two longs strings since `serialize` returns text string in standard URL-encoded notation.

funnel/views/project.py
```diff
@@ -129,23 +131,77 @@ def csv(self):
...
+                draft = Draft.query.filter_by(table=Project.__tablename__, table_row_id=self.obj.uuid).order_by(Draft.updated_at.desc()).first()
+
+                if draft is not None:
+                    if client_revision is None or (client_revision is not None and draft.revision != client_revision):
+                        # draft exists, but the form did not send a revision ID,
+                        # OR revision ID sent by client does not match the last revision ID
+                        return {'error': _("There has been changes to this draft since you last edited it. Please reload.")}, 400
+                    elif client_revision is not None and draft.revision == client_revision:
+                        # revision ID sent my client matches, save updated draft data and update revision ID
+                        draft.body = {'form': request.form}
```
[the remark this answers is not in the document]
iambibhas: I updated it to only take in changes available in data sent by client. check once.

funnel/views/project.py
```diff
@@ -129,23 +131,77 @@ def csv(self):
...
+            if draft is not None:
+                form.revision.data = draft.revision
+
+            if not self.obj.timezone:
+                form.timezone.data = current_app.config.get('TIMEZONE')
+            return render_form(form=form, title=_("Edit project"), submit=_("Save changes"), autosave=True)
+        elif request.method == 'POST':
+            if 'autosave' in request.form and request.form['autosave'] == 'true':
+                if 'revision' not in request.form:
+                    return {'error': _("Form must contain a valid revision ID.")}, 400
```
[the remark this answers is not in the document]
iambibhas: Added a comment.

funnel/models/draft.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+from sqlalchemy_utils import UUIDType
+from . import db, JsonDict, NoIdMixin, TimestampMixin
+
+__all__ = ['Draft']
+
+
+class Draft(NoIdMixin, TimestampMixin, db.Model):
```
jace: `NoIdMixin` already has `TimestampMixin`.

funnel/models/draft.py
```diff
@@ -0,0 +1,17 @@
...
+    """Store for autosaved, unvalidated drafts on behalf of other models"""
+    __tablename__ = 'draft'
+    __uuid_primary_key__ = True
```
jace: This is not relevant without `IdMixin`.

funnel/views/project.py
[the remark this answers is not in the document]
jace: The first autosave will not have a revision. There is nothing to revise.

funnel/views/project.py
```diff
@@ -129,23 +132,81 @@ def csv(self):
...
+                                existing[key] = incoming[key]
+                        draft.body = {'form': existing}
+                        draft.revision = uuid4()
+                else:
+                    # no draft exists, create one
+                    draft = Draft(
+                        table=Project.__tablename__, table_row_id=self.obj.uuid,
+                        body={'form': request.form.items(multi=True)}, revision=uuid4()
+                        )
+                db.session.add(draft)
```
jace: This requires a CSRF check. Without the check, a malicious third party website can dump junk content into the Draft model by performing a cross-site POST (which is what CSRF protects from).

1. Do a CSRF check using `Form().validate_on_submit()` (the base Form object has nothing but a CSRF field).
2. Remove the CSRF field when saving to the Draft model. It will cause problems when restored later.

jace:
Change on this comment: https://github.com/hasgeek/funnel/pull/352#discussion_r255289411

1. The field is `form.revision`, not `revision`.
2. The comment is incorrect since there will be an empty field in the form, and this code is only check for the presence of the field, not the content.

funnel/views/mixins.py
```diff
@@ -130,3 +133,75 @@ def loader(self, profile, project, section):
             ).first_or_404()
         g.profile = section.project.profile
         return section
+
+
+class DraftModelViewMixin(object):
```
jace: `DraftViewMixin`.

funnel/views/mixins.py
```diff
@@ -130,3 +133,75 @@ def loader(self, profile, project, section):
...
+    def get_draft(self, current_obj=None):
+        """
+        Returns the draft object for `current_obj`. Defaults to `self.obj`.
+        `current_obj` is needed in case of multi-model views.
+        """
+        obj = current_obj or self.obj
```
jace: `obj = current_obj if current_obj is not None else self.obj`. An object may intentionally represent itself as a boolean (implementing `__bool__` or `__nonzero__`), so this is not a safe assumption when handling objects of unknown type.

jace: Also, make the parameter just `obj`.

funnel/views/mixins.py
```diff
@@ -130,3 +133,75 @@ def loader(self, profile, project, section):
...
+        return Draft.query.get((self.model.__tablename__, obj.uuid))
+
+    def delete_draft(self, current_obj=None):
+        """
+        Deletes draft for `current_obj`, or `self.obj` if `current_obj` is `None`.
+        """
+        draft = self.get_draft(current_obj)
+        if draft is not None:
+            db.session.delete(draft)
+            db.session.commit()
```
jace: 1. This feels like it should raise an error if no draft exists. Caller must be responsible for calling only when appropriate.

2. Don't commit before the end of the transaction. You don't know what else is part of the transaction. Let the caller be responsible for committing.

funnel/views/mixins.py
```diff
@@ -130,3 +133,75 @@ def loader(self, profile, project, section):
...
+
+    def autosave_get(self, current_obj=None):
+        """
+        Returns a tuple of the current draft revision and the formdata needed to initialize forms
+        """
+        draft = self.get_draft(current_obj)
+        initial_formdata = MultiDict(draft.body['form']) if draft is not None else None
+        return draft.revision, initial_formdata
```
jace: If this method is not handling a GET request (as the `_get` suffix suggests), it should not be called `autosave_get`. This should be `get_draft_data`.

funnel/views/mixins.py
```diff
@@ -130,3 +133,75 @@ def loader(self, profile, project, section):
...
+
+    def autosave_post(self, current_obj=None):
+        """
+        Handles autosave POST requests
+        """
+        if 'form.revision' not in request.form:
+            # as form.autosave is true, the form should have `form.revision` field even if it's empty
+            return {'error': _("Form must contain a revision ID.")}, 400
+        if forms.Form().validate_on_submit():
```
jace: Add comment: `# CSRF check`

funnel/views/mixins.py
```diff
@@ -130,3 +133,75 @@ def loader(self, profile, project, section):
...
+            incoming_data = MultiDict(request.form.items(multi=True))
+            client_revision = incoming_data.pop('form.revision')
+            incoming_data.pop('csrf_token')
```
jace: `incoming_data.pop('csrf_token', None)`. There won't be a token if the form has CSRF disabled. This will be the case in unit tests.

funnel/views/mixins.py
```diff
@@ -130,3 +133,75 @@ def loader(self, profile, project, section):
...
+                            existing[key] = incoming_data[key]
+                    draft.body = {'form': existing}
+                    draft.revision = uuid4()
+            elif draft is None and client_revision:
+                # The form contains a revision ID but no draft exists.
+                # Somebody is making autosave requests with an invalid draft ID.
+                return {'error': _("Invalid revision ID or the existing changes have been submitted already. Please reload.")}, 400
+            else:
+                # no draft exists, create one
+                obj = current_obj or self.obj
```
jace: Move this to the beginning of the method and use the local variable everywhere. Someone editing this in future will fail to notice which object is being used and make mistakes. Also, use the `is None` test, not `bool`.

funnel/views/mixins.py
```diff
@@ -130,3 +133,75 @@ def loader(self, profile, project, section):
...
+                draft = Draft(
+                    table=Project.__tablename__, table_row_id=obj.uuid,
+                    body={'form': incoming_data}, revision=uuid4()
```
jace: I'm uncomfortable with this `body={'form': incoming_data}`. Add a `formdata` property to the Draft model that reads and sets `self.body['form']` and set `formdata` here (using `Draft(formdata=incoming_data, ...)` should work just fine).

funnel/views/mixins.py
```diff
@@ -130,3 +133,75 @@ def loader(self, profile, project, section):
...
+                    )
+            db.session.add(draft)
+            db.session.commit()
+            return {'revision': draft.revision}
+        else:
+            return {'error': _("Invalid CSRF token")}, 401
```
jace: 400 or 403. 401 is a login form.

funnel/models/draft.py
```diff
@@ -0,0 +1,25 @@
...
+    __tablename__ = 'draft'
+
+    table = db.Column(db.UnicodeText, primary_key=True)
+    table_row_id = db.Column(UUIDType(binary=False), primary_key=True)
+    body = db.Column(JsonDict, nullable=False, server_default='{}')
+    revision = db.Column(UUIDType(binary=False))
+
+    @property
+    def formdata(self):
+        return MultiDict(self.body['form']) if 'form' in self.body else MultiDict({})
```
jace: `return MultiDict(self.body.get('form', {}))`

funnel/models/draft.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @formdata.setter
+    def formdata(self, value):
+        self.body = {'form': value}
```
jace: `self.body['form'] = value`. Don't mess with anything else in `self.body`.

funnel/views/mixins.py
```diff
@@ -130,3 +133,80 @@ def loader(self, profile, project, section):
...
+            incoming_data.pop('csrf_token', None)
+
+            # find the last draft
+            draft = self.get_draft(obj)
+
+            if draft is not None:
+                if client_revision is None or (client_revision is not None and str(draft.revision) != client_revision):
+                    # draft exists, but the form did not send a revision ID,
+                    # OR revision ID sent by client does not match the last revision ID
+                    return {'error': _("There has been changes to this draft since you last edited it. Please reload.")}, 400
```
jace: There _have_ been changes to this item since you last edited it. Please reload this page.

funnel/views/mixins.py
```diff
@@ -130,3 +133,80 @@ def loader(self, profile, project, section):
...
+                elif client_revision is not None and str(draft.revision) == client_revision:
+                    # revision ID sent my client matches, save updated draft data and update revision ID
+                    existing = draft.formdata
+                    for key in incoming_data.keys():
+                        if existing[key] != incoming_data[key]:
+                            existing[key] = incoming_data[key]
+                    draft.body = {'form': existing}
```
jace: Set `draft.formdata`

funnel/views/mixins.py
```diff
@@ -130,3 +133,80 @@ def loader(self, profile, project, section):
...
+                if client_revision is None or (client_revision is not None and str(draft.revision) != client_revision):
+                    # draft exists, but the form did not send a revision ID,
+                    # OR revision ID sent by client does not match the last revision ID
+                    return {'error': _("There has been changes to this draft since you last edited it. Please reload.")}, 400
+                elif client_revision is not None and str(draft.revision) == client_revision:
+                    # revision ID sent my client matches, save updated draft data and update revision ID
+                    existing = draft.formdata
+                    for key in incoming_data.keys():
+                        if existing[key] != incoming_data[key]:
+                            existing[key] = incoming_data[key]
```
jace: This is not required. You can simply do `existing.update(incoming_data)`

funnel/views/mixins.py
```diff
@@ -130,3 +133,80 @@ def loader(self, profile, project, section):
...
+                    draft.body = {'form': existing}
+                    draft.revision = uuid4()
+            elif draft is None and client_revision:
+                # The form contains a revision ID but no draft exists.
+                # Somebody is making autosave requests with an invalid draft ID.
+                return {'error': _("Invalid revision ID or the existing changes have been submitted already. Please reload.")}, 400
```
jace: This can happen in the following scenario:

1. Tab A is opened to edit the document
2. Tab A makes some changes and creates a draft
3. Tab B opens the same document and gets the draft
4. Tab B makes some changes and updates the draft
5. Tab B submits, updating the document and deleting the draft
6. Tab A now edits and gets to this error

A variation of this happens if step 2 is skipped. Tab A will now make changes on an obsolete document, losing Tab B's changes. No error will be displayed. To fix this we have to track revision id of not just the draft, but the source document as well. This deserves to be a separate ticket since it's some way off from our original intent of providing autosave functionality.

funnel/views/mixins.py
```diff
@@ -130,3 +133,80 @@ def loader(self, profile, project, section):
...
+                # no draft exists, create one
+                draft = Draft(
+                    table=Project.__tablename__, table_row_id=obj.uuid,
+                    formdata=incoming_data, revision=uuid4()
+                    )
+            db.session.add(draft)
+            db.session.commit()
+            return {'revision': draft.revision}
+        else:
+            return {'error': _("Invalid CSRF token")}, 400
```
jace: I recall we had established a convention for JSON responses earlier:

1. Success: `{"status": "ok", "result": {...}}`
2. Error: `{"status": "error", "error": "error_identifier", "error_description": "English message"}`

The first one I'm ambivalent about, since HTTP status (200 vs 400) is a sufficient indicator of whether the request was successful or not. The second is relevant. The front-end should be able to look up a static error code and invoke custom behaviour. The `error_description` is shown when the front-end does not recognise the error.

funnel/views/project.py
```diff
@@ -129,23 +130,47 @@ def csv(self):
...
+
+            # initialize forms with draft initial formdata.
+            # if no draft exists, initial_formdata is None. wtforms ignore formdata if it's None.
+            if self.obj.parent_project:
+                form = SubprojectForm(obj=self.obj, model=Project, formdata=initial_formdata)
+            else:
+                form = ProjectForm(obj=self.obj, parent=self.obj.profile, model=Project, formdata=initial_formdata)
+
+            if not self.obj.timezone:
+                form.timezone.data = current_app.config.get('TIMEZONE')
```
jace: Unrelated fix, but this should default to the user's timezone, not the app's. Use `current_auth.user.timezone` (it falls back to the app if the user has not specified a timezone).

funnel/views/mixins.py
jace: `draft.formdata = existing`

funnel/views/mixins.py
iambibhas: that doesn't work as `existing` is a `MultiDict` and the `update()` method doesn't replace keys, it just adds and extends.

PR #356, opened 2019-02-07
Added sequence to Venue and VenueRoom

iambibhas:
Dependency https://github.com/hasgeek/baseframe/pull/209

### DIFF
--- a/funnel/forms/venue.py
+++ b/funnel/forms/venue.py
@@ -28,9 +28,6 @@ class VenueForm(forms.Form):
     country = forms.SelectField(__("Country"),
         validators=[forms.validators.Optional(), forms.validators.Length(max=2)],
         choices=country_codes, default="IN")
-    seq = forms.IntegerField(__("Sequence"),
-        description=__("The sequence of the venue on the listing"),
-        validators=[forms.validators.DataRequired(__("Please specify the sequence order"))])
     coordinates = forms.CoordinatesField(__("Location"), description=__("Pick a location on the map"),
         validators=[forms.validators.Optional(), forms.validators.ValidCoordinates()])
 
@@ -42,9 +39,6 @@ class VenueRoomForm(forms.Form):
     bgcolor = forms.StringField(__("Event Color"),
         validators=[forms.validators.DataRequired(), forms.validators.Length(max=6)],
         description=__("RGB Color for the event. Enter without the '#'. E.g. CCCCCC."), default=u"CCCCCC")
-    seq = forms.IntegerField(__("Sequence"),
-        description=__("The sequence of the room on the listing"),
-        validators=[forms.validators.DataRequired(__("Please specify the sequence order"))])
 
     def validate_bgcolor(self, field):
         if not valid_color_re.match(field.data):
--- a/funnel/forms/venue.py
+++ b/funnel/forms/venue.py
@@ -28,6 +28,9 @@ class VenueForm(forms.Form):
     country = forms.SelectField(__("Country"),
         validators=[forms.validators.Optional(), forms.validators.Length(max=2)],
         choices=country_codes, default="IN")
+    seq = forms.IntegerField(__("Sequence"),
+        description=__("The sequence of the venue on the listing"),
+        validators=[forms.validators.DataRequired(__("Please specify the sequence order"))])
     coordinates = forms.CoordinatesField(__("Location"), description=__("Pick a location on the map"),
         validators=[forms.validators.Optional(), forms.validators.ValidCoordinates()])
 
@@ -39,6 +42,9 @@ class VenueRoomForm(forms.Form):
     bgcolor = forms.StringField(__("Event Color"),
         validators=[forms.validators.DataRequired(), forms.validators.Length(max=6)],
         description=__("RGB Color for the event. Enter without the '#'. E.g. CCCCCC."), default=u"CCCCCC")
+    seq = forms.IntegerField(__("Sequence"),
+        description=__("The sequence of the room on the listing"),
+        validators=[forms.validators.DataRequired(__("Please specify the sequence order"))])
 
     def validate_bgcolor(self, field):
         if not valid_color_re.match(field.data):
--- a/funnel/models/project.py
+++ b/funnel/models/project.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from werkzeug.utils import cached_property
+from sqlalchemy.ext.orderinglist import ordering_list
 
 from baseframe import __
 
@@ -96,6 +97,9 @@ class Project(UuidMixin, BaseScopedNameMixin, db.Model):
     parent_project = db.relationship('Project', remote_side='Project.id', backref='subprojects')
     inherit_sections = db.Column(db.Boolean, default=True, nullable=False)
     labels = db.Column(JsonDict, nullable=False, server_default='{}')
+
+    venues = db.relationship("Venue", cascade='all, delete-orphan',
+            order_by='Venue.seq', collection_class=ordering_list('seq', count_from=1))
 
     featured_sessions = db.relationship(
         'Session',
--- a/funnel/models/venue.py
+++ b/funnel/models/venue.py
@@ -14,7 +14,7 @@ class Venue(UuidMixin, BaseScopedNameMixin, CoordinatesMixin, db.Model):
 
     project_id = db.Column(None, db.ForeignKey('project.id'), nullable=False)
     project = db.relationship(Project,
-        backref=db.backref('venues', lazy='dynamic', cascade='all, delete-orphan', order_by='Venue.name'))
+        backref=db.backref('venues', lazy='dynamic', cascade='all, delete-orphan', order_by='Venue.seq'))
     parent = db.synonym('project')
     description = MarkdownColumn('description', default=u'', nullable=False)
     address1 = db.Column(db.Unicode(160), default=u'', nullable=False)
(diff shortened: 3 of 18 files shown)


funnel/forms/venue.py
```diff
@@ -28,6 +28,9 @@ class VenueForm(forms.Form):
     country = forms.SelectField(__("Country"),
         validators=[forms.validators.Optional(), forms.validators.Length(max=2)],
         choices=country_codes, default="IN")
+    seq = forms.IntegerField(__("Sequence"),
+        description=__("The sequence of the venue on the listing"),
+        validators=[forms.validators.DataRequired(__("Please specify the sequence order"))])
```
jace: This should not be in the form. We want drag and drop reordering.

funnel/forms/venue.py
```diff
@@ -39,6 +42,9 @@ class VenueRoomForm(forms.Form):
     bgcolor = forms.StringField(__("Event Color"),
         validators=[forms.validators.DataRequired(), forms.validators.Length(max=6)],
         description=__("RGB Color for the event. Enter without the '#'. E.g. CCCCCC."), default=u"CCCCCC")
+    seq = forms.IntegerField(__("Sequence"),
+        description=__("The sequence of the room on the listing"),
+        validators=[forms.validators.DataRequired(__("Please specify the sequence order"))])
```
jace: Likewise.

funnel/models/venue.py
```diff
@@ -14,7 +14,7 @@ class Venue(UuidMixin, BaseScopedNameMixin, CoordinatesMixin, db.Model):
 
     project_id = db.Column(None, db.ForeignKey('project.id'), nullable=False)
     project = db.relationship(Project,
-        backref=db.backref('venues', lazy='dynamic', cascade='all, delete-orphan', order_by='Venue.name'))
+        backref=db.backref('venues', lazy='dynamic', cascade='all, delete-orphan', order_by='Venue.seq'))
```
jace: Use [SQLAlchemy's ordering_list](https://docs.sqlalchemy.org/en/latest/orm/extensions/orderinglist.html) for the relationship.

funnel/models/venue.py
```diff
@@ -24,6 +24,8 @@ class Venue(UuidMixin, BaseScopedNameMixin, CoordinatesMixin, db.Model):
     postcode = db.Column(db.Unicode(20), default=u'', nullable=False)
     country = db.Column(db.Unicode(2), default=u'', nullable=False)
 
+    seq = db.Column(db.Integer, nullable=False, default=0)
```
jace: No default. Let SQLAlchemy supply the value.

funnel/models/venue.py
```diff
@@ -32,11 +34,13 @@ class VenueRoom(BaseScopedNameMixin, db.Model):
 
     venue_id = db.Column(None, db.ForeignKey('venue.id'), nullable=False)
     venue = db.relationship(Venue,
-        backref=db.backref('rooms', cascade='all, delete-orphan', order_by='VenueRoom.name'))
+        backref=db.backref('rooms', cascade='all, delete-orphan', order_by='VenueRoom.seq'))
```
jace: Ordering list here as well.

funnel/models/venue.py
```diff
@@ -32,11 +34,13 @@ class VenueRoom(BaseScopedNameMixin, db.Model):
...
     parent = db.synonym('venue')
     description = MarkdownColumn('description', default=u'', nullable=False)
     bgcolor = db.Column(db.Unicode(6), nullable=False, default=u'229922')
 
+    seq = db.Column(db.Integer, nullable=False, default=0)
```
jace: No default.

migrations/versions/e417a13e136d_venue_venueroom_sequence.py
```diff
@@ -0,0 +1,64 @@
...
+
+    connection = op.get_bind()
+
+    # update venue sequences
+    projects = connection.execute(sa.select([project.c.id]))
+    project_ids = [proj_tuple[0] for proj_tuple in projects]
+
+    for project_id in project_ids:
+        venues = connection.execute(sa.select([venue.c.id]).where(venue.c.project_id == project_id).order_by('created_at'))
+        venue_ids = [venue_tuple[0] for venue_tuple in venues]
```
jace: Not a big deal, but this can be simpler:

```python
venue_ids = [r[0] for r in connection.execute(sa.select([venue.c.id]).where(venue.c.project_id == project_id).order_by('created_at'))]
```

funnel/forms/venue.py
vidya-ram: @jace Do we add the drag and drop as part of room settings on the schedule edit page.


screen shot 2019-02-07 at 2 36 28 pm

funnel/models/venue.py
iambibhas: @jace ordering_list doesn't work properly with backref. it works if I remove backref from here and define `venues` in `Project` model.

funnel/models/project.py
```diff
@@ -97,6 +98,9 @@ class Project(UuidMixin, BaseScopedNameMixin, db.Model):
     inherit_sections = db.Column(db.Boolean, default=True, nullable=False)
     labels = db.Column(JsonDict, nullable=False, server_default='{}')
 
+    venues = db.relationship("Venue", cascade='all, delete-orphan',
```
jace: `'Venue'`

funnel/models/venue.py
```diff
@@ -24,24 +23,59 @@ class Venue(UuidMixin, BaseScopedNameMixin, CoordinatesMixin, db.Model):
...
+            'read': {
+                'id', 'name', 'title', 'description', 'address1', 'address2', 'city', 'state',
+                'postcode', 'country', 'project_details', 'room_list', 'seq', 'uuid'
+            },
+        },
+    }
+
+    @property
+    def project_details(self):
+        return {'name': self.project.name, 'title': self.project.title, 'uuid': self.project.uuid}
```
jace: Aren't we using `suuid` everywhere? It should be here as well.

funnel/models/venue.py
```diff
@@ -24,24 +23,59 @@ class Venue(UuidMixin, BaseScopedNameMixin, CoordinatesMixin, db.Model):
...
+        'all': {
+            'read': {
+                'id', 'name', 'title', 'description', 'bgcolor', 'seq', 'venue_details', 'scoped_name'
+            },
+        },
+    }
+
+    @property
+    def venue_details(self):
+        return {'name': self.venue.name, 'title': self.venue.title, 'uuid': self.venue.uuid}
```
jace: `suuid`

funnel/views/schedule.py
```diff
@@ -143,10 +143,11 @@ class ProjectScheduleView(ProjectViewMixin, UrlForView, ModelView):
     @render_with('schedule.html.jinja2')
     @requires_permission('view')
     def schedule(self):
-        return dict(project=self.obj, venues=self.obj.venues,
+        return dict(project=self.obj,
             from_date=date_js(self.obj.date), to_date=date_js(self.obj.date_upto),
             sessions=session_list_data(self.obj.scheduled_sessions, with_modal_url='view_popup'),
             timezone=timezone(self.obj.timezone).utcoffset(datetime.now()).total_seconds(),
+            venues=[dict(venue.current_access()) for venue in self.obj.venues],
```
jace: Is the `dict(*.current_access())` necessary? It's unnecessary verbosity, so if something is breaking in the absence of the `dict` wrapper, we should fix it.

funnel/views/venue.py
```diff
@@ -65,27 +65,36 @@ def new_venue(self):
...
-        for room in self.obj.rooms:
-            if room.scoped_name in colors:
-                room.bgcolor = colors[room.scoped_name]
-        db.session.commit()
-        return jsonify(status=True)
+    def update_venue_settings(self):
+        if request.json is None:
+            return {'error': 'Invalid data'}, 400
+        for venue_uuid in request.json.keys():
+            venue = Venue.query.filter_by(uuid=venue_uuid).first()
```
jace: Switch to `suuid`

funnel/views/venue.py
```diff
@@ -65,27 +65,36 @@ def new_venue(self):
...
+            venue.seq = request.json[venue_uuid]['seq']
+            db.session.add(venue)
+            for room in request.json[venue_uuid]['rooms']:
+                room_obj = VenueRoom.query.filter_by(name=room['name'], venue=venue).first()
```
jace: Rooms should also get a uuid column.

funnel/views/schedule.py
iambibhas: @jace the template runs `{{ venues|tojson }}` on it. `current_access()` isn't JSON serializable without `dict()`.

jace: We can fix this in Baseframe by replacing the function Jinja2 uses: http://jinja.pocoo.org/docs/2.10/api/#policies

I will file a separate ticket for this.

jace: Flask does its own thing. It's easier to override. http://flask.pocoo.org/docs/1.0/api/#flask.Flask.json_encoder

funnel/views/venue.py
```diff
@@ -80,16 +80,18 @@ def new_venue(self):
     @requires_permission('edit-venue')
     def update_venue_settings(self):
         if request.json is None:
-            return {'error': 'Invalid data'}, 400
+            return {'error': _("Invalid data")}, 400
```
jace: Shouldn't this be `{'error': 'invalid_data', 'error_description': _("Invalid data")}`?

PR #364, opened 2019-02-20
Using current_permissions for comments

iambibhas:


jace:
Doesn't the corresponding view need the same permission test? Is that already in place?

iambibhas:
Yeah, both of those permissions are granted by the comment model and being checked before the action.

PR #366, opened 2019-02-20
Switch to TimezoneType. Resolves #365

### LINKED ISSUE #365: Use timezone field type
The `timezone` field on the Project model should be replaced with the [timezone type from SQLAlchemy Utils](https://sqlalchemy-utils.readthedocs.io/en/latest/_modules/sqlalchemy_utils/types/timezone.html), using the `pytz` backend. This may entail a migration if the field length changes. It will also require changes to all code that handles project timezone as the field will provide the equivalent of `pytz.timezone(project.timezone)`

jace:
This requires testing as I don't have a full fledged development setup on my laptop. From my reading of the TimezoneType source code, it should work.

### DIFF
--- a/funnel/views/session.py
+++ b/funnel/views/session.py
@@ -3,7 +3,7 @@
 from datetime import datetime
 from pytz import timezone
 from baseframe import _
-from flask import request, render_template, jsonify, redirect, abort
+from flask import request, render_template, jsonify, abort
 from coaster.views import route, render_with, requires_permission, UrlForView, ModelView, requestargs
 from coaster.sqlalchemy import failsafe_add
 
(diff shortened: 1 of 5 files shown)


funnel/views/session.py
```diff
@@ -93,15 +93,15 @@ def view(self):
         return dict(project=self.obj.project, active_session=session_data(self.obj, with_modal_url='view_popup'),
             from_date=date_js(self.obj.project.date), to_date=date_js(self.obj.project.date_upto),
             sessions=session_list_data(self.obj.project.scheduled_sessions, with_modal_url='view_popup'),
-            timezone=timezone(self.obj.project.timezone).utcoffset(datetime.now()).total_seconds(),
+            timezone=self.obj.project.timezone.utcoffset(datetime.utcnow()).total_seconds(),
```
jace: This parameter should be `tzoffset`, not `timezone`. @vidya-ram this will need changes in the JS where the name is used.

funnel/views/session.py
```diff
@@ -93,15 +93,15 @@ def view(self):
...
             venues=[dict(venue.current_access()) for venue in self.obj.project.venues],
             rooms=dict([(room.scoped_name, {'title': room.title, 'bgcolor': room.bgcolor}) for room in self.obj.project.rooms]))
 
     @route('viewsession-popup')
     @render_with('session_view_popup.html.jinja2')
     @requires_permission('view')
     def view_popup(self):
-        return dict(session=self.obj, timezone=self.obj.project.timezone, localize_date=localize_date)
+        return dict(session=self.obj, timezone=self.obj.project.timezone.zone, localize_date=localize_date)
```
jace: The `localize_date` function shouldn't be passed around like this. It should be registered as a Jinja2 template helper.

jace:
@vidya-ram 	I've merged this, but the two issues remain to be addressed.

PR #367, opened 2019-02-20
Using pycountry instead of baseframe staticdata

iambibhas:
Relates to https://github.com/hasgeek/baseframe/issues/203

### DIFF
--- a/funnel/forms/venue.py
+++ b/funnel/forms/venue.py
@@ -6,8 +6,8 @@ import baseframe.forms as forms
 from baseframe.forms.sqlalchemy import QuerySelectField
 from .project import valid_color_re
 
-country_codes = [(country.alpha_2, country.name) for country in pycountry.countries]
 __all__ = ['VenueForm', 'VenueRoomForm', 'VenuePrimaryForm']
+country_codes = [(country.alpha_2, country.name) for country in pycountry.countries]
 
 
 class VenueForm(forms.Form):
--- a/funnel/forms/venue.py
+++ b/funnel/forms/venue.py
@@ -33,7 +33,9 @@ class VenueForm(forms.Form):
 
     def set_queries(self):
         pycountry_locale = gettext.translation('iso3166-2', pycountry.LOCALES_DIR, languages=[get_locale()])
-        self.country.choices = [(country.alpha_2, pycountry_locale.gettext(country.name)) for country in pycountry.countries]
+        countries = [(pycountry_locale.gettext(country.name), country.alpha_2) for country in pycountry.countries]
+        countries.sort()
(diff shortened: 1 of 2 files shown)


funnel/forms/venue.py
```diff
@@ -1,12 +1,12 @@
...
 
+import pycountry
 from baseframe import __
 import baseframe.forms as forms
 from baseframe.forms.sqlalchemy import QuerySelectField
-from baseframe.staticdata import country_codes
 from .project import valid_color_re
-from ..models import Venue
 
+country_codes = [(country.alpha_2, country.name) for country in pycountry.countries]
```
jace: This should be below `__all__`

jace: Also, can the names be localized? This will be a roadblock when we introduce a second language in Funnel.

jace: Lastly, this should be sorted by country name. Is that a guarantee from pycountry? If not, we should be doing it here.

iambibhas: It's already sorted by pycountry. And pycountry has locales built-in, but not sure how to integrate that.

funnel/forms/venue.py
```diff
@@ -27,10 +27,14 @@ class VenueForm(forms.Form):
         validators=[forms.validators.Optional(), forms.validators.Length(max=20)])
     country = forms.SelectField(__("Country"),
         validators=[forms.validators.Optional(), forms.validators.Length(max=2)],
-        choices=country_codes, default="IN")
+        choices=[], default="IN")
```
jace: `'IN'`

funnel/forms/venue.py
```diff
@@ -27,10 +27,14 @@ class VenueForm(forms.Form):
...
     country = forms.SelectField(__("Country"),
         validators=[forms.validators.Optional(), forms.validators.Length(max=2)],
-        choices=country_codes, default="IN")
+        choices=[], default='IN')
     coordinates = forms.CoordinatesField(__("Location"), description=__("Pick a location on the map"),
         validators=[forms.validators.Optional(), forms.validators.ValidCoordinates()])
 
+    def set_queries(self):
+        pycountry_locale = gettext.translation('iso3166-2', pycountry.LOCALES_DIR, languages=[get_locale()])
+        self.country.choices = [(country.alpha_2, pycountry_locale.gettext(country.name)) for country in pycountry.countries]
```
jace: This won't be sorted if we're doing localization outside the source list. We need to sort it again.

jace:
I've added sorting. This method is inefficient for a static list, so ideally we should added a `localized_countries` function in Baseframe that caches the result for each locale. Let's do that the next time a country selector is needed anywhere.

PR #368, opened 2019-02-21
Moving url fields to UrlType

iambibhas:


migrations/versions/9aa60ff2a0ea_migrate_to_urltype.py
```diff
@@ -0,0 +1,41 @@
...
+revision = '9aa60ff2a0ea'
+down_revision = 'e3bf172763bc'
+
+from alembic import op
+import sqlalchemy as sa
+from coaster.sqlalchemy import UrlType
+
+
+def upgrade():
+    op.alter_column('profile', 'logo_url', existing_type=sa.Unicode(2000), type_=UrlType)
```
jace: Don't you need `UrlType()` here? Alembic seems to default to instances rather than class.

PR #369, opened 2019-02-26
Hide contact details in Funnel's check-in page

vidya-ram:


funnel/views/helpers.py
```diff
@@ -38,3 +38,18 @@ def send_mail(sender, to, body, subject):
...
+
+    >>> mask_email(u'foobar@example.com')
+    u'foo****@example.com'
+    >>> mask_email(u'not-email')
+    u'not****'
+    """
+    if '@' not in email:
+        return u'{e}****'.format(e=email[0:3])
+    username, domain = email.split('@')
+    return u'{u}****@{d}'.format(u=username[0:3], d=domain)
```
jace: Why four asterisks if only three characters are being masked?

vidya-ram: Fixed

PR #370, opened 2019-02-28
Remove user groups. Resolves #344

### LINKED ISSUE #344: Remove unused vote break-up by usergroup code
We no longer count votes by user groups, so that functionality needs to be removed. We have a version of it in future, but it will need better architecture anyway, so the current code can go. Some code related to this is in `templates/proposals.html.jinja2`.

jace:
Also replaces use of proposal.id with suuid in two places.


jace:
Only 15 user groups existed, ranging from May 2013 to September 2014, plus one abandoned attempt to use them at ReactFoo Hyderabad in January 2018.

PR #382, opened 2019-03-21
hasgeek PWA

vidya-ram:
1. Add worbox to generate service worker
2. Add manifest.json for the PWA
3. Add offline template to be served when user is offline and visits a enpoint not cached.
4. Add endpoints to serve manifest.json. service-worker.js and offline.html
5. Add meta tags to for PWA theme and icons.


funnel/assets/service-worker-template.js
```diff
@@ -0,0 +1,55 @@
...
+workbox.routing.registerRoute(new RegExp('/^https?\:\/\/static.*/'), new workbox.strategies.NetworkFirst({
+  "cacheName": "assets"
+}), 'GET');
+
+workbox.routing.registerRoute(new RegExp('/^https?:\/\/hasgeek.com\/*/'), new workbox.strategies.NetworkFirst({
+  "cacheName": "assets"
+}), 'GET');
+
+//For development setup caching of assets
+workbox.routing.registerRoute(new RegExp('/^http:\/\/localhost:3000\/static/'), new workbox.strategies.NetworkFirst({
```
jace: On Chrome OS, localhost is `penguin.linux.test`, so need `'/http:\/\/penguin.linux.test:3000\/static/'` covered as well.

vidya-ram: Service workers are only available on secure origins and localhost.

PR #384, opened 2019-03-25
added a reserved name list and passing to profile and project model

### LINKED ISSUE #321: URL structure
The transition from Talkfunnel to HasGeek is a good time to revisit Funnel's URL structure. We're operating with a tension between two crucial elements:

1. **Apps** (and sub-apps, _i.e._, **features**) define the functionality available. They are the basis of user engagement, so they need to be front and centre. When you open a native app and it has a collection of tabs (or sidebar entries) for various features, this is an example of apps taking prominence over brands. Apps are typically finite in number, so it's possible to plan exact placement in the UI.

2. **Brands** (users or their creations, and sub-brands, _i.e._, **projects**) host all the content that users come for, so brands must be positioned front and centre, taking prominence over apps even though apps are more foundational. However, brands are infinite, so planned positioning in UI is hard. They have to occupy the content area instead of the UI chrome. And yet, the prominence is necessary for user comfort.

We've attempted to solve this in two ways so far:

1. `app.tld` and `app.tld/brand` URLs, where the app lives in a subdomain (which is technically convenient). Hasjob and Funnel started life like this, at `jobs.hasgeek.com` and `funnel.hasgeek.com` URLs. After a while we moved to distinct app URLs, so both `hacknight.in/jsfoo` and `hasgeek.tv/jsfoo` existed.

    Unfortunately, this means a brand's content is now scattered across apps, making it hard to build a coherent engagement with an audience.

2. `brand.tld` URLs in two variants:

    a. `brand.tld/app/sub-brand` like `cto.hasjob.co/edit/<hashid>`
    b. `brand.tld/sub-brand/app` like `jsfoo.talkfunnel.com/2018/schedule`

With the new HasGeek website, we're attempting consolidation with the following challenges:

1. Apps always come first because they are the foundation for interaction.
2. Engagement begins before the user defines a brand.
3. Sub-brands come even later.
4. And yet the brand and sub-brand must somehow be the most prominent.

This makes Talkfunnel's URL structure painful because it's inverted. In the interest of giving brands the most prominence, it sacrifices the build-up of engagement, requiring a trained brand manager who is committed to engaging. This is HasGeek's captive Media team.

At the crux of the problem (for this ticket) is in handling namespace collisions between brands and apps. We can either (a) reserve a list of app names that are not available as brand/sub-brand names, with future app names potentially conflicting with brand names, or (b) use an escape prefix such that only one escape prefix is reserved and all other names are available to brands.

Eventframe took the escape prefix approach. Since it allowed all of `tld/app`, `tld/brand/app`, `tld/brand/sub-brand/app` (with `app` optional in each), app names were underscore-prefixed like this: `eventframe.hasgeek.com/jsfoo/_edit`

Unfortunately, _apps are also brands in the user's mind_, and underscores or other characters appearing as quirks in a URL can be off-putting. We have to somehow make all of the following possible:

1. `hasgeek.com`
2. `hasgeek.com/app` (with a list of reserved app names such as `events`, `blog`, etc)
3. `hasgeek.com/brand`
4. `hasgeek.com/brand/app` (`hasgeek.com/<username>/proposals` for a speaker's proposals)
5. `hasgeek.com/brand/sub-brand` (`hasgeek.com/jsfoo/2018`)
6. `hasgeek.com/brand/sub-brand/app` (`hasgeek.com/jsfoo/2018/schedule`)
7. `hasgeek.com/brand/sub-brand/app/content` (`hasgeek.com/jsfoo/2018/proposals/<some-proposal-id>`)

Of these, 1, 6 and 7 exist without conflict. 2 and 3 conflict with each other, as do 4 and 5.

To add to the complication, we may have distinct native apps for each feature (say, submitting proposals vs watching videos), but if the app is buried below the brand in the URL, it has to be discoverable with the limited URL routing available in Android and iOS so that any given URL launches the correct app. This needs a clear set of _cross-platform rules_.

iambibhas:
This fixes #321.


funnel/models/helper.py
```diff
@@ -0,0 +1,7 @@
+reserved_words = [
+    'api', 'proposal', 'video', 'workshop', 'project', 'new', 'blog', 'post',
+    'hacknight', 'event', 'brand', 'edit', 'admin', 'section', 'venue', 'ticket',
+    'kharcha', 'funnel', 'hgtv', 'hasjob', 'boxoffice', 'app'
+]
+
+RESERVED_NAMES = reserved_words + [w + 's' for w in reserved_words]
```
jace: Don't do this hack. Put the explicit list of reserved names in the list, one per line (so that future amends will have a clean diff).

The file needs the standard `# -*- coding: utf-8 -*-` on top.

jace: This list of reserved names needs to be incorporated as well. https://github.com/hasgeek/lastuser/blob/a02560240ee40c578cb7e1992621e80681c573d9/instance/settings-sample.py#L44-L68

jace: Make the local list here sorted.

jace:
Add `org`, `orgs`, `delete`.

PR #392, opened 2019-04-08
Implement labels (resolves #313)

### LINKED ISSUE #313: Labels
The `Section` model is a bit crude in that (a) sections must be re-defined for every project, and (b) each proposal can have only one section. Project owners may want more fine-grained control. Enter labels:

1. `Profile` can have zero or more `Labelset` instances attached to it. A labelset contains labels and operates in either checkbox mode (select any number) or radio mode (select one)
2. `Project` can have zero or more labelsets enabled for that particular project. These labels now appear in the UI.
3. `Label` instances are attached to `Labelset` (parent) and to `Proposal` in many-to-many relationships.

Existing sections can be migrated to labelsets and enabled for each project. Some code for this is already in Funnel in `models/labels.py` via commit a4477f27.

## Simplified proposal

1. A `label` model defines a piece of text and an optional colour code.
2. Proposals and labels have a many-to-many relationship.
3. Labels may have sub-labels. When this happens, only one of the sub-labels can be applied to a Proposal.
4. Labels can be parented by `Project` or `Profile` (making them available across all projects), or may be free standing, making them universal across all profiles. For the sake of simplicity, we may limit ourselves to `Project`, or to just `Project` and `Profile`.

jace:
Dependency https://github.com/hasgeek/baseframe/pull/217


jace:
The draft models here do not have colour code columns. This needs some thinking.

1. If the same label (say "Pending review") across projects has different colour codes, it will be confusing to the proposer. Instead, we could auto-calculate colour codes from the text, so that the same text will result in the same colour each time.

2. On the flip side, colours have cultural significance. Green, yellow and red are natural choices for "accepted", "pending" and "rejected". If the auto generator picks the wrong colours, the user will want to override. Users must have a choice, or the auto generator must be trained with a thesaurus.

3. Separately, we could pick a colour palette for the labelset and assign colours to labels (a) sequentially from the palette, or (b) auto pick based on title so that the same title will always pick the same item number in the palette.

funnel/models/project.py
```diff
@@ -485,6 +488,17 @@ def roles_for(self, actor=None, anchors=()):
         roles.update(self.profile.roles_for(actor, anchors))
         return roles
 
+    @with_roles(call={'admin'})
+    def assign_labelset(self, labelset):
```
jace: Where is this method needed? Labelsets never get reassigned. They are permanently a part of the project they were created in.

funnel/views/label.py
```diff
@@ -0,0 +1,164 @@
...
+            ).first_or_404()
+        lset = Labelset.query.filter_by(project=proj, name=labelset).first_or_404()
+        label = self.model.query.filter_by(labelset=lset, name=label).first_or_404()
+        g.profile = label.labelset.project.profile
+        return label
+
+    @route('edit', methods=['GET', 'POST'])
+    @lastuser.requires_login
+    @requires_permission('edit_project')
+    def edit(self):
```
jace: Labels should be immutable. No edit action.

funnel/views/label.py
```diff
@@ -0,0 +1,164 @@
...
+            form.populate_obj(self.obj)
+            db.session.commit()
+            flash(_("Your label has been edited"), 'info')
+            return redirect(self.obj.labelset.project.url_for('labelsets'), code=303)
+        return render_form(form=form, title=_("Edit label"), submit=_("Save changes"))
+
+    @route('delete', methods=['GET', 'POST'])
+    @lastuser.requires_login
+    @requires_permission('edit_project')
+    def delete(self):
```
jace: Labels can be archived, not deleted.

funnel/views/proposal.py
```diff
@@ -175,20 +168,12 @@ def json(self):
...
-            form.section.query = Section.query.filter(Section.project == self.obj.project, Section.public == True).order_by('title')  # NOQA
-        if len(list(form.section.query.all())) == 0:
-            # Don't bother with sections when there aren't any
-            del form.section
+        form = get_proposal_form(ProposalForm, obj=self.obj.formdata, model=Proposal, parent=self.obj.project)
         if self.obj.user != g.user:
             del form.speaking
         if form.validate_on_submit():
             form.populate_obj(self.obj.formdata)
+            form.populate_obj_labels(self.obj)  # this is needed to assign the labels
```
jace: This should not be required if association proxies are used correctly.

funnel/models/label.py
```diff
@@ -85,10 +85,26 @@ class Label(BaseScopedNameMixin, db.Model):
     def restricted(self):
         return self.parent_label._restricted if self.parent_label else self._restricted
 
+    @restricted.setter
+    def restricted(self, value):
+        self._restricted = value
+
+    @restricted.expression
+    def restricted(cls):
+        return cls._restricted == True  # NOQA
```
jace: This is incorrect. It does not match the getter.

funnel/models/label.py
```diff
@@ -85,10 +85,26 @@ class Label(BaseScopedNameMixin, db.Model):
...
     def archived(self):
         return self._archived or self.parent_label._archived if self.parent_label else False
 
+    @archived.setter
+    def archived(self, value):
+        self._archived = value
+
+    @archived.expression
+    def archived(cls):
+        return cls._archived == True  # NOQA
```
jace: Incorrect as it does not match getter.

funnel/models/label.py
```diff
@@ -85,10 +85,26 @@ class Label(BaseScopedNameMixin, db.Model):
     def restricted(self):
         return self.parent_label._restricted if self.parent_label else self._restricted
 
+    @restricted.setter
+    def restricted(self, value):
+        self._restricted = value
```
jace: This should refuse to be set if it has a parent, as the parent's value overrides.

funnel/models/proposal.py
```diff
@@ -88,6 +88,38 @@ def __setattr__(self, attr, value):
...
+        if self.obj is not None:
+            parent_label = Label.query.filter(Label.is_parent == True).filter_by(name=name, project=self.obj.project).first()  # NOQA
+            label = Label.query.filter(Label.parent_label_id == parent_label.id, Label.name == value).first()
+            if label is not None:
+                self.obj.assign_label(label)
+
+
+class ProposalLabelProxy(object):
+    def __get__(self, obj, cls=None):
+        self.obj = obj
```
jace: You can't mutate self since it's an instance on the class.

funnel/models/proposal.py
```diff
@@ -88,6 +88,38 @@ def __setattr__(self, attr, value):
...
+        if obj is not None:
+            return ProposalLabelProxyWrapper(obj)
+
+    def keys(self):
+        return [pl.name for pl in self.obj.project.labels] if self.obj is not None else []
```
jace: This method needs to be in the wrapper.

funnel/models/proposal.py
```diff
@@ -88,6 +88,38 @@ def __setattr__(self, attr, value):
             self.data[attr] = value
 
 
+class ProposalLabelProxyWrapper(object):
+    obj = None
```
jace: Remove this.

funnel/models/proposal.py
```diff
@@ -109,16 +107,15 @@ def __setattr__(self, name, value):
...
 
+    def keys(self):
+        return [pl.name for pl in self.obj.project.labels] if self.obj is not None else []
+
 
 class ProposalLabelProxy(object):
     def __get__(self, obj, cls=None):
-        self.obj = obj
         if obj is not None:
             return ProposalLabelProxyWrapper(obj)
```
jace: `else: return self` required.

funnel/models/label.py
```diff
@@ -1,67 +1,167 @@
...
+    def archived(self, value):
+        if self.parent_label:
+            raise ValueError("Cannot archive a child label")
+        self._archived = value
+
+    @archived.expression
+    def archived(cls):
+        return case([
+            (cls.parent_label_id != None, db.select([Label._archived]).where(Label.id == cls.parent_label_id).as_scalar())  # NOQA
+        ], else_=cls._archived)
```
jace: This is incorrect. Child labels can be archived, and that archived flag takes priority.

funnel/forms/label.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+
+from baseframe import __
+import baseframe.forms as forms
+
+__all__ = ['LabelForm', 'LabelOptionForm']
+
+
+class LabelForm(forms.Form):
+    id = forms.IntegerField("", widget=forms.HiddenInput(), validators=[forms.validators.Optional()])
```
jace: Integer id should never be exposed to the front-end. If this is unavoidable, use UUID primary keys.

jace: Use the `huuid` column represented as a StringField to get a hex representation without dashes.

funnel/forms/label.py
```diff
@@ -0,0 +1,25 @@
...
+        validators=[forms.validators.DataRequired(__(u"This can’t be empty")), forms.validators.Length(max=250)])
+    icon_emoji = forms.StringField("")
+    required = forms.BooleanField(__("Make this label mandatory in proposal forms"), default=False,
+        description=__("If checked, proposers must select one of the options"))
+    restricted = forms.BooleanField(__("Restrict use of this label to editors"), default=False,
+        description=__("If checked, only editors and reviewers can apply this label on proposals"))
+
+
+class LabelOptionForm(forms.Form):
+    id = forms.IntegerField("", widget=forms.HiddenInput(), validators=[forms.validators.Optional()])
```
jace: Likewise

funnel/forms/proposal.py
```diff
@@ -6,36 +6,65 @@
...
+        field_kwargs = {
+            'label': label.form_label_text,
+            'description': label.description,
+            'validators': [forms.validators.DataRequired(__("Please select one"))] if label.required else [],
+        }
+        if label.has_options:
+            FieldType = forms.RadioField
+            field_kwargs['choices'] = [(option.name, option.title) for option in label.options if not option.archived]
+        else:
+            FieldType = forms.BooleanField
```
jace: We are not showing this type in the proposal form, so this does not arise.

funnel/forms/proposal.py
```diff
@@ -6,36 +6,65 @@
...
+
+        if for_admin:
```
jace: Do not reuse forms for different interfaces. Make these separate forms. This form is not suitable for showing on the proposal view page because there is no way to remove an option.

funnel/models/proposal.py
```diff
@@ -354,6 +305,26 @@ def roles_for(self, actor=None, anchors=()):
             roles.remove('reader')  # https://github.com/hasgeek/funnel/pull/220#discussion_r168724439
         return roles
 
+    def assign_label(self, label):
```
jace: IMO, this method should be on the Label class as `apply_to(proposal)`.

jace: The reason being that we treat functionality in layers. Proposals are a base layer of functionality and labels are added on top. Lower layers should not be dependent on upper layers. If we remove some functionality in future, it should be as simple as removing the db model and the corresponding code module.

jace: We also need a corresponding `remove_from(proposal)`.

funnel/views/label.py
```diff
@@ -0,0 +1,194 @@
...
+        form = LabelForm(model=Label, parent=self.obj.parent)
+        emptysubform = LabelOptionForm(MultiDict({}))
+        if form.validate_on_submit():
+            # This form can send one or multiple values for title and icon_emoji.
+            # If the label doesn't have any sublabel, one value is sent for each list,
+            # and those values are also available at `form.data`.
+            # But in case there are sublabels, the sublabel values are in the list
+            # in the order they appeared on the create form.
+            titlelist = request.values.getlist('title')
+            emojilist = request.values.getlist('icon_emoji')
```
jace: Use `zip` for this. It will make the code clearer. https://www.geeksforgeeks.org/zip-in-python/

funnel/views/label.py
```diff
@@ -0,0 +1,194 @@
...
+
+            label = Label(title=form.data.get('title'), icon_emoji=form.data.get('icon_emoji'), project=self.obj)
+            label.restricted = form.data.get('restricted')
+            label.make_name()
+            self.obj.labels.append(label)
+            self.obj.labels.reorder()
+            db.session.add(label)
+
+            for idx, title in enumerate(titlelist):
+                subform = LabelOptionForm(MultiDict({
```
jace: I think the form will accept a regular dict and doesn't need a multidict.

funnel/views/label.py
```diff
@@ -0,0 +1,194 @@
...
+                    'title': titlelist[idx], 'icon_emoji': emojilist[idx]
+                    }), meta={'csrf': False})  # parent form has valid CSRF token
+
+                if not subform.validate():
+                    flash(_("Error with a sublabel: {}").format(subform.errors.pop()), category='error')
```
jace: `subform.errors` is a view. AFAIK, the `pop` could have side effects.

funnel/views/label.py
```diff
@@ -0,0 +1,194 @@
...
+    model = Label
+    route_model_map = {
+        'profile': 'project.profile.name', 'project': 'project.name',
+        'label': 'name'}
+
+    def loader(self, profile, project, label):
+        proj = Project.query.join(Profile).filter(
+                Profile.name == profile, Project.name == project
+            ).first_or_404()
+        label = self.model.query.filter_by(project=proj, name=label).first_or_404()
```
jace: We should be doing `model.query` in the view. Make this a method named `get` in the model.

funnel/views/label.py
```diff
@@ -0,0 +1,194 @@
...
+        return label
+
+    @route('edit', methods=['GET', 'POST'])
+    @lastuser.requires_login
+    @render_with('labels_form.html.jinja2')
+    @requires_permission('edit_project')
+    def edit(self):
+        emptysubform = LabelOptionForm(MultiDict({}))
+        subforms = []
+        if not self.obj.main_label:
```
jace: This is confusing because for a moment I couldn't recall if `main_label` points to the main label, or is a property indicating that this is a main label (which should have been `is_main_label`). Let's add a property named `is_main_label` that simply returns `not self.main_label`.

jace: Or `is_top_level` to make it less ambiguous.

funnel/views/label.py
```diff
@@ -0,0 +1,194 @@
...
+            return redirect(self.obj.project.url_for('labels'), code=303)
+
+        if form.validate_on_submit():
+            idlist = request.values.getlist('id')
+            titlelist = request.values.getlist('title')
+            emojilist = request.values.getlist('icon_emoji')
+
+            idlist.pop(0)
+            titlelist.pop(0)
+            emojilist.pop(0)
```
jace: `for id, title, emoji in zip(idlist, titlelist, emojilist)`

migrations/versions/0b25df40d307_add_labels.py
```diff
@@ -0,0 +1,51 @@
...
+        sa.Column('archived', sa.Boolean(), nullable=False),
+        sa.Column('restricted', sa.Boolean(), nullable=False),
+        sa.Column('required', sa.Boolean(), nullable=False),
+        sa.Column('name', sa.Unicode(length=250), nullable=False),
+        sa.Column('title', sa.Unicode(length=250), nullable=False),
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(['main_label_id'], ['label.id'], ondelete='CASCADE'),
+        sa.ForeignKeyConstraint(['project_id'], ['project.id'], ondelete='CASCADE'),
+        sa.PrimaryKeyConstraint('id'),
+        sa.UniqueConstraint('main_label_id', 'name')
```
jace: This is incorrect. The constraint is `'project_id', 'name'`

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+        sec_count = conn.scalar(
+                sa.select([sa.func.count('*')]).select_from(section).where(section.c.project_id == proj['id'])
+            )
+        if sec_count > 0:
+            # the project has some sections
+            # create section labelset for the project
+            labset = conn.execute(label.insert().values({
+                    'project_id': proj['id'], 'name': u"section", 'title': u"Section",
+                    'seq': 1, 'description': u"", 'restricted': False, 'archived': False,
+                    'required': True, 'created_at': datetime.now(), 'updated_at': datetime.now()
```
jace: `datetime.now` returns time in local timezone. Always use `datetime.utcnow`, but in this case, use `sa.func.now` to push it server side.

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+                lab = conn.execute(label.insert().values({
+                    'main_label_id': labset[0],
+                    'project_id': proj['id'], 'name': sec['name'], 'title': sec['title'],
+                    'seq': index, 'description': u"", 'restricted': False, 'archived': False,
+                    'required': True, 'created_at': datetime.now(), 'updated_at': datetime.now()
+                }).returning(label.c.id, label.c.name)).first()
+
+                proposals = conn.execute(
+                        proposal.select().where(proposal.c.section_id == sec['id']).where(proposal.c.project_id == sec['project_id'])
+                    )
```
jace: This will miss proposals in a project linked to sections in another project -- an error condition resulting from proposals that have been moved around. Leave a comment indicating this happens. We'll come around to finding a solution before deploying.

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+    # Migrate sections -
+    projects = conn.execute(project.select())
+    for proj in projects:
+        sec_count = conn.scalar(
+                sa.select([sa.func.count('*')]).select_from(section).where(section.c.project_id == proj['id'])
+            )
+        if sec_count > 0:
+            # the project has some sections
+            # create section labelset for the project
+            labset = conn.execute(label.insert().values({
```
jace: Instead of `labelset`, call this `sections_label`. We'll need the different labels when dealing with errors like with the proposal sections below.

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+                proposals = conn.execute(
+                        proposal.select().where(proposal.c.section_id == sec['id']).where(proposal.c.project_id == sec['project_id'])
+                    )
+                for prop in proposals:
+                    pl = conn.execute(proposal_label.insert().values({
+                        'proposal_id': prop['id'], 'label_id': lab['id'], 'created_at': datetime.now()
+                    }))
+
+        # technical level
+        labset = conn.execute(label.insert().values({
```
jace: `techlevel_label`

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+            'project_id': proj['id'], 'name': u"technical-level", 'title': u"Technical Level",
```
jace: Sentence case, "Technical level"

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+            'seq': 1, 'description': u"", 'restricted': False, 'archived': False,
```
jace: `'seq': 2` (it's okay if the project didn't have sections and there was no 1)

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+            'required': True, 'created_at': datetime.now(), 'updated_at': datetime.now()
+        }).returning(label.c.id)).first()
+        tl_list = [('beginner', "Beginner"), ('intermediate', "Intermediate"), ('advanced', "Advanced")]
+        for index, tl in enumerate(tl_list):
```
jace: `start=1`

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+                    'seq': index, 'description': u"", 'restricted': False, 'archived': False,
+                    'required': True, 'created_at': datetime.now(), 'updated_at': datetime.now()
+                }).returning(label.c.id, label.c.name)).first()
+
+            proposals = conn.execute(
+                    proposal.select().where(proposal.c.project_id == proj['id']).where(proposal.c.technical_level == tl_title)
+                )
+            for prop in proposals:
+                pl = conn.execute(proposal_label.insert().values({
+                    'proposal_id': prop['id'], 'label_id': lab['id'], 'created_at': datetime.now()
```
jace: `func.now`

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+                }))
+
+        # session type
+        labset = conn.execute(label.insert().values({
+                'project_id': proj['id'], 'name': u"session-type", 'title': u"Session Type",
```
jace: `"Session type"`

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+                'seq': 1, 'description': u"", 'restricted': False, 'archived': False,
```
jace: `'seq': 3`

migrations/versions/e2be4ab896d3_migrate_old_labels.py
```diff
@@ -0,0 +1,149 @@
...
+                'required': True, 'created_at': datetime.now(), 'updated_at': datetime.now()
```
jace: `func.now`

tests/conftest.py
```diff
@@ -1,18 +1,45 @@
...
-from funnel.models import Profile, Project, User
+import uuid
+from funnel import app
+from funnel.models import db, Profile, Project, User, Label, Proposal, Team
+
+
+@app.route('/usertest')
+def user_test():
+    from coaster.auth import current_auth
+    return current_auth.user.username if current_auth.user is not None else "anon"
```
jace: `"<anon>"` so that it has characters that are invalid in usernames. You could have a user named `anon`.

funnel/views/label.py
iambibhas: zip doesn't make much clean as I need to use `enumerate` for sequence. even if I zip, i have to unzip after enumerate.

funnel/forms/label.py
iambibhas: using `name` field instead.

PR #401, opened 2019-04-19
Membership models

jace:


funnel/models/membership.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def revoke(self, actor):
+        self.revoked_at = db.func.utcnow()
+        self.revoked_by = actor
+
+    def replace(self, actor, **roles):
+        if not set(roles.keys()).issubset(self.__role_columns__):
+            raise AttributeError("Unknown role")
+        self.revoked_at = db.func.utcnow()
+        self.revoked_by = actor
```
iambibhas: This can call the revoke function above?

jace: It could.

1. I'm evaluating having an additional column for the revocation reason, to indicate termination or replacement.

2. This code was originally a state manager transition, so one method couldn't call the other. I removed the state manager because it was too verbose for what the model needed.

funnel/models/membership.py
```diff
@@ -0,0 +1,371 @@
...
+    def revoked_by_id(cls):
+        """Id of user who revoked the membership"""
+        return db.Column(
+            None, db.ForeignKey('user.id', ondelete='SET NULL'), nullable=True
+        )
+
+    @declared_attr
+    def revoked_by(cls):
+        """User who revoked the membership"""
+        return db.relationship(User, foreign_keys=[cls.revoked_by_id])
```
jace: We need a matching `granted_by`, to record the identity of the person who created the membership.

funnel/models/membership.py
```diff
@@ -0,0 +1,371 @@
...
+    __parent_column__ = None
+
+    #: Start time of membership, ordinarily a mirror of created_at except
+    #: for records created when the member table was added to the database
+    granted_at = immutable(
+        db.Column(db.TIMESTAMP(timezone=True), nullable=False, default=db.func.utcnow())
+    )
+    #: End time of membership, ordinarily a mirror of updated_at
+    revoked_at = db.Column(db.TIMESTAMP(timezone=True), nullable=True)
+
```
jace: Missing column: `is_invite` (default `False`).

funnel/models/membership.py
```diff
@@ -0,0 +1,371 @@
...
+User.active_profile_owner_memberships = db.relationship(
+    ProfileAdminMembership,
+    lazy='dynamic',
+    primaryjoin=db.and_(
+        ProfileAdminMembership.user_id == User.id,
+        ProfileAdminMembership.active,
+        ProfileAdminMembership.is_owner.is_(True),
+    ),
+)
```
jace: Change all `active` relationships to add a condition: `is_invite.is_(False)`.

funnel/models/membership.py
```diff
@@ -0,0 +1,371 @@
...
+
+Project.active_usher_memberships = db.relationship(
+    ProjectCrewMembership,
+    lazy='dynamic',
+    primaryjoin=db.and_(
+        ProjectCrewMembership.project_id == Project.id,
+        ProjectCrewMembership.active,
+        ProjectCrewMembership.is_usher.is_(True),
+    ),
+)
```
jace: Change all `active` relationships to add a condition: `is_invite.is_(False)`.

funnel/models/membership.py
jace: We forgot about invited memberships entirely in the initial design. Invited memberships are an optional feature that become necessary when an abuse pattern emerges. We have two paths to make invitation-based memberships mandatory:

1. Make invitations a requirement for all members.
2. Make invitations a requirement for some members who have enabled an optional flag in their settings.

The invite flag here solves for path 1 (but not path 2, since it won't stop invite spam). Adding it now makes it easier to turn it on later if required.

Semantics of the flag:

1. When invitations are mandatory, the admin can invite someone to be a member, but can't directly add them. The membership record is created with `is_invite=True, granted_by=current_auth.actor`.

2. The invited user gets a notification (email, push) and a link to an accept or decline page.

3. If the invited user chooses to decline, the existing membership record's `.revoke()` is invoked. This records `revoked_at=db.func.utcnow(), revoked_by=current_auth.actor`. Note that the actor is now the invited user, not the admin who invited them.

4. If the invited user chooses to accept, the existing membership's record's `.replace()` is invoked. This records a revocation as above, plus in the new record `is_invite=False, granted_by=current_auth.actor`. Note that in this case, it will appear that the user granted themselves a membership. The identity of the person who invited them is only available from the previous record. This should be okay as the `granted_by` column is an audit detail and not a product feature.

jace: We could do something better with `is_invite` to make it easier to construct a timeline of membership changes. Instead of a boolean, name it `record_type` with one of the following integer values:

1. Add (directly added as a member)
2. Amend (roles were amended)
3. Invite (user was invited)
4. Accept (user accepted an invite)

A decline/revoke type is not necessary as those are captured on the last membership record via the `revoked_at` and `revoked_by` columns.

If we do this, then we can make a timeline entry stating that "User X accepted an invite" instead of "User X granted themselves some roles". However, it will change the filter conditions elsewhere to explicitly exclude item 3, which is awkward. A boolean condition looks cleaner there. A boolean condition could be mimicked by retaining `is_invite` as a hybrid property that has `return self.record_type == MEMBERSHIP_RECORD_TYPE.INVITE`.

funnel/models/membership.py
```diff
@@ -0,0 +1,436 @@
...
+            nullable=False,
+            index=True,
+        )
+    )
+    user = immutable(
+        db.relationship(
+            User,
+            foreign_keys=[user_id],
+            backref=db.backref(
+                'profile_crew_memberships',
```
jace: Now I remember why we kept the `user` relationship in the subclass instead of the mixin: for the backrefs. Still, this needs a `@with_roles(..., grants={'subject'})`.

funnel/models/membership.py
```diff
@@ -0,0 +1,436 @@
...
+            db.CheckConstraint(
+                'is_editor IS TRUE OR is_concierge IS TRUE OR is_usher IS TRUE',
+                name='project_crew_membership_has_role',
+            )
+        )
+        return tuple(args)
+
+    @property
+    def user_details(self):
+        return {'fullname': self.user.fullname, 'username': self.user.username}
```
jace: Incude `avatar_url`.

funnel/models/membership.py
```diff
@@ -0,0 +1,436 @@
...
+            roles.add('project_editor')
+        if self.is_concierge:
+            roles.add('project_concierge')
+        if self.is_usher:
+            roles.add('project_usher')
+        return roles
+
+    def offered_roles_verbose(self):
+        roles = self.offered_roles()
+        return {role.replace('project_', '').title() for role in roles}
```
jace: Yuck! Make a separate lookup dictionary somewhere. Maybe in the view file.

funnel/models/membership.py
```diff
@@ -0,0 +1,436 @@
...
+
+    def roles_for(self, actor, anchors=()):
+        """Roles available to the specified actor and anchors"""
+        roles = super(ProjectCrewMembership, self).roles_for(actor, anchors)
+        roles.update(self.parent.roles_for(actor, anchors))
```
jace: Change to:
```python
if 'profile_admin' in self.project.profile.roles_for(actor, anchors):
    roles.add('editor')
```
This way it is much clearer where the access rights are coming from.

funnel/models/membership.py
```diff
@@ -0,0 +1,436 @@
...
+
+# Project relationships: all crew, vs specific roles
+
+Project.active_crew_memberships = db.relationship(
+    ProjectCrewMembership,
+    lazy='dynamic',
+    primaryjoin=db.and_(
+        ProjectCrewMembership.project_id == Project.id,
+        ProjectCrewMembership.active,
+        ProjectCrewMembership.is_invite.is_(False),
```
jace: Use `~ProjectCrewMembership.is_invite`.

Using a comparison to False will result in this awkward condition: `(membership.is_invite == TYPE.INVITE) is False` whereas using `~` will result in `not (membership.is_invite == TYPE.INVITE)`, which PostgreSQL will (hopefully) automatically rewrite as `membership.is_invite != TYPE.INVITE)`

jace: Or maybe include an `is_not_invite` hybrid property.

funnel/models/membership.py
jace: ^ This was only a good idea when `is_invite` was a boolean.

funnel/templates/event.html.jinja2
```diff
@@ -142,7 +142,7 @@
 
     var eventConfig = {};
     eventConfig.checkin = {
-      isAdmin: {%- if project.current_roles.admin %} true {% else %} false {%- endif %},
+      isAdmin: {%- if project.current_roles.profile_admin %} true {% else %} false {%- endif %},
```
jace: This has to become `isConcierge` and `project.current_roles.project_concierge`.

funnel/templates/admin.html.jinja2
```diff
@@ -25,12 +25,12 @@
...
           {%- for event in events %}
             <li>
-              <a href="{{ url_for('event', profile=project.profile.name, project=project.name, name=event.name) }}" class="mui--text-subhead">{{ event.title }}</a>{%- if project.current_roles.admin %}&nbsp;<a href="{{ url_for('event_edit', profile=project.profile.name, project=project.name, name=event.name) }}" aria-label="{% trans %}Edit{% endtrans %}" title="Edit"><i class="material-icons mui--text-subhead pull-right" aria-hidden="true">edit</i></a>{%- endif %}
+              <a href="{{ url_for('event', profile=project.profile.name, project=project.name, name=event.name) }}" class="mui--text-subhead">{{ event.title }}</a>{%- if project.current_roles.profile_admin %}&nbsp;<a href="{{ url_for('event_edit', profile=project.profile.name, project=project.name, name=event.name) }}" aria-label="{% trans %}Edit{% endtrans %}" title="Edit"><i class="material-icons mui--text-subhead pull-right" aria-hidden="true">edit</i></a>{%- endif %}
             </li>
           {%- endfor %}
         </ol>
       </div>
-      {%- if project.current_roles.admin %}
+      {%- if project.current_roles.profile_admin %}
```
jace: `profile_admin` will _not_ be available in the project's roles, and also cannot be used for anything other than editing crew memberships.

funnel/templates/membership_add_email.md
```diff
@@ -0,0 +1 @@
+**{{ invited_by.pickername }}** has added you to **{{ project.title }}** as **{{ roles }}**.
```
jace: Use `granted_by` since that is the name in the membership model.

funnel/templates/membership_add_invite_email.md
```diff
@@ -0,0 +1,3 @@
+**{{ invited_by.pickername }}** invited you to join **{{ project.title }}**.
```
jace: `granted_by` for consistency, although it is fairly contorted to say you've been granted an invite.

funnel/templates/membership_add_invite_email.md
```diff
@@ -0,0 +1,3 @@
...
+
+[Accept invite]({{ link }})
```
jace: Include a link to decline.

funnel/templates/membership_revoke_notification_email.md
```diff
@@ -0,0 +1 @@
+You have been removed from **{{ project.title }}** as a member by **{{ revoked_by.pickername }}**.
```
jace: "as a crew member". A variation of this template is required for self removal.

funnel/templates/event.html.jinja2
jace: isAdmin needs to be removed.

funnel/templates/event.html.jinja2
```diff
@@ -142,6 +142,7 @@
     var eventConfig = {};
     eventConfig.checkin = {
       isAdmin: {%- if project.current_roles.profile_admin %} true {% else %} false {%- endif %},
+      isConcierge: {%- if current_auth.permissions.checkin_event %} true {% else %} false {%- endif %},
```
jace: Don't use permissions. Use roles.

funnel/templates/admin.html.jinja2
```diff
@@ -24,7 +24,7 @@
         <ol class="mui-list--aligned">
           {%- for event in events %}
             <li>
-              <a href="{{ url_for('event', profile=project.profile.name, project=project.name, name=event.name) }}" class="mui--text-subhead">{{ event.title }}</a>{%- if project.current_roles.profile_admin %}&nbsp;<a href="{{ url_for('event_edit', profile=project.profile.name, project=project.name, name=event.name) }}" aria-label="{% trans %}Edit{% endtrans %}" title="Edit"><i class="material-icons mui--text-subhead pull-right" aria-hidden="true">edit</i></a>{%- endif %}
+              <a href="{{ event.url_for('view') }}" class="mui--text-subhead">{{ event.title }}</a>{%- if project.current_roles.admin %}&nbsp;<a href="{{ event.url_for('edit') }}" aria-label="{% trans %}Edit{% endtrans %}" title="Edit"><i class="material-icons mui--text-subhead pull-right" aria-hidden="true">edit</i></a>{%- endif %}
```
jace: Drop all references to the admin role.

funnel/models/project_membership.py
```diff
@@ -0,0 +1,187 @@
...
+                'crew_memberships',
+                lazy='dynamic',
+                cascade='all, delete-orphan',
+                passive_deletes=True,
+            ),
+        )
+    )
+    parent = immutable(db.synonym('project'))
+
+    user_id = immutable(
```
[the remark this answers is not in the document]
iambibhas: We will need the backref when we introduce the tab in user account view showing which projects they are a member of. It could be done without backref as well though.

funnel/models/project_membership.py
```diff
@@ -0,0 +1,187 @@
...
+        args.append(
+            db.CheckConstraint(
+                'is_editor IS TRUE OR is_concierge IS TRUE OR is_usher IS TRUE',
+                name='project_crew_membership_has_role',
+            )
+        )
+        return tuple(args)
+
+    @property
+    def user_details(self):
```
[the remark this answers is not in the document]
iambibhas: I think this is needed to show the user details from each membership record which front end wants in form of JSON.

funnel/views/membership.py
```diff
@@ -0,0 +1,306 @@
...
+                            'errors': membership_form.errors,
+                        },
+                        400,
+                    )
+                else:
+                    new_membership = ProjectCrewMembership(
+                        project=self.obj, granted_by=current_auth.user
+                    )
+                    membership_form.populate_obj(new_membership)
+                    new_membership.record_type = MEMBERSHIP_RECORD_TYPE.DIRECT_ADD
```
jace: The default type is DIRECT_ADD. This line is unnecessary.

funnel/views/membership.py
```diff
@@ -0,0 +1,306 @@
...
+    @requires_roles({'editor'})
+    def edit(self):
+        previous_membership = self.obj
+        membership_form = ProjectCrewMembershipForm(obj=previous_membership)
+
+        if request.method == 'POST':
+            if membership_form.validate_on_submit():
+                previous_membership.replace(
+                    actor=current_auth.user,
+                    record_type=MEMBERSHIP_RECORD_TYPE.AMEND,
```
jace: The `replace` method should change the type to `AMEND`. Do not specify it in the view.

migrations/versions/8297ea2c9645_data_migrate_teams_to_membership.py
```diff
@@ -0,0 +1,121 @@
...
+down_revision = '8cb98498a659'
+
+from uuid import uuid4
+
+from alembic import op
+from sqlalchemy.sql import column, table
+from sqlalchemy_utils import UUIDType
+import sqlalchemy as sa  # NOQA
+
+from funnel.models.membership import MEMBERSHIP_RECORD_TYPE
```
jace: Don't import from models in a migration. Reproduce the values here. You don't need the LabeledEnum, just a simple class or dict with name and value.

tests/unit/models/test_membership.py
```diff
@@ -0,0 +1,106 @@
...
+        test_db.session.commit()
+
+        assert 'project_editor' not in new_project.roles_for(new_user)
+        assert 'project_concierge' in new_project.roles_for(new_user)
+        assert 'project_usher' in new_project.roles_for(new_user)
+
+        # let's try replacing the roles in place
+        new_membership3 = new_membership2.replace(
+            actor=new_user2,
+            record_type=MEMBERSHIP_RECORD_TYPE.AMEND,
```
jace: Don't set type here. `replace` should do it.

tests/unit/models/test_membership.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+        test_db.session.commit()
+        assert 'project_editor' in new_project.roles_for(new_user)
+        assert 'project_concierge' not in new_project.roles_for(new_user)
+        assert 'project_usher' not in new_project.roles_for(new_user)
+        assert new_membership3.record_type == MEMBERSHIP_RECORD_TYPE.AMEND
+
+        # replace() can replace a single role as well, rest stays as they were
+        new_membership4 = new_membership3.replace(
+            actor=new_user2, record_type=MEMBERSHIP_RECORD_TYPE.AMEND, is_usher=True
```
jace: As above.

tests/unit/models/test_membership.py
```diff
@@ -0,0 +1,106 @@
...
+        assert 'project_usher' in new_project.roles_for(new_user)
+        # offered_roles should also return all valid roles
+        assert new_membership4.offered_roles() == {'project_editor', 'project_usher'}
+        assert new_membership4.record_type == MEMBERSHIP_RECORD_TYPE.AMEND
+
+        # can't replace with an unknown role
+        with pytest.raises(AttributeError):
+            new_membership2.replace(
+                actor=new_user2,
+                record_type=MEMBERSHIP_RECORD_TYPE.AMEND,
```
jace: As above.

funnel/models/membership.py
```diff
@@ -0,0 +1,140 @@
...
+from sqlalchemy.ext.hybrid import hybrid_property
+
+from baseframe import __
+from coaster.sqlalchemy import StateManager, immutable, with_roles
+from coaster.utils import LabeledEnum
+
+from . import BaseMixin, UuidMixin, db
+from .user import User
+
+__all__ = ['MEMBERSHIP_RECORD_TYPE']
```
jace: Don't export it. It should not be used outside membership model modules.

funnel/models/membership.py
```diff
@@ -0,0 +1,140 @@
...
+        self.revoked_at = db.func.utcnow()
+        self.revoked_by = actor
+        new = type(self)(user=self.user, granted_by=self.granted_by)
+
+        # if existing record type is INVITE, let it be,
+        # unless the new record type is ACCEPT.
+        if self.record_type != MEMBERSHIP_RECORD_TYPE.INVITE:
+            new.record_type = MEMBERSHIP_RECORD_TYPE.AMEND
+        elif record_type == MEMBERSHIP_RECORD_TYPE.ACCEPT:
+            new.record_type = MEMBERSHIP_RECORD_TYPE.ACCEPT
```
jace: This `elif` clause will not be reached because the `if` clause also matches. It is also unnecessary as `ACCEPT` transitions to `AMEND`.

funnel/models/membership.py
```diff
@@ -0,0 +1,140 @@
...
+    @with_roles(call={'editor'})
+    def replace(self, actor, record_type, **roles):
+        if not set(roles.keys()).issubset(self.__data_columns__):
+            raise AttributeError("Unknown role")
+        self.revoked_at = db.func.utcnow()
+        self.revoked_by = actor
+        new = type(self)(user=self.user, granted_by=self.granted_by)
+
+        # if existing record type is INVITE, let it be,
+        # unless the new record type is ACCEPT.
```
jace: Comment doesn't make sense.

funnel/models/membership.py
```diff
@@ -0,0 +1,140 @@
...
+class ImmutableMembershipMixin(UuidMixin, BaseMixin):
+    """
+    Support class for immutable memberships
+    """
+
+    __uuid_primary_key__ = True
+    #: List of columns that will be copied into a new row when a membership is amended
+    __data_columns__ = ()
+    #: Parent column ('profile_id' or 'project_id' in the subclasses)
+    __parent_column__ = None
```
jace: IMO, this should be replaced with a `parent_id` synonym (defined in subclasses), following the pattern established by the `BaseScoped*` mixin classes (but using `parent_id` instead of `parent`, to allow membership amendment without loading the parent object from the db).

funnel/models/membership.py
```diff
@@ -0,0 +1,140 @@
...
+        new = type(self)(user=self.user, granted_by=self.granted_by)
+
+        # if existing record type is INVITE, let it be,
+        # unless the new record type is ACCEPT.
+        if self.record_type != MEMBERSHIP_RECORD_TYPE.INVITE:
+            new.record_type = MEMBERSHIP_RECORD_TYPE.AMEND
+        elif record_type == MEMBERSHIP_RECORD_TYPE.ACCEPT:
+            new.record_type = MEMBERSHIP_RECORD_TYPE.ACCEPT
+
+        setattr(new, self.__parent_column__, getattr(self, self.__parent_column__))
```
jace: This can be removed and the `parent` parameter can be included when `new` is instantiated above.

funnel/models/membership.py
```diff
@@ -0,0 +1,140 @@
...
+            if column in roles:
+                setattr(new, column, roles[column])
+            else:
+                setattr(new, column, getattr(self, column))
+        db.session.add(new)
+        return new
+
+    @with_roles(call={'subject'})
+    def accept(self, actor):
+        return self.replace(actor, record_type=MEMBERSHIP_RECORD_TYPE.ACCEPT)
```
jace: 1. Gate this: `if self.record_type == MEMBERSHIP_RECORD_TYPE.INVITE` required
2. No need to send `record_type` to `replace`. It's supposed to check for you.

funnel/models/profile_membership.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    A users can be an administrator of a profile and optionally an owner.
+    Owners can manage other administrators.
+    """
+
+    __tablename__ = 'profile_admin_membership'
+
+    # List of role columns in this model
+    __data_columns__ = ('is_owner',)
+    __parent_column__ = 'profile_id'
```
jace: Replace with `parent_id = db.synonym('profile_id')` below, after defining the `profile_id` column.

funnel/models/profile_membership.py
```diff
@@ -0,0 +1,169 @@
...
+            Profile,
+            backref=db.backref(
+                'admin_memberships',
+                lazy='dynamic',
+                cascade='all, delete-orphan',
+                passive_deletes=True,
+            ),
+        )
+    )
+    parent = immutable(db.synonym('profile'))
```
jace: Include `parent_id` here.

funnel/models/profile_membership.py
```diff
@@ -0,0 +1,169 @@
...
+                backref=db.backref(
+                    'profile_admin_memberships',
+                    lazy='dynamic',
+                    cascade='all, delete-orphan',
+                    passive_deletes=True,
+                ),
+            )
+        ),
+        grants={'subject'},
+    )
```
jace: This `user_id` column and `user` relationship should move to the mixin class (with backref removed).

funnel/models/profile_membership.py
```diff
@@ -0,0 +1,169 @@
...
+    # Profile roles:
+    is_owner = immutable(db.Column(db.Boolean, nullable=False, default=False))
+
+    @property
+    def user_details(self):
+        return {
+            'fullname': self.user.fullname,
+            'username': self.user.username,
+            'avatar': self.user.avatar,
+        }
```
jace: If this is a recurring use case, move it into the mixin class.

funnel/models/profile_membership.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @property
+    def edit_url(self):
+        return self.url_for('edit', _external=True)
+
+    @property
+    def delete_url(self):
+        return self.url_for('delete', _external=True)
```
jace: I sense we're doing something wrong if these two are required.

funnel/models/membership.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Subclasses must gate these methods in __roles__
+
+    @with_roles(call={'subject', 'editor'})
+    def revoke(self, actor):
+        self.revoked_at = db.func.utcnow()
+        self.revoked_by = actor
+
+    @with_roles(call={'editor'})
+    def replace(self, actor, record_type, **roles):
```
jace: This needs a gate:
```python
if self.revoked_at is not None:
    raise TypeError("This membership record has already been revoked")
```

Documentation suggests TypeError is more appropriate than ValueError here: https://docs.python.org/3/library/exceptions.html#TypeError

funnel/models/membership.py
jace: Cleaner:
```python
if self.record_type == MEMBERSHIP_RECORD_TYPE.INVITE:
    new.record_type = MEMBERSHIP_RECORD_TYPE.ACCEPT
else:
    new.record_type = MEMBERSHIP_RECORD_TYPE.AMEND
```

funnel/models/membership.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def offered_roles(self):
+        """Roles offered by this membership record"""
+        return set()
+
+    # Subclasses must gate these methods in __roles__
+
+    @with_roles(call={'subject', 'editor'})
+    def revoke(self, actor):
+        self.revoked_at = db.func.utcnow()
```
jace: This also needs a gate (as below). Can't revoke twice.

funnel/models/profile_membership.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def offered_roles(self):
+        """Roles offered by this membership record"""
+        roles = {'profile_admin'}
+        if self.is_owner:
+            roles.add('profile_owner')
+        return roles
```
jace: I'm not entirely clear about this, but my sense is that these should be just `admin` and `owner` (assuming nothing else is accidentally granting `owner` on a profile).

The tests for ownership:
1. `Profile.current_roles.owner`
2. `Project.current_roles.profile_owner` (tentative; Project will here be checking the parent and reproducing the role)

funnel/models/profile_membership.py
```diff
@@ -0,0 +1,169 @@
...
+
+# Add active membership relationships to Profile and User
+
+Profile.active_admin_memberships = db.relationship(
+    ProfileAdminMembership,
+    lazy='dynamic',
+    primaryjoin=db.and_(
+        ProfileAdminMembership.profile_id == Profile.id,
+        ProfileAdminMembership.active,
+        ~ProfileAdminMembership.is_invite,
```
jace: _Maybe_ the `active` expression should exclude invites so that this relationship is simpler. As long as nothing else is using `active` to only check for revocation status.

funnel/models/profile_membership.py
```diff
@@ -0,0 +1,169 @@
...
+        ProfileAdminMembership.is_owner.is_(True),
+        ~ProfileAdminMembership.is_invite,
+    ),
+)
+
+User.active_profile_invitations = db.relationship(
+    ProfileAdminMembership,
+    lazy='dynamic',
+    primaryjoin=db.and_(
+        ProfileAdminMembership.user_id == User.id, ProfileAdminMembership.is_invite
```
jace: This relationship is missing a test for not-revoked status.

funnel/models/project_membership.py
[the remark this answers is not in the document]
jace: It's not necessary. The `user` relationship will serialize automatically in a JSON dump. Use Coaster's data profiles. https://github.com/hasgeek/coaster/pull/238

This property is not necessary.

funnel/models/project_membership.py
```diff
@@ -0,0 +1,187 @@
...
+    def offered_roles(self):
+        """Roles offered by this membership record"""
+        roles = set()
+        if self.is_editor:
+            roles.add('project_editor')
+        if self.is_concierge:
+            roles.add('project_concierge')
+        if self.is_usher:
+            roles.add('project_usher')
+        return roles
```
jace: Unlike with the `profile_owner` thing above, here the `editor` and `project_editor` roles need to be kept distinct, as `editor` means anyone who is authorised to edit the project, which includes the concierge role.

For the sake of consistency we should also retain the `profile_*` prefixes.

funnel/models/project_membership.py
```diff
@@ -0,0 +1,187 @@
...
+
+# Project relationships: all crew, vs specific roles
+
+Project.active_crew_memberships = db.relationship(
+    ProjectCrewMembership,
+    lazy='dynamic',
+    primaryjoin=db.and_(
+        ProjectCrewMembership.project_id == Project.id,
+        ProjectCrewMembership.active,
+        ~ProjectCrewMembership.is_invite,
```
jace: Here too we should place the invite record type test inside the active test. It's not an active membership if it's only an invite.

funnel/templates/admin.html.jinja2
jace: This has to be changed.

funnel/templates/admin.html.jinja2
```diff
@@ -41,7 +41,7 @@
     </div>
   </div>
 
- {%- if project.current_roles.admin %}
+ {%- if project.current_roles.profile_admin %}
```
jace: Likewise.

funnel/templates/event.html.jinja2
```diff
@@ -80,7 +80,7 @@
               <td class='js-searchable' data-th="Company">{{ company }}</td>
               <td class='buttongrp-column'>
                 <ul class="mui-list--unstyled">
-                  {{#if isAdmin || isConcierge}}
+                  {{#if isEditor || isConcierge}}
```
jace: For consistency, these two should be `isProjectEditor` and `isProjectConcierge`.

funnel/templates/membership_add_email.md
```diff
@@ -0,0 +1,3 @@
+**{{ granted_by.pickername }}** has added you to **{{ project.title }}** as a crew.
```
jace: "...as a crew member."

funnel/templates/membership_revoke_notification_email.md
```diff
@@ -0,0 +1 @@
+You have been removed from **{{ project.title }}** as a crew member by **{{ revoked_by.pickername }}**.
```
jace: "You have been removed as a crew member from..."

jace: Also, all email templates need `{% trans %}`

funnel/templates/project.html.jinja2
```diff
@@ -102,7 +94,7 @@
                 <a href="{{ url_for('login') }}" class='mui-btn mui-btn--raised mui-btn--primary' data-action="Login to respond">{% trans %}Login to respond{% endtrans %}</a>
               </p>
             {%- endif %}
-            {% if project.current_roles.admin or current_auth.permissions.view_rsvps %}
+            {% if project.current_roles.project_concierge or current_auth.permissions.view_rsvps %}
```
jace: The `or...` part has to be removed. Only roles from now on.

funnel/templates/proposal.html.jinja2
```diff
@@ -118,20 +118,17 @@
...
             </div>
             <div class="card__body">
-              {%- if proposal.current_roles.owner %}
                 <ul class="mui-list--aligned">
                   <li><a href="{{ proposal.url_for('edit') }}">{% trans %}Edit details{% endtrans %}</a></li>
                   <li><a href="{{ proposal.url_for('delete') }}">{% trans %}Delete{% endtrans %}</a></li>
                 </ul>
                 <hr class="mui-divider">
-              {%- endif %}
-              {% if proposal.current_roles.admin %}
```
jace: Proposals don't have an 'admin' role. However, proposals will inherit the `project_editor` role from the project.

funnel/templates/proposals.html.jinja2
```diff
@@ -76,15 +76,7 @@
...
-        <div class="grid__col-xs-12">
-          {% if current_auth.is_authenticated and project.current_roles.admin %}
-            {{ admin_panel(project, cfp_transition_form=cfp_transition_form) }}
-          {%- endif %}
-        </div>
-      </div>
-
-      <div class="grid">
-       {% if project.current_roles.admin or project.current_roles.reviewer or not project.cfp_state.NONE -%}
+       {% if project.current_roles.profile_admin or project.current_roles.reviewer or not project.cfp_state.NONE -%}
```
jace: 1. Remove `profile_admin` here.
2. Projects have an editor role, but not a reviewer role.

funnel/views/profile.py
```diff
@@ -133,7 +133,7 @@ def view(self):
             .all()
         )
         draft_projects = [
-            proj for proj in self.obj.draft_projects if proj.current_roles.admin
+            proj for proj in self.obj.draft_projects if proj.current_roles.profile_admin
```
jace: Change to: `draft_projects = self.obj.draft_projects if self.obj.current_roles.profile_admin else []`

funnel/views/project.py
```diff
@@ -389,7 +390,7 @@ def save(self):
     @route('admin', methods=['GET', 'POST'])
     @render_with('admin.html.jinja2')
     @lastuser.requires_login
-    @requires_permission('checkin_event')
+    @requires_roles({'profile_admin', 'project_usher'})
```
jace: Why does an usher have access to something named `admin`? Which of these is incorrect? The role or the view name?

funnel/views/project.py
```diff
@@ -414,6 +415,25 @@ def admin(self):
         }
 
 
+    @route('settings', methods=['GET', 'POST'])
+    @render_with('settings.html.jinja2')
+    @lastuser.requires_login
+    @requires_roles({'profile_admin', 'project_usher'})
```
jace: Both roles appear to be wrong.

PR #405, opened 2019-05-01
Clear labels from admin panel, unarchive feature

iambibhas:


funnel/views/label.py
```diff
@@ -172,6 +172,22 @@ def archive(self):
...
+        if not self.obj.archived:
+            flash(_("This label is not archived"), category='error')
+        else:
+            form = forms.Form()
+            if form.validate_on_submit():
+                self.obj.archived = False
+                db.session.commit()
+                flash(_("The label has been archived"), category='success')
+            else:
+                flash(_("CSRF token is missing"), category='error')
```
jace: Never tell a user "CSRF token is missing". What is the user supposed to do? There is no CSRF token in the UI. Instead, tell the user to submit again to confirm the action.

funnel/views/proposal.py
```diff
@@ -62,9 +62,9 @@ def proposal_data(proposal):
             ('objective', proposal.objective.html),
             ('description', proposal.description.html),
             ('requirements', proposal.requirements.html),
-            ('slides', proposal.slides.url),
+            ('slides', proposal.slides.url if proposal.slides is not None else ''),
```
jace: This can be achieved with just `or ''`

funnel/views/proposal.py
```diff
@@ -62,9 +62,9 @@ def proposal_data(proposal):
...
             ('links', proposal.links),
-            ('preview_video', proposal.preview_video.url),
+            ('preview_video', proposal.preview_video.url if proposal.preview_video is not None else ''),
```
jace: Same.

funnel/views/proposal.py
iambibhas: That will break if `proposal.slides` is missing/`None`. which is why I added it.

funnel/forms/proposal.py
```diff
@@ -10,6 +10,19 @@
     'ProposalMoveForm', 'ProposalLabelsAdminForm']
 
 
+class ClearableRadioField(forms.RadioField):
```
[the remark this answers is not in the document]
iambibhas: It's not. I got to know about that decision later that fields can not be required in admin widget form. So this field is no longer needed. removing it.

funnel/views/proposal.py
jace: That is what `or ''` solves for.

iambibhas: `proposal.slides.url` raises `AttributeError` is `proposal.slides` is not set.

jace: Ah, missed the `.url` bit.

PR #406, opened 2019-05-06
fixes draft list on profile page

iambibhas:
Also, adds validation to `date` and `date_upto` in ProjectForm as strftime can't handle years smaller than 1900.

### DIFF
--- a/funnel/forms/project.py
+++ b/funnel/forms/project.py
@@ -89,6 +89,16 @@ class ProjectForm(forms.Form):
         if not valid_color_re.match(field.data):
             raise forms.ValidationError("Please enter a valid color code")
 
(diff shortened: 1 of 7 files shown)


funnel/forms/project.py
```diff
@@ -89,6 +89,16 @@ def validate_bg_color(self, field):
         if not valid_color_re.match(field.data):
             raise forms.ValidationError("Please enter a valid color code")
 
+    # We run `strftime()` on `date` and `date_upto` for generating `datelocation`.
+    # And `strftime()` canot handle dates before year 1900.
```
jace: This was fixed in Python 3.2. Another reason to upgrade. https://bugs.python.org/issue1777412

PR #410, opened 2019-05-10
Optional explicitly mentioned in the description of form fields

vidya-ram:


funnel/forms/profile.py
```diff
@@ -20,7 +20,8 @@ class EditProfileForm(forms.Form):
     description = forms.MarkdownField(__("Welcome message"),
         validators=[forms.validators.DataRequired(_(u"Please write a message for the landing page"))],
         description=__("This welcome message will be shown on the landing page."))
-    logo_url = forms.URLField(__("Logo URL"), description=__("Profile logo"),
+    logo_url = forms.URLField(__("Logo URL"),
+        description=__("Optional — Profile logo will be shown on project landing page"),
```
jace: Profile landing page or project landing page?

jace: The dashes need to be n-dashes. (Also below.) That will require either using a unicode string -- `u"Optional – The profile logo will be shown on the profile’s landing page"` -- or using `from __future__ import unicode_literals` at the top of the file.

funnel/forms/participant.py
```diff
@@ -10,11 +10,21 @@
 class ParticipantForm(forms.Form):
     fullname = forms.StringField(__("Fullname"), validators=[forms.validators.DataRequired()])
     email = forms.EmailField(__("Email"), validators=[forms.validators.DataRequired(), forms.validators.Length(max=80)])
-    phone = forms.StringField(__("Phone number"), validators=[forms.validators.Length(max=80)])
-    city = forms.StringField(__("City"), validators=[forms.validators.Length(max=80)])
-    company = forms.StringField(__("Company"), validators=[forms.validators.Length(max=80)])
-    job_title = forms.StringField(__("Job title"), validators=[forms.validators.Length(max=80)])
-    twitter = forms.StringField(__("Twitter"), validators=[forms.validators.Length(max=15)])
+    phone = forms.StringField(__("Phone number"),
+        description=__("Optional"),
```
jace: In these fields we need to explain why they are being asked. Just "Optional" isn't enough.

PR #414, opened 2019-05-14
removing old session cookie

iambibhas:


funnel/__init__.py
```diff
@@ -21,7 +21,6 @@
 pages = FlatPages()
 rq = RQ()
 
-
```
jace: Two blank lines required before a new section.

PR #415, opened 2019-05-14
Deleting cookies based on config

iambibhas:


funnel/views/helpers.py
```diff
@@ -57,8 +57,13 @@ def mask_email(email):
 
 
 def clear_old_session(response):
-    if 'session' in request.cookies and app.config.get('SESSION_COOKIE_NAME') != 'session':
-        response.set_cookie('session', '', expires=0, httponly=True)
+    if app.config.get('DELETE_COOKIES'):
+        for cookie_name in app.config.get('DELETE_COOKIES'):
```
jace: Remove the `if` and do `for cookie_name, domains in app.config.get('DELETE_COOKIES', {}):`

funnel/views/helpers.py
```diff
@@ -57,8 +57,13 @@ def mask_email(email):
...
+            if cookie_name in request.cookies:
+                domain = app.config.get('DELETE_COOKIES')[cookie_name]['domain']
```
jace: `for domain in domains`

instance/hasgeekapp-sample.py
```diff
@@ -12,3 +11,8 @@
 ASSET_MANIFEST_PATH = "static/build/manifest.json"
 
 SESSION_COOKIE_NAME = 'root_session'
+DELETE_COOKIES = {
+    'session': {'domain': '.hasgeekapp.local', 'without_dot': True},
+    'root_session': {'domain': '.hasgeekapp.local', 'without_dot': False},
```
jace: Don't use `without_dot`, just list domains. We may have a future need for specific paths as well.

PR #422, opened 2019-05-31
Removed legacy name column and views related to it

iambibhas:
Fixes #323.

PR #424, opened 2019-05-31
Add support to scan attendee badge for contact exchange

vidya-ram:


funnel/views/index.py
```diff
@@ -34,6 +34,20 @@ def funnelindex():
...
 
+@app.route('/account')
+@funnelapp.route('/account')
+@lastuser.requires_login
+def account():
+    return render_template('account.html.jinja2')
+
+
+@app.route('/contact_exchange')
+@funnelapp.route('/contact_exchange')
```
jace: Whatever name we use here has to be in the reserved names list.

jace: It also needs to be a carefully considered name that accounts for future features. I'd say `contacts` is more appropriate.

funnel/views/participant.py
```diff
@@ -127,37 +128,34 @@ def participant_edit(profile, project, participant):
     return render_form(form=form, title=_(u"Edit Participant"), submit=_(u"Save changes"))
 
 
-@app.route('/<profile>/<project>/participant', methods=['GET', 'POST'])
-@funnelapp.route('/<project>/participant', methods=['GET', 'POST'], subdomain='<profile>')
+@app.route('/participant', methods=['GET', 'POST'])
+@funnelapp.route('/participant', methods=['GET', 'POST'], subdomain='<profile>')
```
jace: This too needs to be under `/contacts`.

funnel/views/participant.py
```diff
@@ -127,37 +128,34 @@ def participant_edit(profile, project, participant):
...
+            contact_exchange = ContactExchange(user=current_auth.actor, participant=participant)
             db.session.add(contact_exchange)
             db.session.commit()
         except IntegrityError:
             current_app.logger.warning(u"Contact Exchange already present")
             db.session.rollback()
         return jsonify(participant=participant_data(participant, project.id, full=True))
     else:
-        return jsonify(message=u"Unauthorized contact exchange", code=401)
+        return make_response(jsonify(status='error', message=u"Unauthorized contact exchange"), 401)
```
jace: This should be `403`. `401` is a login prompt.

funnel/views/participant.py
vidya-ram: Can this be `/contacts/participant`
Can we now remove the GET method of this endpoint?

funnel/templates/contacts.html.jinja2
```diff
@@ -0,0 +1,84 @@
...
+        {{elseif error}}
+          <p class="mui--text-danger">{{ error }}</p>
+        {{/if}}
+        <canvas id="qrreader-canvas"></canvas>
+
+        {{#if showModal}}
+        <div id="status-msg" class="modal" tabindex="-1" role="dialog">
+          <div class="modal__header">
+            <a class="modal__close" href="javascript:void(0);" on-click="closeModal(event)"><i class="material-icons">clear</i></a>
+            <h3 class="mui--text-title">Contact exchange</h3>
```
jace: "Scanned contacts"

It is not an exchange if the other party doesn't also receive contacts. We haven't built this ability yet.

funnel/templates/macros.html.jinja2
```diff
@@ -190,6 +190,9 @@
               <li>
                 <a href="{{ project.url_for('events') }}" title="Scan badges to check-in">{% trans %}Scan badges to check-in{% endtrans %}</a>
               </li>
+              <li>
+                <a href="{{ url_for('contacts') }}" title="Scan badges to check-in">{% trans %}Scan contact{% endtrans %}</a>
```
jace: This is not for check-in.

migrations/versions/c38fa391613f_contact_exchange_model_change.py
```diff
@@ -0,0 +1,58 @@
...
+
+Revision ID: c38fa391613f
+Revises: 252f9a705901
+Create Date: 2019-05-20 14:41:42.347664
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'c38fa391613f'
+down_revision = '252f9a705901'
```
jace: This will need revising when #422 is closed.

funnel/views/index.py
vidya-ram: Updated it `account/contacts`

funnel/views/participant.py
vidya-ram: Add contacts view and updated the endpoint to `/contacts/connect`

jace:
@vidya-ram the contacts UI should indicate the project from which the contact was scanned, and the date of the scan.

PR #429, opened 2019-06-03
Deprecate date and date_upto field from Project model

### LINKED ISSUE #383: Remove Project's date, date_upto and datelocation fields
Project dates should be calculated from the schedule. We can offer cached fields to make queries faster, but these will be datetime fields, not date fields, so `date` and `date_upto` should be replaced with `schedule_start_at` and `schedule_end_at`.

Since the `date` and `date_upto` fields are used in UI for the schedule editor, this needs attention.

The `datelocation` field is also deprecated by the `location` field introduced in #287 and can be removed (data may need to be migrated).

vidya-ram:
Fixes #383.


funnel/forms/project.py
```diff
@@ -34,18 +34,7 @@ class ProjectForm(forms.Form):
...
-        __("End date"),
-        validators=[
-            forms.validators.DataRequired(__("End date is required")),
-            forms.validators.GreaterThanEqualTo('date', __("End date cannot be before start date"))
-            ]
-        )
-    tagline = forms.StringField(__("Tagline"),
-        validators=[forms.validators.DataRequired(), forms.validators.Length(max=250)],
-        filters=[forms.filters.strip()],
+    tagline = forms.StringField(__("Tagline"), validators=[forms.validators.DataRequired(), forms.validators.Length(max=250)],
```
jace: `tagline` needs the `strip` filter.

funnel/models/project.py
```diff
@@ -189,30 +187,32 @@ def datelocation(self):
...
-            strf_date = "%d %b %Y"
-        datelocation = datelocation_format.format(
-            date=self.date.strftime(strf_date),
-            date_upto=self.date_upto.strftime("%d %b"),
-            year=self.date.year)
-        return datelocation if not self.location else u', '.join([datelocation, self.location])
+        daterange = u""
+        if self.schedule_start_at and self.schedule_end_at:
+            daterange_format = u"{schedule_start_at}–{schedule_end_at} {year}"
+            if self.schedule_start_at == self.schedule_end_at:
```
jace: This will not happen since they are datetime, not date. Compare the date values after casting to project timezone.

funnel/models/project.py
```diff
@@ -189,30 +187,32 @@ def datelocation(self):
...
+                # if both dates are same, in case of single day project
+                strf_date = ""
+                daterange_format = u"{schedule_end_at} {year}"
```
jace: Since you are rendering dates instead of datetimes, change the parameter names. I read this code expecting the actual `schedule_end_at` was being rendered here and only after scrolling down realised it was the date portion. Use `start_date` and `end_date`.

funnel/models/project.py
```diff
@@ -189,30 +187,32 @@ def datelocation(self):
...
+            elif self.schedule_start_at.month != self.schedule_end_at.month:
+                # If multi-day event across months
+                strf_date = "%d %b"
+            elif self.schedule_start_at.year != self.schedule_end_at.year:
+                # if the start date and end dates are in different years,
+                strf_date = "%d %b %Y"
+            daterange = daterange_format.format(
+                schedule_start_at=self.schedule_start_at.astimezone(self.timezone).strftime(strf_date),
+                schedule_end_at=self.schedule_end_at.astimezone(self.timezone).strftime("%d %b"),
+                year=self.schedule_start_at.astimezone(self.timezone).year)
```
jace: Year has to be from the end date.

funnel/templates/project.html.jinja2
```diff
@@ -14,7 +14,7 @@
     "@type": "Event",
     "name": "{{ project.title }}",
     "url": "{{ project.url_for(_external=true) }}",
-    "startDate": "{{ project.date }}",
+    "startDate": "{{ project.schedule_start_at }}",
```
jace: These need `.isoformat()` as far as I can tell. Check the spec.

funnel/templates/project.html.jinja2
```diff
@@ -35,7 +35,7 @@
      ],
     {% endif -%}
     "description": "{{ project.tagline|safe }}",
-    "endDate": "{{ project.date_upto }}",
+    "endDate": "{{ project.schedule_end_at }}",
```
jace: Likewise.

funnel/views/project.py
```diff
@@ -28,8 +28,8 @@ def project_data(project):
         'title': project.title,
         'datelocation': project.datelocation,
         'timezone': project.timezone.zone,
-        'start': project.date.isoformat() if project.date else None,
-        'end': project.date_upto.isoformat() if project.date_upto else None,
+        'start': project.schedule_start_at.date().isoformat() if project.schedule_start_at else None,
+        'end': project.schedule_end_at.date().isoformat() if project.schedule_end_at else None,
```
jace: Cast into local timezone first! Use `project.schedule_start_at.astimezone(project.timezone).date()`

funnel/views/schedule.py
```diff
@@ -145,7 +145,7 @@ class ProjectScheduleView(ProjectViewMixin, UrlForView, ModelView):
     def schedule(self):
         schedule_transition_form = ProjectScheduleTransitionForm(obj=self.obj)
         return dict(project=self.obj,
-            from_date=date_js(self.obj.date), to_date=date_js(self.obj.date_upto),
+            from_date=date_js(self.obj.schedule_start_at), to_date=date_js(self.obj.schedule_end_at),
```
jace: The data type is changing from date to datetime. Does this affect the template?

funnel/views/schedule.py
```diff
@@ -193,13 +193,8 @@ def edit_schedule(self):
...
             'scheduled': session_list_data(self.obj.scheduled_sessions, with_modal_url='edit', with_delete_url=True)
             }
-        # Set the proper range for the calendar to allow for date changes
-        first_session = Session.query.filter(Session.scheduled, Session.project == self.obj).order_by(Session.start.asc()).first()
-        last_session = Session.query.filter(Session.scheduled, Session.project == self.obj).order_by(Session.end.desc()).first()
-        from_date = (first_session and first_session.start.date() < self.obj.date and first_session.start) or self.obj.date
-        to_date = (last_session and last_session.start.date() > self.obj.date_upto and last_session.start) or self.obj.date_upto
         return dict(project=self.obj, proposals=proposals,
-            from_date=date_js(from_date), to_date=date_js(to_date),
+            from_date=date_js(self.obj.schedule_start_at), to_date=date_js(self.obj.schedule_end_at),
```
jace: Likewise. Is this compatible?

funnel/views/session.py
```diff
@@ -90,7 +90,7 @@ class SessionView(SessionViewMixin, UrlForView, ModelView):
     @render_with('schedule.html.jinja2', json=True)
     def view(self):
         return dict(project=self.obj.project, active_session=session_data(self.obj, with_modal_url='view_popup'),
-            from_date=date_js(self.obj.project.date), to_date=date_js(self.obj.project.date_upto),
+            from_date=date_js(self.obj.project.schedule_start_at), to_date=date_js(self.obj.project.schedule_end_at),
```
jace: Likewise again.

funnel/views/schedule.py
iambibhas: It doesn't. Previous they were timestamps created from `date` object, now they are timestamps created from `datetime` object.

funnel/templates/project.html.jinja2
iambibhas: Using `isoformat()`.

funnel/templates/project.html.jinja2
iambibhas: Using `isoformat()`.

funnel/models/project.py
```diff
@@ -210,30 +214,34 @@ def datelocation(self):
...
-        elif self.date.year != self.date_upto.year:
-            # if the start date and end dates are in different years,
-            strf_date = "%d %b %Y"
-        datelocation = datelocation_format.format(
-            date=self.date.strftime(strf_date),
-            date_upto=self.date_upto.strftime("%d %b"),
-            year=self.date.year)
-        return datelocation if not self.location else u', '.join([datelocation, self.location])
+        daterange = u""
+        if self.schedule_start_at and self.schedule_end_at:
```
jace: Don't need to test for both, since sessions already enforce that `start_at` and `end_at` must exist together. If there is a `schedule_start_at`, then `schedule_end_at` is also guaranteed to exist.

jace: However, the test is harmless as both properties are cached. **But** you need to test with `is not None` because `date(0, 0, 0)` is falsy.

funnel/models/project.py
```diff
@@ -549,16 +557,14 @@ def roles_for(self, actor=None, anchors=()):
...
+            Project.state.PUBLISHED))  # NOQA
 
 
 Profile.draft_projects = db.relationship(
         Project, lazy='dynamic',
         primaryjoin=db.and_(
             Profile.id == Project.profile_id, Project.parent_id == None,
-            db.or_(Project.state.DRAFT, Project.cfp_state.DRAFT)),
-        order_by=Project.date.desc())  # NOQA
+            db.or_(Project.state.DRAFT, Project.cfp_state.DRAFT)))  # NOQA
```
jace: We should start dropping `# NOQA` in query statements by switching to `column.is_(None)` and `column.isnot(None)`

funnel/templates/project.html.jinja2
```diff
@@ -14,7 +14,7 @@
     "@type": "Event",
     "name": "{{ project.title }}",
     "url": "{{ project.url_for(_external=true) }}",
-    "startDate": "{{ project.date }}",
+    "startDate": "{{ project.schedule_start_at.isoformat() }}",
```
jace: Does this need to be a date or a datetime?

funnel/views/schedule.py
jace: Looks like they need to be cast into the project's timezone though.

PR #430, opened 2019-06-03
Add deprecation notice to Talkfunnel

### LINKED ISSUE #427: Add deprecation notice to Talkfunnel
Talkfunnel needs the following deprecation notice on the home page:

> Talkfunnel will be discontinued in 2020 and is no longer accepting new projects. Please use HasGeek

vidya-ram:
Screen Shot 2019-06-03 at 2 46 30 PM

Fixes #427

### DIFF
--- a/funnel/static/css/app.css
+++ b/funnel/static/css/app.css
@@ -219,6 +219,11 @@ html.touch .collapsible__body--disable {
   background: transparent !important;
 }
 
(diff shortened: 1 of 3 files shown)


jace:
1. This notice should not be dismissable.

2. The notice needs to be at the same or higher level of readability than the text following it. This one is _less_ prominent.

PR #436, opened 2019-06-06
Remove proposal part labels (resolves #403), sections and unused columns

### LINKED ISSUE #403: Remove part labels in project model
As a resolution for #163, renaming the "Objective" and "Description" columns to "Abstract" and "Outline", we introduced a "labels" JSON dictionary in #164.

This is a hack that needs to be removed. Similar to the change in thinking introduced by labels in #313, we have two distinct concerns here:

1. The platform needs a _summary_ to show in restricted spaces such as the proposal list, and a fuller _body_ when the proposal is viewed standalone.

2. The project editor will want customisable fields (first discussed in #109), but this can be satisfied by offering an template for the body, without using custom fields. The word "body" need not appear anywhere. It could simply be the dominant textarea.

To do:

1. [x] Remove the `labels` JsonDict (it may be renamed to `part_labels` by the time this ticket is picked up) and all code that depends on it
2. [x] Rename the `objective` database column to `abstract` (a migration)
3. [x] Introduce proposal templates (maybe in a separate ticket) (now in #437)

Templates could either be fields stored in the Project model that are copied to a proposal, or they could be actual proposals put into a "template" state. When a new proposal is created, the user will be given the option of copying an existing template.

jace:
Also removes the obsolete `session_type` and `technical_level` columns that have been converted to labels as of #392.


jace:
We will need a separate ticket picking up from #403 that replaces _all_ the content fields with these three:

1. `body_text`/`body_html` (Markdown; contains the full proposal)
2. `description` (Text; contains an auto-generated description: the first paragraph of the body with HTML stripped. However the user may optionally edit this description)
3. `description_custom` (Boolean; indicates that the user customised the description and it should not be auto-generated)

PR #466, opened 2019-07-02
Move flash messages to the top of the page

### LINKED ISSUE #457: Flash message placement and color
Flash messages on project page appear after the second navbar, in the middle of the screen. They need to go to the top of the screen. 

Screen Shot 2019-06-28 at 9 25 03 AM

Also, without a category(`success`, `error` etc), the flash message background does not have a color. So it appears that it's just floating there among all the other text with a X button on extreme right.

vidya-ram:
Fixes #457
Depends on https://github.com/hasgeek/baseframe/pull/236

### DIFF
--- a/funnel/static/css/app.css
+++ b/funnel/static/css/app.css
@@ -714,6 +714,9 @@ html.touch .collapsible__body--disable {
   font-size: 24px;
   line-height: 32px;
 }
+.funnel-notice .alert__text .funnel-notice-txt {
+  color: #F0A40C;
+}
 
 .cfp-action-btn {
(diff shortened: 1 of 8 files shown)


jace:
1. Can you include before/after screenshots?
2. Does this fix the bug where flash messages are not shown on the home page and instead appear on the first other page the user visits?

vidya-ram:
Screen Shot 2019-07-03 at 10 24 38 AM
Screen Shot 2019-07-03 at 10 25 08 AM
Screen Shot 2019-07-03 at 10 26 06 AM

PR #468, opened 2019-07-03
UI fixes

vidya-ram:
1. Add username to account profile card
2. Fix size of logo in navbar in mobile and align the navbar items
3. Update the UI of contacts page
4. Change to `description.html` in schedule template
5. On exiting schedule page, save scroll position in browser session storage and scroll to last viewed section of schedule on revisiting the schedule page.

### DIFF
--- a/funnel/assets/js/schedule_view.js
+++ b/funnel/assets/js/schedule_view.js
@@ -22,10 +22,10 @@ const Schedule = {
         headerHeight: '',
         pageDetails: {
           url: window.location.href,
-          title: $('meta[name=DC\\.title]').attr('content'),
+          title: `Schedule – ${schedule.config.projectTitle}`,
           projectTitle: schedule.config.projectTitle,
-          talkTitle: '',
-          description: $('meta[name=description]').attr('content')
+          pageTitle: 'Schedule',
+          description: schedule.config.pageDescription
         },
         getTimeStr(time) {
           return new Date(parseInt(time, 10)).toLocaleTimeString().replace(/(.*)\D\d+/, '$1');
@@ -52,12 +52,12 @@ const Schedule = {
       },
       updateMetaTags: function(pageDetails) {
         $('title').html(pageDetails.title);
-        $('meta[name=DC\\.title]').attr('content', pageDetails.talkTitle);
-        $('meta[property=og\\:title]').attr('content', pageDetails.talkTitle);
+        $('meta[name="DC.title"]').attr('content', pageDetails.pageTitle);
+        $('meta[property="og:title"]').attr('content', pageDetails.pageTitle);
         $('meta[name=description]').attr('content', pageDetails.description);
-        $('meta[property=og\\:description]').attr('content', pageDetails.description);
+        $('meta[property="og:description"]').attr('content', pageDetails.description);
         $('link[rel=canonical]').attr('href', pageDetails.url);
-        $('meta[property=og\\:url]').attr('content', pageDetails.url);
+        $('meta[property="og:url"]').attr('content', pageDetails.url);
       },
       handleBrowserHistory() {
         // On closing modal, update browser history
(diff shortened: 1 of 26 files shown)


funnel/assets/js/schedule_view.js
```diff
@@ -23,7 +23,7 @@ const Schedule = {
         pageDetails: {
           url: window.location.href,
           title: $('title').html(),
-          projectTitle: $('title').html().split(' — ')[1],
+          projectTitle: $('title').html().split(' – ')[1],
```
jace: Don't do this. [Pass the unvarnished title separately in a title meta tag and read from that](https://stackoverflow.com/questions/21076201/difference-between-meta-name-title-tag-and-title-title-tag).

funnel/static/css/app.css
```diff
@@ -279,8 +286,9 @@ html.touch .collapsible__body--disable {
   display: inline-block;
 }
 .hg-app .header--fixed .header__nav .header__right-nav {
-  width: 60%;
+  width: 66.6%;
```
jace: If this is meant to be ⅔, shouldn't it be 66.67%?

funnel/assets/js/schedule_view.js
jace: Use `dc.title` for the meta tag. I think Baseframe does this already, as we've implemented the Dublin Core metadata set since the beginning.

vidya-ram: Fixed. Use `{{ project.title }}` from backend.

funnel/templates/event.html.jinja2
```diff
@@ -1,7 +1,7 @@
 {% extends "layout.html.jinja2" %}
 {% from "macros.html.jinja2" import rsvpform, rsvpscript %}
 {% from "baseframe/forms.html.jinja2" import renderfield %}
-{% block title %}{{ project.title }}{% endblock %}
+{% block title %}Setup events{% endblock %}{% block subtitle %} — {{ project.title }}{% endblock %}
```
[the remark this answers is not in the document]
jace: Found it. In the beginning of each of these templates, add after the extends tag:

```jinja2
{% set subtitle = "whatever" %}
```

And in `layout.html.jinja2`'s `titletags` block, simply use `{% if subtitle %} – {{ subtitle }}{% endif %}`

I think we should call this something other than subtitle, because it's likely to clash with the app's use of subtitles for blurbs (later if not now). This could be `title_suffix` or `title_ext`.

funnel/templates/event.html.jinja2
```diff
@@ -1,7 +1,8 @@
 {% extends "layout.html.jinja2" %}
 {% from "macros.html.jinja2" import rsvpform, rsvpscript %}
 {% from "baseframe/forms.html.jinja2" import renderfield %}
-{% block title %}{{ project.title }}{% endblock %}
+{% block title %}Setup events{% endblock %}
+{% set title_suffix = project.title %}
```
jace: I think this needs to be right below extends, or maybe above it. Can you confirm that it works in this position? If so, it's okay.

funnel/templates/funnelindex.html.jinja2
```diff
@@ -1,6 +1,7 @@
 {% extends "layout.html.jinja2" %}
 
 {% block title %}{% if g.profile %}{{ g.profile.title }}{% else %}{{ config['SITE_TITLE'] }}{% endif %}{% endblock %}
+{% block subtitle %}{% endblock %}
```
jace: This block should not be necessary.

funnel/templates/layout.html.jinja2
```diff
@@ -2,7 +2,7 @@
 {% from "baseframe/components.html.jinja2" import hgtopnav with context %}
 
 {%- block titletags %}
-  <title>{% block title %}{{ title }}{% if config['SITE_TITLE'] %} — {{ config['SITE_TITLE'] }}{% endif %}{% endblock %}</title>
+  <title>{% block title %}{{ title }}{% endblock %}{% block title_suffix %}{% if title_suffix %} – {{ title_suffix }}{% else %} – {{ config['SITE_TITLE'] }}{% endif %}{% endblock %}</title>
```
jace: Don't need a block named `title_suffix`. Just the `if` condition is fine.

funnel/templates/index.html.jinja2
```diff
@@ -1,6 +1,7 @@
 {% extends "layout.html.jinja2" %}
 
 {% block title %}{% if profile %}{{ profile.title }}{% else %}{{ config['SITE_TITLE'] }}{% endif %}{% endblock %}
+{% block subtitle %}{% endblock %}
```
jace: Same here. There is no longer a `subtitle` block.

funnel/templates/event.html.jinja2
vidya-ram: Yes, it works

PR #473, opened 2019-07-08
Black and isort config

jace:


jace:
Flake8 plugins to include in the final PR: https://pypi.org/project/flake8-awesome/

PR #474, opened 2019-07-10
Added redirect for project redirect

iambibhas:


funnel/views/mixins.py
```diff
@@ -19,12 +19,21 @@ def loader(self, profile, project, session=None):
...
 
+    def after_loader(self):
+        if isinstance(self.obj, ProjectRedirect):
+            if self.obj.project:
+                g.profile = self.obj.project.profile
+                return redirect(self.obj.project.url_for())
+            else:
+                abort(410)
+        g.profile = self.obj.profile
+        super(ProjectViewMixin, self).after_loader()
```
jace: Needs a `return` here. `after_loader` is allowed to return a response.

PR #475, opened 2019-07-10
Updated UI of project page

vidya-ram:


funnel/models/project.py
```diff
@@ -550,6 +550,9 @@ def roles_for(self, actor=None, anchors=()):
         roles.update(self.profile.roles_for(actor, anchors))
         return roles
 
+    def is_saved_by(self, user):
+        return self.saved_by.filter_by(user=user).first() is not None
```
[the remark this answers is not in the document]
jace: You need to test if `user is not None` because the call from the view is not doing this.

funnel/views/project.py
```diff
@@ -110,16 +111,24 @@ def view(self):
...
+            'project_currently_saved': project_currently_saved}
 
     @route('proposals')
     @render_with('proposals.html.jinja2')
     @requires_permission('view')
     def view_proposals(self):
         cfp_transition_form = ProjectCfpTransitionForm(obj=self.obj)
-        return {'project': self.obj, 'cfp_transition_form': cfp_transition_form}
+        project_save_form = ProjectSaveForm()
+        project_currently_saved = self.obj.is_saved_by(current_auth.user)
```
jace: Instead of repeating this code, move it into a method in the class. Say, `self.project_details()`, which gives you a dict that you can add more attributes to before returning to `render_with`.

funnel/views/project.py
```diff
@@ -263,6 +280,7 @@ def schedule_transition(self):
     @requires_permission('view')
     def rsvp(self):
         form = RsvpForm()
+        print form.data
```
jace: Remove print statement.

funnel/views/project.py
jace: You could also delegate this to the template. It can call `current_view.save_details()` if and only if required, saving you some computation.

funnel/views/project.py
```diff
@@ -281,6 +299,42 @@ def rsvp(self):
...
+                if proj_save is None:
+                    proj_save = SavedProject(user=current_auth.user, project=self.obj)
+                    form.populate_obj(proj_save)
+                    db.session.commit()
+                    return {'status': 'success'}
+                else:
+                    return {
+                        'status': 'error', 'error_identifier': 'project_save_exists',
+                        'error_description': _("You have already saved this project")
+                        }, 400
```
jace: This is not an error. For idempotency you return exactly like if the operation was successful.

funnel/views/project.py
```diff
@@ -281,6 +299,42 @@ def rsvp(self):
...
+            else:
+                if proj_save is not None:
+                    db.session.delete(proj_save)
+                    db.session.commit()
+                    return {'status': 'success'}
+                else:
+                    return {
+                        'status': 'error', 'error_identifier': 'project_save_invalid',
+                        'error_description': _("You have not saved this project yet")
+                        }, 400
```
jace: Likewise.

funnel/views/project.py
```diff
@@ -281,6 +299,42 @@ def rsvp(self):
...
+    def save(self):
+        form = ProjectSaveForm()
+        if form.validate_on_submit():
+            proj_save = SavedProject.query.filter_by(user=current_auth.user, project=self.obj).first()
+            if form.save.data:
+                if proj_save is None:
+                    proj_save = SavedProject(user=current_auth.user, project=self.obj)
+                    form.populate_obj(proj_save)
+                    db.session.commit()
+                    return {'status': 'success'}
```
jace: The valid values for status (by convention) are `ok` and `error`. [We've been using it a while](https://github.com/hasgeek/lastuser/blame/18e3af9259dd81913d9c70a725eaed19fba09d05/lastuser_oauth/views/resource.py#L114).

funnel/views/project.py
```diff
@@ -281,6 +299,42 @@ def rsvp(self):
...
+                else:
+                    return {
+                        'status': 'error', 'error_identifier': 'project_save_exists',
+                        'error_description': _("You have already saved this project")
+                        }, 400
+            else:
+                if proj_save is not None:
+                    db.session.delete(proj_save)
+                    db.session.commit()
+                    return {'status': 'success'}
```
jace: `{'status': 'ok'}`

funnel/views/project.py
jace: Use `'error'` as the key, not `'error_identifier'` (moot in this case because it's not to be used).

funnel/views/project.py
```diff
@@ -281,6 +299,42 @@ def rsvp(self):
...
+                else:
+                    return {
+                        'status': 'error', 'error_identifier': 'project_save_invalid',
+                        'error_description': _("You have not saved this project yet")
+                        }, 400
+        else:
+            return {
+                'status': 'error', 'error_identifier': 'project_save_form_invalid',
```
jace: As above.

funnel/views/session.py
```diff
@@ -165,6 +166,40 @@ def feedback(self, id_type, userid, content, presentation, min_scale=0, max_scal
...
+        if form.validate_on_submit():
+            session_save = SavedSession.query.filter_by(user=current_auth.user, session=self.obj).first()
+            if session_save is not None:
+                db.session.delete(session_save)
+                flash(_(u"You are no longer subscribed to this session"), category='success')
+            else:
+                flash(_(u"Invalid session"), category='error')
+        else:
+            flash(_(u"Cannot unsubscribe from this session: {}".format(", ".join(form.errors))), category='error')
+        return redirect(self.obj.url_for(), code=303)
```
jace: These two methods need to be merged as with the SavedProject view.

funnel/forms/session.py
```diff
@@ -21,3 +23,11 @@ class SessionForm(forms.Form):
     featured = forms.BooleanField(__("This is a featured session"), default=False)
     start_at = forms.HiddenField(__("Start Time"), validators=[forms.validators.DataRequired()])
     end_at = forms.HiddenField(__("End Time"), validators=[forms.validators.DataRequired()])
+
+
+class SavedSessionForm(forms.Form):
+    save = forms.BooleanField(
+        __("Save this session?"),
+        validators=[forms.validators.InputRequired()],
```
jace: This validator cannot be used for HTML form submissions. It's JSON-only. Your options are:

1. Don't support HTML forms at all, requiring JSON-only (and thereby JS-only)
2. Drop the validator so that it works for both. However, the absence of a `save` attribute will imply removal, which will be unexpected behaviour for a JSON submission.
3. Use a `SelectField` with choices False and True, and use `coerce=coaster.utils.getbool` so that it works for HTML form submission.

funnel/forms/project.py
```diff
@@ -126,6 +134,14 @@ def set_queries(self):
         self.cfp_transition.choices = self.edit_obj.cfp_state.transitions().items()
 
 
+class SavedProjectForm(forms.Form):
+    save = forms.BooleanField(
+        __("Save this project?"),
+        validators=[forms.validators.InputRequired()],
```
jace: Same concern as with the `SavedSessionForm` validator below.

funnel/models/project.py
```diff
@@ -387,17 +392,22 @@ def calendar_weeks(self):
         #         Session.scheduled
         #         ).group_by(db.text('date')).order_by(db.text('date'))
 
+        # if the project is within 2 weeks, send current week as well
```
jace: This is incorrect. It checks if the current date is 14 days from the start date. What we want to know is whether the current date is within two weeks (Mon-Fri) of the event's week. If Fifthel starts June 25 (Thursday), then this code will start showing the current date on June 11, which is in the middle of a week. It should start showing from the Monday itself.

Use `isoweek` to get the first day of the week in which the schedule starts, and compare if it's within 14 days of the date on which the current week starts. Or if isoweek supports week math directly, use it. Don't compare week numbers as that will fail around a new year.

jace: Looks like you can indeed do `Week.withdate(datetime.now(project.timezone).date()) in (event_week - 1, event_week - 2)`.

I'm not sure, but this may work too: `current_week >= event_week - 2`

funnel/models/project.py
```diff
@@ -374,7 +379,7 @@ def calendar_weeks(self):
             SELECT DATE_TRUNC('day', "start_at" AT TIME ZONE :timezone) AS date, COUNT(*) AS count
             FROM "session" WHERE "project_id" = :project_id AND "start_at" IS NOT NULL AND "end_at" IS NOT NULL
             GROUP BY date ORDER BY date;
-            ''')).params(timezone=self.timezone.zone, project_id=self.id)
+            ''')).params(timezone=self.timezone.zone, project_id=self.id).all()
```
jace: Since you are modifying the list, don't use `.all()`. Use `list(...)` instead. It ensures you have a local list and aren't mutating someone else's list.

funnel/models/project.py
jace: I can't remember if we had a discussion on `datetime.now(timezone)` vs `utcnow().astimezone(timezone)`. I'm okay with either. The first is standard Python, the second is one less import but one more function call.

funnel/templates/proposals.html.jinja2
```diff
@@ -5,9 +5,40 @@
 {% block description %}{{ project.title }}{% if g.profile %} &ndash; {{ g.profile.description|striptags }}{% endif %}{% endblock %}
 
 {%- block pageheaders %}
-  {% assets "css_leaflet" -%}
-    <link rel="stylesheet" type="text/css" href="{{ ASSET_URL }}" />
-  {%- endassets -%}
+  <script type="application/ld+json">
+  {
+    "@context": "http://schema.org",
+    "@type": "Event",
```
jace: Event schema should be in the project page and maybe the schedule page. It's not required for the proposals page.

funnel/templates/proposals.html.jinja2
```diff
@@ -5,9 +5,40 @@
...
+    "name": {{ project.title|tojson }},
+    "url": {{ project.url_for(_external=true)|tojson }},
+    "startDate": {% if project.schedule_start_at %}{{ project.schedule_start_at.astimezone(project.timezone).isoformat()|tojson }}{% else %}null{% endif %},
```
jace: Apparently `startDate` is mandatory for the event schema, so this `if` condition should be around the entire schema tag. Don't include the schema if there are no scheduled dates.

funnel/templates/schedule.html.jinja2
```diff
@@ -15,6 +15,43 @@
...
 {% endblock %}
 
+{%- block pageheaders %}
+  <script type="application/ld+json">
+  {
+    "@context": "http://schema.org",
+    "@type": "Event",
+    "name": {{ project.title|tojson }},
+    "url": {{ project.url_for(_external=true)|tojson }},
+    "startDate": {% if project.schedule_start_at %}{{ project.schedule_start_at.astimezone(project.timezone).isoformat()|tojson }}{% else %}null{% endif %},
```
jace: This `if` condition should move to cover the entire tag.

funnel/views/project.py
```diff
@@ -281,6 +298,45 @@ def rsvp(self):
...
+                    proj_save = SavedProject(user=current_auth.user, project=self.obj)
+                    form.populate_obj(proj_save)
+                    db.session.commit()
+                    return {'status': 'ok'}
+                else:
+                    return {
+                        'status': 'error',
+                        'error': 'project_save_exists',
+                        'error_description': _("You have already saved this project")
+                    }
```
jace: Remove this. Breaks idempotency. The operation is successful even if no action was required. You could distinguish by returning HTTP status code 201 vs 200, with the same `{'status': 'ok'}` for both.

funnel/views/project.py
```diff
@@ -281,6 +298,45 @@ def rsvp(self):
...
+                if proj_save is not None:
+                    db.session.delete(proj_save)
+                    db.session.commit()
+                    return {'status': 'ok'}
+                else:
+                    return {
+                        'status': 'error',
+                        'error': 'project_save_invalid',
+                        'error_description': _("You have not saved this project yet")
+                    }
```
jace: Same as above.

funnel/views/session.py
```diff
@@ -165,6 +174,45 @@ def feedback(self, id_type, userid, content, presentation, min_scale=0, max_scal
...
+                    session_save = SavedSession(user=current_auth.user, session=self.obj)
+                    form.populate_obj(session_save)
+                    db.session.commit()
+                    return {'status': 'ok'}
+                else:
+                    return {
+                        'status': 'error',
+                        'error': 'session_save_exists',
+                        'error_description': _("You have already saved this session")
+                    }
```
jace: Same as above. Remove as it breaks idempotency.

funnel/views/session.py
```diff
@@ -165,6 +174,45 @@ def feedback(self, id_type, userid, content, presentation, min_scale=0, max_scal
...
+                if session_save is not None:
+                    db.session.delete(session_save)
+                    db.session.commit()
+                    return {'status': 'ok'}
+                else:
+                    return {
+                        'status': 'error',
+                        'error': 'session_save_invalid',
+                        'error_description': _("You have not saved this session yet")
+                    }
```
jace: As above.

funnel/views/project.py
jace: This still applies -- SavedProjectForm is required by every view under Project, as the save button will appear in all of them (in desktop view).

funnel/views/session.py
```diff
@@ -102,13 +103,21 @@ class SessionView(SessionViewMixin, UrlForView, ModelView):
...
+            'project': self.obj.project,
+            'active_session': session_data(self.obj, with_modal_url='view_popup'),
+            'from_date': date_js(self.obj.project.schedule_start_at),
+            'to_date': date_js(self.obj.project.schedule_end_at),
+            'sessions': session_list_data(self.obj.project.scheduled_sessions, with_modal_url='view_popup'),
             # FIXME: This timezone by UTC offset is not accounting for DST. Look up where it's being used and fix it
-            timezone=utcnow().astimezone(self.obj.project.timezone).utcoffset().total_seconds(),
-            venues=[venue.current_access() for venue in self.obj.project.venues],
-            rooms=dict([(room.scoped_name, {'title': room.title, 'bgcolor': room.bgcolor}) for room in self.obj.project.rooms]))
+            'timezone': utcnow().astimezone(self.obj.project.timezone).utcoffset().total_seconds(),
```
jace: The `FIXME` can be resolved if we replace `utcnow()` with `project.schedule_start_at` (if present, falling back to `utcnow`). Also, the parameter should be called `tzoffset`, not `timezone`. Templates need a corresponding update @vidya-ram.

funnel/models/project.py
```diff
@@ -387,17 +393,29 @@ def calendar_weeks(self):
...
+        # if the project's week is within next 2 weeks, send current week as well
+        now = utcnow().astimezone(self.timezone)
+
+        if self.schedule_start_at is not None:
+            current_week = Week.withdate(now)
+            schedule_start_week = Week.withdate(self.schedule_start_at)
+
+            if schedule_start_week > current_week and (schedule_start_week - current_week) <= 2:
+                if (schedule_start_week - current_week) == 2:
+                    session_dates.insert(0, (now + timedelta(days=7), 0))
```
jace: Something about this construct looks off. It needs a comment, and possibly cleaner nesting of `if` statements. However, this isn't a blocking issue.

PR #482, opened 2019-07-23
Comment email fix

iambibhas:
Not tested yet. This fixes the issue where after a proposal is transfered to a new speaker, comment emails still going to the creator.


funnel/templates/proposal.html.jinja2
```diff
@@ -238,8 +238,6 @@
           <div class="mui--hide">
             <input type="hidden" name="form.id" value="newcomment"/>
             {{ commentform.hidden_tag() }}
-            {{ commentform.parent_id() }}
-            {{ commentform.comment_edit_id() }}
```
jace: Were these redundant? `hidden_tag` is supposed to include all hidden fields, so if we were repeating these fields, that's a good catch.

funnel/views/commentvote.py
```diff
@@ -93,27 +93,27 @@ def new_comment(self):
                 if commentform.parent_id.data:
                     parent = Comment.query.filter_by(suuid=commentform.parent_id.data).first_or_404()
                     if parent.user.email:  # FIXME: https://github.com/hasgeek/funnel/pull/324#discussion_r241270403
-                        if parent.user == self.obj.user:  # check if parent comment & proposal owner are same
-                            if not current_auth.user == parent.user:  # check if parent comment is by proposal owner
-                                send_mail_info.append({'to': self.obj.user.email or self.obj.email,
-                                    'subject': u"{project} Funnel: {proposal}".format(project=self.obj.project.title, proposal=self.obj.title),
+                        if parent.user == self.obj.speaker:  # check if parent comment is by the proposal owner
```
jace: Use `owner`. It will return `speaker` or if missing, `user`.

funnel/views/commentvote.py
```diff
@@ -125,6 +125,8 @@ def new_comment(self):
                 if item.get('to'):
                     # Sender is set to None to prevent revealing email.
                     send_mail(sender=None, body=email_body, **item)
+        else:
+            flash(_("Something went wrong, please try again"), category='error')
```
jace: Errors are idempotent. If something went wrong, it will go wrong again. This needs a more accurate error message.

funnel/templates/proposal.html.jinja2
iambibhas: Yeah. These were redundant. `hidden_tag` already had these.

funnel/views/commentvote.py
iambibhas: This is validation error. Comment form can't show validation error right now. It needs to be redone to show these. I'll check if it can be done with this PR.

funnel/forms/proposal.py
```diff
@@ -93,7 +93,9 @@ class ProposalLabelAdminForm(forms.Form):
 
 class ProposalTransferForm(forms.Form):
     user = forms.UserSelectField(
-        __("Transfer to"), validators=[forms.validators.DataRequired()]
+        __("Transfer to"),
+        description=__("Transfer this proposal to another speaker"),
+        validators=[forms.validators.DataRequired()],
```
jace: Minor style issue: WTForms suggests that label and validators are always the first two parameters. https://wtforms.readthedocs.io/en/stable/fields.html#fields-with-custom-constructors.

jace: To me pairing label and description makes more sense, so we could ignore the WTForms recommendation and order them by label, description, filters and validators (since filters are applied before validation).

funnel/views/commentvote.py
```diff
@@ -171,6 +171,10 @@ def new_comment(self):
                 if item.get('to'):
                     # Sender is set to None to prevent revealing email.
                     send_mail(sender=None, body=email_body, **item)
+        else:
+            # FIXME: ValidationError here. But we cannot return the error as
+            # it redirects to the proposal page, where the comment form is.
+            flash(_("Something went wrong, please try again"), category='error')
         # Redirect despite this being the same page because HTTP 303 is required to not break
         # the browser Back button
         return redirect(to_redirect, code=303)
```
jace: This needs to be an Ajax call. There's no need to do a full page submit anymore.

funnel/views/commentvote.py
jace: We should render the errors via the flash message for now, but the correct fix is to turn this into an Ajax call.

funnel/forms/proposal.py
iambibhas: That's when you subclass the field to make a custom field. As we're explicitly passing `validators=`, it should be fine. I like label and description together as well.

funnel/forms/comment.py
```diff
@@ -15,6 +15,17 @@ class CommentForm(forms.Form):
         validators=[forms.validators.DataRequired()],
     )
 
+    def get_verbose_errors(self):
+        verbose_error_list = []
+        for field_name, errors in self.errors.items():
+            for err in errors:
```
jace: `getitem` call needs to be here. `field = getattr(self, field_name)`.

funnel/views/commentvote.py
```diff
@@ -99,42 +100,41 @@ def new_comment(self):
...
+                        if parent.user == self.obj.owner:
+                            # parent comment is by the proposal owner
+                            if not parent.user == current_auth.user:
+                                # parent comment is not by the curernt user
                                 send_mail_info.append(
                                     {
-                                        'to': self.obj.user.email or self.obj.email,
-                                        'subject': u"{project} Funnel: {proposal}".format(
+                                        'to': self.obj.owner.email or self.obj.email,
+                                        'subject': u"[New Comment] {project}: {proposal}".format(
```
jace: Sentence case: `[New comment]`

funnel/views/commentvote.py
```diff
@@ -99,42 +100,41 @@ def new_comment(self):
...
                                 )
-                        else:  # send mail to parent comment owner & proposal owner
+                        else:
                             if not parent.user == current_auth.user:
+                                # send mail to parent comment owner
                                 send_mail_info.append(
                                     {
                                         'to': parent.user.email,
-                                        'subject': u"{project} Funnel: {proposal}".format(
+                                        'subject': u"[New Comment] {project}: {proposal}".format(
```
jace: Likewise

jace: What if we use 💬 instead of `[New comment]`?

funnel/views/commentvote.py
```diff
@@ -99,42 +100,41 @@ def new_comment(self):
...
                                 )
-                            if not self.obj.user == current_auth.user:
+                            if not self.obj.owner == current_auth.user:
+                                # send mail to proposal owner
                                 send_mail_info.append(
                                     {
-                                        'to': self.obj.user.email or self.obj.email,
-                                        'subject': u"{project} Funnel: {proposal}".format(
+                                        'to': self.obj.owner.email or self.obj.email,
+                                        'subject': u"[New Comment] {project}: {proposal}".format(
```
jace: Sentence case or 💬

funnel/views/commentvote.py
```diff
@@ -99,42 +100,41 @@ def new_comment(self):
                     parent = Comment.query.filter_by(
```
jace: This one has been tripping me every time I look at it. I keep thinking the "parent" is the proposal, but it's the parent comment. We've made this mistake in the model itself where the column is called "parent_id". We should:

1. Rename the "parent" variable here to "in_reply_to_comment"
2. In a separate PR, rename the column to "in_reply_to_id" and replace all references to "parent" with "in_reply_to" (the term "In-Reply-To" is an email header to indicate a reply, with the reference message id.)

jace: Please file a ticket for the rename.

funnel/views/commentvote.py
```diff
@@ -145,11 +145,11 @@ def new_comment(self):
...
                         comment.parent = parent
                 else:  # for top level comment
-                    if not self.obj.user == current_auth.user:
+                    if not self.obj.owner == current_auth.user:
                         send_mail_info.append(
                             {
-                                'to': self.obj.user.email or self.obj.email,
-                                'subject': u"{project} Funnel: {proposal}".format(
+                                'to': self.obj.owner.email or self.obj.email,
+                                'subject': u"[New Comment] {project}: {proposal}".format(
```
jace: Here too.

funnel/views/commentvote.py
iambibhas: Using `[💬]`. Let's try it for sometime. I'll remove the `[` `]` if they dont look good. Only `💬` before the subject wasn't looking good to me.

funnel/views/commentvote.py
```diff
@@ -109,7 +109,7 @@ def new_comment(self):
                                 send_mail_info.append(
                                     {
                                         'to': self.obj.owner.email or self.obj.email,
-                                        'subject': u"[New Comment] {project}: {proposal}".format(
+                                        'subject': u"[💬] {project}: {proposal}".format(
```
jace: The brackets are not needed when using emoji. Can you remove them? @iambibhas

iambibhas: Removed them. I kept them because otherwise it seems a bit lacking in clues of these being emails about comments. The bubble right beside the project name and proposal name seems so tiny.

PR #484, opened 2019-07-24
Current sessions

iambibhas:


funnel/models/project.py
```diff
@@ -436,6 +436,27 @@ def calendar_weeks(self):
             ]
         }
 
+    @property
+    def current_sessions(self):
```
jace: This is too expensive to be a property. It should be a method.

funnel/models/project.py
```diff
@@ -436,6 +436,27 @@ def calendar_weeks(self):
...
+        from ..models import Session
```
jace: Move the import to the bottom of the file. Importing within the method has a per-call cost. Annotate with `# isort:skip` to stop isort from moving it up.

PR #485, opened 2019-07-25
Add session url to session title

vidya-ram:


funnel/models/session.py
```diff
@@ -39,35 +45,60 @@ class Session(UuidMixin, BaseScopedIdNameMixin, db.Model):
...
+                'speaker_bio',
+                'start_at',
+                'end_at',
+                'venue_room',
+                'is_break',
                 'banner_image_url',
-                }
-            }
+            },
+            'call': {'url_for'},
```
iambibhas: This is the actual change. Rest is black formatted.

PR #499, opened 2019-09-10
using state transition for rsvp

iambibhas:


funnel/models/rsvp.py
```diff
@@ -40,23 +38,60 @@ class Rsvp(TimestampMixin, db.Model):
         'state',
         db.CHAR(1),
         StateManager.check_constraint('state', RSVP_STATUS),
-        default=RSVP_STATUS.A,
+        default=RSVP_STATUS.MAYBE,
```
jace: This default should remain `awaiting`.

funnel/models/rsvp.py
```diff
@@ -40,23 +38,60 @@ class Rsvp(TimestampMixin, db.Model):
...
         nullable=False,
     )
     state = StateManager('_state', RSVP_STATUS, doc="RSVP answer")
 
+    @with_roles(call={'auth'})
```
jace: Wrong role. Use `owner`. Using `auth` allows anyone who is logged in to change anyone else's response status.

funnel/models/rsvp.py
```diff
@@ -40,23 +38,60 @@ class Rsvp(TimestampMixin, db.Model):
...
+    )
+    def rsvp_no(self):
+        pass
+
+    @with_roles(call={'auth'})
+    @state.transition(
+        None,
+        state.MAYBE,
+        title=__("Maybe"),
+        message=__("Your RSVP has been saved"),
```
jace: Change these messages to `"Your response has been saved"`

funnel/models/rsvp.py
```diff
@@ -40,23 +38,60 @@ class Rsvp(TimestampMixin, db.Model):
...
         if user:
             result = cls.query.get((project.id, user.id))
             if not result and create:
                 result = cls(project=project, user=user)
-                db.session.add(result)
+                # we need to show the transition form on the project page but
+                # every person who opens the project page may not want to RSVP.
+                # so this flag lets us create an Rsvp object without adding it to the db.
+                if session_add:
+                    db.session.add(result)
```
jace: This is not safe. SQLAlchemy has session cascades, so if `User` or `Project` has a backref to `Rsvp`, then it will be automatically added to the session. Instead, pass `create=False` if you don't want to create a database object. Remove this code.

funnel/models/rsvp.py
```diff
@@ -40,23 +38,60 @@ class Rsvp(TimestampMixin, db.Model):
...
             return result
 
 
-def _project_rsvp_for(self, user, create=False):
-    return Rsvp.get_for(self, user, create)
+def _project_rsvp_for(self, user, create=False, session_add=False):
+    return Rsvp.get_for(self, user, create, session_add)
```
jace: Remove.

funnel/templates/project.html.jinja2
```diff
@@ -83,8 +83,18 @@
             <div id="boxoffice-widget"><p class="mui--text-body1">Loading...</p></div>
           {% elif project.allow_rsvp -%}
             <h2 class="mui--text-display1 mui--text-left project-section__headline">{% trans %}Attending this event?{% endtrans %}</h2>
+            {% if current_rsvp is not none %}
+            <p>Your current RSVP is: {{ current_rsvp.state.label.title }}
```
jace: Don't use the word "RSVP" in UI. Use `Your current status:`

funnel/views/project.py
```diff
@@ -123,12 +122,16 @@ class ProjectView(ProjectViewMixin, DraftViewMixin, UrlForView, ModelView):
     @render_with('project.html.jinja2')
     @requires_permission('view')
     def view(self):
-        rsvp_form = RsvpForm(obj=self.obj.rsvp_for(g.user))
+        current_rsvp = self.obj.rsvp_for(current_auth.user)
         transition_form = ProjectTransitionForm(obj=self.obj)
         schedule_transition_form = ProjectScheduleTransitionForm(obj=self.obj)
         project_save_form = SavedProjectForm()
+        rsvp_form = RsvpTransitionForm(
+            obj=self.obj.rsvp_for(current_auth.user, create=True, session_add=False)
```
jace: Don't use transitions to get the initial set of choices. Specify them elsewhere. See if you can extract them from a classmethod instead of an instance.

funnel/views/project.py
```diff
@@ -332,21 +335,19 @@ def schedule_transition(self):
         return redirect(self.obj.url_for())
 
     @route('rsvp', methods=['POST'])
-    @render_with('rsvp.html.jinja2')
     @lastuser.requires_login
-    @requires_permission('view')
-    def rsvp(self):
-        form = RsvpForm()
+    def rsvp_transition(self):
+        rsvp = Rsvp.get_for(self.obj, current_auth.user, create=True, session_add=True)
```
jace: Don't create before validating the form. Use `rsvp = Rsvp.get_for(self.obj, current_auth.user, create=False)`

funnel/views/project.py
```diff
@@ -332,21 +335,19 @@ def schedule_transition(self):
...
+        form = RsvpTransitionForm(obj=rsvp)
         if form.validate_on_submit():
-            rsvp = Rsvp.get_for(self.obj, g.user, create=True)
-            form.populate_obj(rsvp)
```
jace: Insert:
```python
if not rsvp:
    rsvp = Rsvp.get_for(self.obj, current_auth.user, create=True)
```

funnel/views/project.py
```diff
@@ -332,21 +335,19 @@ def schedule_transition(self):
...
+            transition()
             db.session.commit()
-            if request.is_xhr:
-                return {'project': self.obj, 'rsvp': rsvp, 'rsvp_form': form}
-            else:
-                return redirect(self.obj.url_for(), code=303)
+            flash(transition.data['message'], 'success')
         else:
-            abort(400)
+            flash(_("Invalid RSVP option for this project"), 'error')
```
jace: `"This response is not valid"`

funnel/models/rsvp.py
jace: Also specify a grant for `owner` in the relationship.

jace:
Here's a fix:

1. Add an "Unknown" state to the RSVP enum, with value `None`
2. Make this the default, but also set `nullable=False` in the column
3. Provide a `default_transitions` classmethod that will create an empty object, extract the transitions, and then delete the empty object (`db.session.delete`)
4. Use this from the form's `set_queries` when no object is provided

funnel/models/rsvp.py
iambibhas: > Also specify a grant for owner in the relationship.

What do you mean by that?

funnel/forms/project.py
```diff
@@ -228,11 +228,22 @@ class SavedProjectForm(forms.Form):
...
+class RsvpTransitionForm(forms.Form):
+    transition = forms.SelectField(
+        __("Status"), validators=[forms.validators.DataRequired()]
     )
 
+    def set_queries(self):
+        # usually you need to use an instance's state.transitions to find
+        # all the valid transitions for the current state of the instance.
+        # But for RSVP, we're showing all the options all the time, so this
+        # can is valid. We're also doing this because we want to load the
```
jace: "call" is valid.

funnel/templates/project.html.jinja2
```diff
@@ -83,8 +83,18 @@
             <div id="boxoffice-widget"><p class="mui--text-body1">Loading...</p></div>
           {% elif project.allow_rsvp -%}
             <h2 class="mui--text-display1 mui--text-left project-section__headline">{% trans %}Attending this event?{% endtrans %}</h2>
+            {% if current_rsvp is not none %}
+            <p>Your current status: {{ current_rsvp.state.label.title }}
```
jace: `{% trans %}` needed. You'll have to incorporate the title as a variable in the trans tag so the whole paragraph is sent for translation.

jace: Missing `</p>`

funnel/views/project.py
```diff
@@ -332,21 +333,19 @@ def schedule_transition(self):
...
             db.session.commit()
-            if request.is_xhr:
-                return {'project': self.obj, 'rsvp': rsvp, 'rsvp_form': form}
-            else:
-                return redirect(self.obj.url_for(), code=303)
+            flash(transition.data['message'], 'success')
         else:
-            abort(400)
+            flash(_("This response is not valid"), 'error')
+            abort(403)
```
jace: Don't abort. You have handled the form submission, so let the redirect happen.

PR #508, opened 2019-10-01
Fixed permissions with ticket client view and primary venue null issue

iambibhas:


funnel/models/event.py
```diff
@@ -3,6 +3,8 @@
 import base64
 import os
 
+from coaster.sqlalchemy import PermissionMixin
+
```
jace: Not required.

funnel/models/event.py
```diff
@@ -334,6 +336,12 @@ def import_from_list(self, ticket_list):
                 # Ensure that the new or updated participant has access to events
                 ticket.participant.add_events(ticket_type.events)
 
+    def permissions(self, user, inherited=None):
+        perms = super(TicketClient, self).permissions(user, inherited)
+        if self.project is not None and isinstance(self.project, PermissionMixin):
```
jace: You don't need the `isinstance` test because you already know what a project is.

PR #513, opened 2019-11-04
Use baseframe-mui version 0.2.0 and add select2 as separate asset

vidya-ram:
Baseframe dependency: https://github.com/hasgeek/baseframe/pull/242

### DIFF
--- a/funnel/_version.py
+++ b/funnel/_version.py
@@ -2,7 +2,7 @@
 
 __all__ = ['__version__', '__version_info__']
 
-__version__ = '0.2.0'
+__version__ = '0.1.0'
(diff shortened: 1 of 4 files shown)


funnel/_version.py
```diff
@@ -2,7 +2,7 @@
 
 __all__ = ['__version__', '__version_info__']
 
-__version__ = '0.1.0'
+__version__ = '0.2.0'
```
jace: This appears to have no connection with the other changes.

vidya-ram:
https://github.com/hasgeek/baseframe/tree/select2

PR #522, opened 2019-12-02
Switch to font awesome 5 icon fonts

vidya-ram:


vidya-ram:
Fixed as part of #523

PR #523, opened 2019-12-20
Update Homepage UI

vidya-ram:
1. Remove material icons and update all icons to font-awesome 5(regular)
2. Update homepage UI
3. Add back the `Past Projects` section to homepage
4. For text truncation, switch to `jquery.truncate8.js`
5. Add `livestream_urls` to project

Dependency https://github.com/hasgeek/baseframe/pull/250


funnel/models/project.py
```diff
@@ -540,11 +556,11 @@ def calendar_weeks(self):
                 and (schedule_start_week - current_week) <= 2
             ):
                 if (schedule_start_week - current_week) == 2:
-                    session_dates.insert(0, (now + timedelta(days=7), 0))
-                session_dates.insert(0, (now, 0))
+                    session_dates.insert(0, (now + timedelta(days=7), None, None, 0))
+                session_dates.insert(0, (now, None, None, 0))
```
jace: Could you document what `(now, None, None, 0)` is? A comment will help.

migrations/versions/20c10335b553_add_project_livestream_urls.py
```diff
@@ -0,0 +1,32 @@
...
+import sqlalchemy as sa  # NOQA
+
+
+def upgrade():
+    op.add_column(
+        'project',
+        sa.Column(
+            'livestream_urls',
+            sa.ARRAY(sa.UnicodeText(), dimensions=1),
+            server_default='{}',
```
jace: Is this the correct server default?

jace: Your model doesn't have a `server_default`, so the migration should not have one either. If you want to insert default values when adding the column, then alter it immediately after to remove the server default.

PR #528, opened 2020-01-06
Withdrawn patch

qurbat:


funnel/views/project.py
```diff
@@ -157,18 +157,22 @@ def json(self):
...
-                    self.obj
-                ),  # TODO: Remove when the native app switches over
-                'venues': [venue_data(venue) for venue in self.obj.venues],
-                'rooms': [room_data(room) for room in self.obj.rooms],
-                'proposals': [proposal_data(proposal) for proposal in proposals],
-                'schedule': schedule_data(self.obj),
-            }
-        )
+        # return JSON if project is published
+        if self.obj.state.value == 1:
```
jace: 1. Don't use raw values. They are hard to track down when upgrading the list of states.
2. This sort of gate should be in a view decorator, not inside the view. In this case, there is a `@requires_permission('view')`. That permission should not be granted if the user is not supposed to see this. Code for that is in the `permissions` method on the model.

funnel/views/project.py
```diff
@@ -184,16 +188,20 @@ def csv(self):
...
-            content_type='text/csv',
-            headers=[
-                (
-                    'Content-Disposition',
-                    'attachment;filename="{project}.csv"'.format(project=self.obj.name),
-                )
-            ],
-        )
+        # return CSV if project is published
+        if self.obj.state.value == 1:
```
jace: As above.

jace:
Don't change the title when you withdraw a patch. Titles are important for retaining context of what this is for and why it was not accepted.

PR #530, opened 2020-01-07
Update project.py model

qurbat:
Makes it so only members of the Admin, Review and Check-in teams will be able to access a project while it is in the withdrawn state

### DIFF
--- a/funnel/models/project.py
+++ b/funnel/models/project.py
@@ -704,6 +704,13 @@ class Project(UuidMixin, BaseScopedNameMixin, db.Model):
     def permissions(self, user, inherited=None):
         perms = super(Project, self).permissions(user, inherited)
         perms.add('view')
+        if self.state.WITHDRAWN:
+            if user not in self.admin_team.users:
+                abort(410)
+            elif user not in self.review_team.users:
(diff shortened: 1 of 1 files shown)


qurbat:
Issue #524

funnel/models/project.py
```diff
@@ -704,6 +704,13 @@ def location_geonameid(self):
     def permissions(self, user, inherited=None):
         perms = super(Project, self).permissions(user, inherited)
         perms.add('view')
+        if self.state.WITHDRAWN:
+            if user not in self.admin_team.users:
+                abort(410)
+            elif user not in self.review_team.users:
+                abort(410)
+            elif user not in self.checkin_team.users:
+                abort(410)
```
jace: `abort` is a view action. You should not be performing view actions in a model, which is only responsible for data state and data integrity checks. In this case, the `if` condition should not offer the `view` permission when the project is in withdrawn state.

Separately, since this changes data access, it needs a ticket presenting a case for (a) why a project should not be accessible in the withdrawn state, and (b) what is broken or not broken by such a change. The ticket and argument should precede this pull request.

PR #531, opened 2020-01-07
Inline title for home page cards

jace:


funnel/models/project.py
```diff
@@ -268,6 +269,15 @@ def __repr__(self):
             self.title,
         )
 
+    @property
+    def title_inline(self):
+        """Suffix a colon if the title does not end in ASCII sentence punctuation"""
+        title = self.title.strip()
+        if title and self.tagline.strip():
```
jace: Both instances of `.strip()` are only required if we're not validating title and tagline properly to strip whitespace during data entry in the form. That needs to be checked, and if done correctly, these cases of `.strip()` should be removed.

jace: Will also require checking the database for trailing whitespace as there may be old data predating the form validation.

jace: @qurbat you should investigate this.

jace:
This will break i18n and should be considered a temporary measure. Colons aren't universal punctuation. Example: https://softwareengineering.stackexchange.com/a/234856/23112

jace:
Even spaces _after_ colons are subject to language rules. http://docs.translatehouse.org/projects/localization-guide/en/latest/guide/translation/punctuation.html

Issue #532, opened 2020-01-08
Introduce delete functionality to event set-up page


qurbat:
As of now, users cannot remove objects they have added to the page located at **{profile}/{project}/admin.** This results in a situation where users are unable to perform cleanup on an event that has been setup improperly.

A _delete_ function should be added here, along with a _reset_ function that can be used to revert an event to a "fresh" state.

qurbat:
Re-opened to add delete functionality for **ticket_client**

PR #537, opened 2020-01-09
Add delete function for /event and /ticket_type

qurbat:
Introduces the ability to delete (using `render_delete_sqla`) the following endpoints:

1.  `{profile}/{project}/event/{name}/delete`
2. `{profile}/{project}/ticket_type/{name}/delete`

See: #532


qurbat:
@iambibhas This does not introduces UI changes as of yet. If it looks good, I will proceed with adding buttons for both **event** and **ticket_type**.

funnel/views/event.py
```diff
@@ -189,6 +189,22 @@ def edit(self):
             return redirect(self.obj.project.url_for('admin'), code=303)
         return render_form(form=form, title=_(u"Edit event"), submit=_(u"Save changes"))
 
+    @route('delete', methods=['GET', 'POST'])
+    @requires_permission('edit_event')
```
jace: Should be `delete_event`

funnel/views/event.py
```diff
@@ -265,6 +281,22 @@ def edit(self):
             form=form, title=_(u"Edit ticket type"), submit=_(u"Save changes")
         )
 
+    @route('delete', methods=['GET', 'POST'])
+    @requires_permission('edit_event')
```
jace: Same.

PR #540, opened 2020-01-13
Add form validator for icon_emoji

qurbat:
1. Add emoji validation function to baseframe https://github.com/hasgeek/baseframe/pull/253
2. Add emoji form validator to funnel https://github.com/hasgeek/funnel/pull/540


qurbat:
This seems to properly validate supplied emojis, but I still haven't figured out how to make an error message pop-up when the validation fails.

PR #544, opened 2020-01-27
Update the UI of calendar widget

vidya-ram:
1. Updated calendar widget UI
  a) Show only the first event week on the project cards in the homepage
  b) Show full calendar(two weeks prior and all event days) in the project page
  c) Show the first event's start and end time as tooltip
3. Add default banner in the project page header if it has no banner image
4. Change saved bookmark icon filled colour.
5. Link the venue on the header in the project page

Project page:
Screen Shot 2020-01-30 at 7 33 45 PM

Card in home page
Screen Shot 2020-01-30 at 7 33 59 PM


runfrontendtests.sh
```diff
@@ -6,5 +6,5 @@ python -m tests.e2e.frontend_tests_initdb
 python runcypressserver.py &
 SERVER_PID=$!
 cd funnel/assets
-npx cypress run --browser chrome --record --key $RECORD_KEY
+npx cypress run --browser chrome
```
vidya-ram: Cypress limit for this month has exceeded so removed the record key to run tests. Will add  it back before merging.

PR #547, opened 2020-01-30
Moving to Python3

iambibhas:
Related to https://github.com/hasgeek/hgapp/issues/39


funnel/models/project.py
```diff
@@ -324,7 +324,7 @@ def datelocation(self):
                 end_date=schedule_end_at_date.strftime("%d %b"),
                 year=schedule_end_at_date.year,
             )
-        return u', '.join(filter(None, [daterange, self.location]))
+        return ', '.join([_f for _f in [daterange, self.location] if _f])
```
jace: Is this an accurate replacement?

funnel/models/proposal.py
```diff
@@ -532,7 +532,7 @@ def url_id_name(self):
         likely never be called directly on an instance. It exists for the SQL
         comparator that will be called to load the instance.
         """
-        return unicode(self.url_id)
+        return self.url_id
```
jace: `return str(self.url_id)` as it's an integer.

funnel/views/proposal.py
```diff
@@ -188,7 +188,7 @@ def view(self):
         delcommentform = DeleteCommentForm()
 
         links = [
-            Markup(linkify(unicode(escape(l))))
+            Markup(linkify(escape(l)))
```
jace: Needs to be `str()` as `escape` returns a HTML string.

funnel/views/schedule.py
```diff
@@ -34,53 +34,63 @@
...
+        + list(
+            dict(
+                {'modal_url': session.url_for(with_modal_url)} if with_modal_url else {}
+            ).items()
+        )
+        + list(
+            dict(
+                {'delete_url': session.url_for('delete')} if with_delete_url else {}
+            ).items()
+        )
```
jace: This ugly block should be replaced with a single dict and two update operations.

funnel/models/project.py
iambibhas: Seems so.

PR #578, opened 2020-04-06
Hindi i18n

qurbat:
Hindi translations for strings in Funnel


babel.cfg
```diff
@@ -0,0 +1,3 @@
+[python: **.py]
+[jinja2: **/templates/**.html.jinja2]
```
jace, line 2: Should be `**.jinja2`. They could be anywhere.

PR #584, opened 2020-04-13
Update init db for frontend tests

vidya-ram:


funnel/assets/cypress/fixtures/proposal.json
```diff
@@ -9,5 +9,6 @@
   "location": "Bangalore",
   "room": "Taj M G Road – Banquet hall",
   "venue_room": "in Banquet hall Taj M G Road",
-  "time": "9:05"
+  "time": "9:05",
+  "email": "cypress-user@gmail.com"
```
iambibhas, line 13: is this valid email format?

tests/e2e/frontend_tests_initdb.py
```diff
@@ -1,11 +1,29 @@
...
-from funnel.models import db
+from funnel.models import Organization, Profile, User, db
 
 
 def init_models():
-    with app.app_context():
+    with app.test_request_context():
         db.create_all()
+        profile_admin = User(username='admin-cypress', fullname='admin-cypress')
+        profile_admin._set_password('cypress129')
```
iambibhas, line 10: call these `user_admin` maybe. `Profile` is a different model.

funnel/assets/cypress/fixtures/proposal.json
vidya-ram: It's valid email but not a valid gmail format. I have changed the domain to random one.

PR #587, opened 2020-04-14
Add videos to session

vidya-ram:


funnel/models/video.py
```diff
@@ -15,6 +15,84 @@ class VideoException(Exception):
...
+    video_source = 'raw'
+    video_id = video_url
+
+    parsed = urllib.parse.urlparse(video_url)
+    if parsed.netloc is None:
+        raise ValueError("Invalid video URL")
+
+    if parsed.netloc in ['youtube.com', 'www.youtube.com', 'm.youtube.com']:
+        if parsed.path == '/watch':
+            queries = urllib.parse.parse_qs(parsed.query)
```
jace, line 28: Nitpick: this is conventionally `args` rather than `queries`. Not a serious issue though.

funnel/views/proposal.py
```diff
@@ -42,7 +42,7 @@
     'speaker',
     'email',
     'slides',
-    'preview_video',
+    'video_url',
```
jace, line 45: `video` right?

migrations/versions/41b3af7e4449_migrating_preview_video_to_video_field.py
```diff
@@ -0,0 +1,129 @@
...
+proposal = table(
+    'proposal',
+    column('id', sa.Integer()),
+    column('preview_video', sa.UnicodeText()),
+    column('video_id', sa.UnicodeText()),
+    column('video_source', sa.UnicodeText()),
+)
+
+
+troublesome_filename = 'preview-video-troublesone.csv'
```
jace, line 36: `troublesone` -> `troublesome`

migrations/versions/41b3af7e4449_migrating_preview_video_to_video_field.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def upgrade():
+    conn = op.get_bind()
+
+    proposals = conn.execute(
+        proposal.select().where(proposal.c.preview_video != None)
+    )  # NOQA
```
jace, line 44: Use `isnot`

migrations/versions/41b3af7e4449_migrating_preview_video_to_video_field.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+    )
+
+    op.drop_column('proposal', 'preview_video')
+
+
+def downgrade():
+    op.add_column(
+        'proposal', sa.Column('preview_video', sa.UnicodeText(), nullable=True)
+    )
```
jace, line 106: This needs to be reverse-populated from the video_id/video_source columns. Also, where is the migration to add them?

funnel/views/proposal.py
iambibhas: This is used when generating CSV. Do we want to put a dict in there? I thought video_url might be better for it.

migrations/versions/41b3af7e4449_migrating_preview_video_to_video_field.py
iambibhas: Added that

PR #590, opened 2020-04-16
Fix minor issues

vidya-ram:
1. Show timezone for event duration on calendar widget
2. Change proposal submission end note to use warning icon
3. Use tomorrow and Day after text on calendar widget
4. Fix function getVideoTypeAndId to work for non youtube and vimeo urls
5. Fix mui-btn--nostyle class
6. Remove unused css (fa style, material icon styles)

Screen Shot 2020-04-16 at 2 39 48 PM
Screen Shot 2020-04-16 at 4 13 04 PM

### DIFF
--- a/funnel/assets/js/util.js
+++ b/funnel/assets/js/util.js
@@ -159,19 +159,17 @@ export const Utils = {
       let eventDay = Date.parse(eventDate);
       // Find the difference between event and today's date in UTC
       let counting = Math.round((eventDay - today) / singleDay);
+      let dayText = ['Today', 'Tomorrow', 'Day after'];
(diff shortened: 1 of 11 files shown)


funnel/assets/js/util.js
```diff
@@ -159,19 +159,17 @@ export const Utils = {
       let eventDay = Date.parse(eventDate);
       // Find the difference between event and today's date in UTC
       let counting = Math.round((eventDay - today) / singleDay);
+      let dayText = ['Today', 'Tomorrow', 'Day after'];
```
jace, line 162: These need to pass through gettext, so ideally from backend (for now)

PR #608, opened 2020-04-28
Show the next upcoming event week in the project cards on the home page

vidya-ram:


funnel/models/project.py
```diff
@@ -592,15 +592,21 @@ def calendar_weeks(self, leading_weeks=True):
...
+                # Set upcoming for weeks with event dates in the future
+                weeks[weekobj.week]['upcoming'] = weekobj >= current_week
                 weeks[weekobj.week]['year'] = weekobj.year
                 # Order is important, and we need dict to count easily
                 weeks[weekobj.week]['dates'] = OrderedDict()
             for wdate in weekobj.days():
                 weeks[weekobj.week]['dates'].setdefault(wdate, 0)
                 if project_date.date() == wdate:
+                    # If the event is over don't set upcoming for current week
+                    if wdate.isoformat() < today:
```
jace, line 608: This is comparing strings. Compare dates directly.

vidya-ram: Changed to compare dates

PR #623, opened 2020-05-05
basic feature for project RSVP

iambibhas:


funnel/models/project.py
```diff
@@ -927,6 +927,30 @@ def is_saved_by(self, user):
 )
 
 
+@Project.features('has_rsvp')
+def project_has_rsvp(obj):
+    return (
+        not obj.schedule_state.PAST
+        and obj.schedule_state.PUBLISHED
```
jace, line 934: Put `PUBLISHED` first as it's a cheaper condition to evaluate.

funnel/models/project.py
```diff
@@ -927,6 +927,30 @@ def is_saved_by(self, user):
...
+        and (
+            obj.boxoffice_data is None
+            or 'item_collection_id' not in obj.boxoffice_data
+            or not obj.boxoffice_data['item_collection_id']
+        )
```
jace, line 939: All this is cheaper than `PAST` which requires a column property and may involve a db roundtrip. Maybe put the `PAST` condition last.

funnel/models/project.py
```diff
@@ -927,6 +927,30 @@ def is_saved_by(self, user):
...
+
+
+@Project.features('has_tickets')
+def project_has_tickets(obj):
+    return (
+        not obj.schedule_state.PAST
+        and obj.schedule_state.PUBLISHED
+        and obj.boxoffice_data is not None
+        and 'item_collection_id' in obj.boxoffice_data
+        and obj.boxoffice_data['item_collection_id']
```
jace, line 950: Likewise.

funnel/templates/proposal.html.jinja2
```diff
@@ -23,17 +23,19 @@
...
-      {% if project.boxoffice_data.buy_btn_text -%}
-        {{ project.boxoffice_data.buy_btn_text }}
-      {% else %}
-        {% trans %}Tickets{% endtrans %}
-      {% endif %}
-    {%- else -%}
-      {% trans %}Attend{% endtrans %}
-    {%- endif %}
-  </a>
+  {% if project.features.has_tickets() or project.features.has_rsvp() %}
```
jace, line 26: Make this a new feature named `has_tickets_or_rsvp`. It can contain:

```python
return obj.features.has_tickets() or obj.features.has_rsvp()
```

jace: Rationale: we want to aim for the simplicity of Mustache templates, where the `if` condition in a template has only `if single_boolean_condition`. Moving the feature tests into a centralised module will make it easier to monitor the growth in complexity.

PR #626, opened 2020-05-06
need email text for send_mail

iambibhas:


funnel/views/membership.py
```diff
@@ -96,7 +96,7 @@ def new_member(self):
 
                     send_mail_async.queue(
                         sender=None,
-                        to=new_membership.user.email,
+                        to=new_membership.user.email.email,
```
jace, line 99: The convention we're using is `str(user.email)`. It's documented in the UserEmail model.

PR #632, opened 2020-05-11
Add notice in the UI for non public project pages

vidya-ram:
Screen Shot 2020-05-11 at 2 30 08 PM
<hr>
Screen Shot 2020-05-11 at 2 30 04 PM

### DIFF
--- a/funnel/models/project.py
+++ b/funnel/models/project.py
@@ -954,8 +954,8 @@ def project_has_tickets_or_rsvp(obj):
     return obj.features.has_tickets() or obj.features.has_rsvp()
 
 
-@Project.features('schedule_publish_alert')
-def project_has_sessions(obj):
+@Project.features('schedule_no_sessions')
+def project_has_no_sessions(obj):
     return obj.schedule_state.PUBLISHED and not obj.schedule_start_at
 
 
--- a/funnel/models/project.py
+++ b/funnel/models/project.py
@@ -954,6 +954,11 @@ def project_has_tickets_or_rsvp(obj):
     return obj.features.has_tickets() or obj.features.has_rsvp()
 
 
+@Project.features('schedule_publish_alert')
+def project_has_sessions(obj):
+    return obj.schedule_state.PUBLISHED and not obj.schedule_start_at
+
+
 class ProjectRedirect(TimestampMixin, db.Model):
     __tablename__ = "project_redirect"
 
--- a/funnel/templates/macros.html.jinja2
+++ b/funnel/templates/macros.html.jinja2
@@ -99,7 +99,7 @@
       {% if project.view_for('edit').is_available() and project.schedule_state.PUBLISHED and not project.schedule_start_at %}
         <div class="grid__col-sm-12">
           <div class="alert alert--warning">
-            <p class="alert__text">Please add session to the schedule for the project to be listed on the profile and homepage.</p>
+            <p class="alert__text">This project will not be listed as it has no sessions in the schedule</p>
           </div>
         </div>
       {% endif %}
--- a/funnel/templates/macros.html.jinja2
+++ b/funnel/templates/macros.html.jinja2
@@ -96,7 +96,7 @@
 {% macro project_header(project, project_save_form, class='', current_page='project') %}
   <div class="mui-container project-banner {% if class %}{{ class }}{% endif %}">
     <div class="grid">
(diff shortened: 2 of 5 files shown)


funnel/templates/macros.html.jinja2
```diff
@@ -96,6 +96,13 @@
 {% macro project_header(project, project_save_form, class='', current_page='project') %}
   <div class="mui-container project-banner {% if class %}{{ class }}{% endif %}">
     <div class="grid">
+      {% if project.view_for('edit').is_available() and project.schedule_state.PUBLISHED and not project.schedule_start_at %}
```
jace, line 99: This should be a feature. Possible name: `schedule_publish_alert`

funnel/templates/macros.html.jinja2
```diff
@@ -96,6 +96,13 @@
...
+        <div class="grid__col-sm-12">
+          <div class="alert alert--warning">
+            <p class="alert__text">Please add session to the schedule for the project to be listed on the profile and homepage.</p>
```
jace, line 102: "This project will not be listed as it has no sessions in the schedule"

funnel/templates/schedule_edit.html.jinja2
```diff
@@ -12,6 +12,11 @@
 {% endblock %}
 
 {% block top_title %}
+  {% if project.schedule_state.PUBLISHED and not project.schedule_start_at %}
+    <div class="alert alert--warning">
+      <p class="alert__text">Please add session to the schedule for the project to be listed on the profile and homepage.</p>
+    </div>
+  {% endif %}
```
jace, line 19: This warning may not work as the page does not reload when individual sessions are added or removed.

funnel/templates/macros.html.jinja2
vidya-ram: Added `schedule_publish_alert` to project features

funnel/templates/schedule_edit.html.jinja2
vidya-ram: Do I add this check while editing session ?

vidya-ram: Screen Shot 2020-05-11 at 3 21 54 PM

jace: This works though it's easy to miss.

funnel/templates/macros.html.jinja2
jace: It's not being used here?

funnel/models/project.py
```diff
@@ -954,6 +954,11 @@ def project_has_tickets_or_rsvp(obj):
     return obj.features.has_tickets() or obj.features.has_rsvp()
 
 
+@Project.features('schedule_publish_alert')
+def project_has_sessions(obj):
```
jace, line 958: If we use the BEM-inspired convention, this will be:
```python
@Project.features('schedule__no_sessions')
def project_has_no_sessions(obj):
```

funnel/views/session.py
```diff
@@ -198,6 +198,12 @@ def delete(self):
         else:
             self.obj.make_unscheduled()
         db.session.commit()
+        if self.obj.project.features.schedule_publish_alert():
+            return jsonify(
+                status=True,
+                modal_url=modal_url,
+                msg='This project will not be listed as it has no sessions in the schedule',
```
jace, line 205: `msg=_("This project will not be listed as it has no sessions in the schedule"),`

funnel/templates/schedule_edit.html.jinja2
```diff
@@ -13,7 +13,10 @@
 
 {% block top_title %}
   {% if project.features.schedule_publish_alert() %}
-    <div class="alert alert--warning">
+    <div class="alert alert--warning alert--dismissable" id="schedule-publish-alert">
+      <a class="alert__close" href="javascript:void(0);" aria-label="{% trans %}Close{% endtrans %}">
+        {{ faicon(icon='times', icon_size='subhead', baseline=true) }}
+      </a>
       <p class="alert__text">This project will not be listed as it has no sessions in the schedule</p>
     </div>
```
jace: Looks like this block should be in a macro as it's repeated markup for alerts.

vidya-ram: Used baseframe alertbox

PR #633, opened 2020-05-13
Restrict comments to participants of project only

iambibhas:


funnel/models/project.py
```diff
@@ -879,6 +879,14 @@ def roles_for(self, actor=None, anchors=()):
             if crew_membership is not None:
                 roles.update(crew_membership.offered_roles())
 
+            if (
+                self.rsvp_for(actor) is not None
+                or self.participants.filter_by(user=actor).one_or_none() is not None
```
jace, line 884: I believe you can do `actor in self.participants`

jace: This may be simpler using the `grants` mechanism. Make sure the `rsvp` relationship is defined first as it's more commonly used.

funnel/models/project.py
```diff
@@ -879,6 +879,14 @@ def roles_for(self, actor=None, anchors=()):
...
+            ):
+                roles.add('attendee')
+
+            if 'attendee' in roles or 'crew' in roles:
+                roles.add('participant')
```
jace, line 889: This will continue to work using the grants mechanism as "roles" here is a LazyRoleSet. It will look for existing role grants.

funnel/templates/project_comments.html.jinja2
```diff
@@ -46,7 +46,15 @@
               {{ commenttree(comments, project.commentset, project, csrf_form) }}
             </ul>
           {% endif %}
-          {{ commentformtemplate(project, commentform, delcommentform, current_auth) }}
+          {% if project.features.post_comment() %}
+            {{ commentformtemplate(project, commentform, delcommentform, current_auth) }}
+          {% else %}
+            {% if project.features.has_tickets() %}
```
jace, line 52: We've set a convention. These should be named `comments__new` (double underscore) and `tickets` respectively.

funnel/views/proposal.py
```diff
@@ -304,7 +309,7 @@ def transition(self):
             abort(403)
         return redirect(self.obj.url_for())
 
-    @route('next')
+    @route('next')  # NOQA
```
jace, line 312: Specify the exception code. `# NOQA: A003`

funnel/views/commentvote.py
```diff
@@ -91,7 +97,7 @@ def loader(self, commentset, profile=None):
 
     @route('new', methods=['POST'])
     @requires_login
-    @requires_permission('new_comment')
+    @requires_roles({'parent_participant'})
```
jace, line 100: Unsure about 'parent'. Why not just 'participant'?

PR #639, opened 2020-05-14
Site membership

iambibhas:


funnel/models/site_membership.py
```diff
@@ -0,0 +1,58 @@
...
+            roles.add('comment_moderator')
+        if self.is_user_moderator:
+            roles.add('user_moderator')
+        roles.add('site_admin')
+        return roles
+
+    def roles_for(self, actor=None, anchors=()):
+        # FIXME: is this method required?
+        roles = super(SiteMembership, self).roles_for(actor, anchors)
+        return roles
```
jace, line 58: Not anymore with `grants_via`

jace: Also not required if it adds nothing.

funnel/models/user.py
```diff
@@ -188,6 +188,28 @@ def name(cls):  # NOQA: N805
...
+    @property
+    def is_user_moderator(self):
+        from .site_membership import SiteMembership
+
+        return (
+            SiteMembership.query.filter_by(
+                user=self, is_user_moderator=True, is_active=True
+            ).one_or_none()
+            is not None
+        )
```
jace, line 211: We can stick these in from site_membership.py. Avoids import inside the function.

funnel/views/account.py
```diff
@@ -91,6 +93,46 @@ def account(self):
...
 
+    @route('siteadmin/comments', endpoint='siteadmin_comments', methods=['GET', 'POST'])
+    @requires_login
+    @render_with('siteadmin_comments.html.jinja2')
+    def siteadmin_comments(self):
+        if not (
+            current_auth.user.is_comment_moderator
+            or current_auth.user.is_user_moderator
+        ):
+            return redirect('/')
```
jace, line 104: abort(403)

jace: There's a hack for this, actually. Use a ModelView with `model = User` and set `loader()` to simply return `current_auth.user`. That way you can use `@requires_roles`

funnel/models/membership.py
```diff
@@ -137,7 +147,9 @@ def replace(self, actor, **roles):
             raise AttributeError("Unknown role")
         self.revoked_at = db.func.utcnow()
         self.revoked_by = actor
-        new = type(self)(user=self.user, parent=self.parent, granted_by=self.granted_by)
+        new = type(self)(
+            user=self.user, parent_id=self.parent_id, granted_by=self.granted_by
+        )
```
jace, line 152: Why the switch to `parent_id`?

funnel/templates/siteadmin_layout.html.jinja2
```diff
@@ -0,0 +1,54 @@
...
+        <div>
+          <p id="footer-twitter">
+            <a href="https://twitter.com/hasgeek" target="_blank">{% trans %}Follow @hasgeek{% endtrans %}</a>
+          </p>
+          <a href="https://www.facebook.com/HasGeek" target="_blank">{% trans %}Like HasGeek's page on Facebook{% endtrans %}</a>
+        </div>
+      </div>
+    </div>
+  </div>
+{% endblock %}
```
jace, line 51: You don't need this if you're extending layout.

funnel/views/account.py
jace: Bad hack as we're not operating on the User (unlike in /account views). Better to create a non-persistent `Site` object that implements `RoleMixin`.

funnel/views/account.py
```diff
@@ -91,6 +94,54 @@ def account(self):
...
+    @requires_login
+    @render_with('siteadmin_comments_delete.html.jinja2')
+    def siteadmin_comments_delete(self):
+        if not (
+            current_auth.user.is_comment_moderator
+            or current_auth.user.is_user_moderator
+        ):
+            return redirect('/')
+
+        Form()
```
jace, line 137: ?

funnel/models/membership.py
iambibhas: Because `parent_id` can be `None`. so `self.parent` might raise an error I thought. Havent tested that yet. could that be an issue?

funnel/views/account.py
iambibhas: This was because of flake8 and pre-commit. Temporary change just to make a commit.

funnel/models/site_membership.py
```diff
@@ -0,0 +1,75 @@
...
+        return tuple(args)
+
+    def offered_roles(self):
+        """Roles offered by this membership record"""
+        roles = set()
+        if self.is_comment_moderator:
+            roles.add('comment_moderator')
+        if self.is_user_moderator:
+            roles.add('user_moderator')
+        roles.add('site_admin')
```
iambibhas, line 52: Should we assign this role to user inside `User` model's `roles_for()`?

funnel/templates/siteadmin_comments.html.jinja2
```diff
@@ -0,0 +1,83 @@
...
+  action="{{ url_for('siteadmin_comments_delete') }}"
+  method="post"
+  class="form-inline"
+>
+  {{ comment_delete_form.hidden_tag() }}
+  <table class="mui-table mui-table--bordered mui-table--responsive">
+    <thead>
+      <tr>
+        <th>
+          <button id="btn-comments-selectall" class="mui-btn mui-btn--accent">
```
vidya-ram, line 19: This can be a checkbox

funnel/templates/siteadmin_layout.html.jinja2
```diff
@@ -0,0 +1,30 @@
...
+        </div>
+      </div>
+    </div> #}
+
+    <div class="grid__col-xs-12 grid__col-sm-12 grid__col-md-12 grid__col-lg-12">
+      {% block admincontentblock %}
+      {% endblock %}
+    </div>
+  </div>
+{%- endblock %}
```
vidya-ram, line 30: This page doesn't seem to have a route

iambibhas: It's a base template. All the other siteadmin templates will extend from it.

funnel/views/account.py
```diff
@@ -91,6 +95,62 @@ def account(self):
...
+            Comment.query.filter(
+                Comment.uuid_b58.in_(request.form.getlist('comment_id'))
+            ).delete(synchronize_session=False)
+            db.session.commit()
+            flash(
+                _("Comment(s) successfully deleted."), category='info',
+            )
+        else:
+            flash(
+                _("There was a problem deleting the comments. Please try again."),
```
jace, line 148: Convention is no trailing period, even if it looks a little odd here because the first sentence will have a period.

funnel/models/site_membership.py
```diff
@@ -0,0 +1,99 @@
...
+                    cls.is_site_editor.is_(True),
+                ),
+                name='site_membership_has_role',
+            )
+        )
+        return tuple(args)
+
+    def offered_roles(self):
+        """Roles offered by this membership record"""
+        roles = set()
```
jace, line 59: Since `site_admin` is always granted, maybe start with `roles={'site_admin'}`. Reduces one `add` call.

PR #641, opened 2020-05-15
Delete rooms field from session form if venue has no rooms

vidya-ram:


funnel/views/session.py
```diff
@@ -203,7 +210,8 @@ def delete(self):
                 status=True,
                 modal_url=modal_url,
                 msg=_(
-                    "This project will not be listed as it has no sessions in the schedule"
+                    "This project will not be listed as it has no sessions in the"
+                    " schedule"
```
jace, line 214: Our convention for wrapped lines is to put the space at the end of the previous line.

PR #645, opened 2020-05-15
Fix login related flash messages on account page

vidya-ram:
1. Use block `basecontent` from layout to include  block `messages`.
2. Tab scroll function call was included back that removed accidentally in a previous PR.
3. Fix share url and title. 
4. Change bookmark to solid icon

### DIFF
--- a/funnel/assets/js/app.js
+++ b/funnel/assets/js/app.js
@@ -14,6 +14,7 @@ $(() => {
   Utils.collapse();
   Utils.smoothScroll();
   Utils.navSearchForm();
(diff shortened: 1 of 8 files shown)


funnel/assets/js/app.js
```diff
@@ -14,6 +14,7 @@ $(() => {
   Utils.collapse();
   Utils.smoothScroll();
   Utils.navSearchForm();
+  Utils.scrollTabs();
```
vidya-ram, line 17: Tab scroll function call was included back that was removed accidentally in a previous PR.

PR #647, opened 2020-05-16
Restrict allowed schemes and domains for images

jace:
Requires https://github.com/hasgeek/baseframe/pull/296

### DIFF
--- a/funnel/forms/project.py
+++ b/funnel/forms/project.py
@@ -2,6 +2,7 @@
 
 import re
 
+from flask import current_app
 from wtforms.ext.sqlalchemy.fields import QuerySelectMultipleField
 from wtforms.widgets import CheckboxInput, ListWidget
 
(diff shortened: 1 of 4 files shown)


funnel/forms/project.py
```diff
@@ -72,7 +73,14 @@ class ProjectForm(forms.Form):
...
         validators=[
             forms.validators.Optional(),
-            forms.validators.ValidUrl(),
+            forms.validators.ValidUrl(
+                allowed_schemes=lambda: current_app.config.get(
+                    'IMAGE_URL_SCHEMES', ('https',)
+                ),
+                allowed_domains=lambda: current_app.config.get('IMAGE_URL_DOMAINS'),
+                message_schemes=__("A https:// URL is required"),
+                message_domains=__("Images must be hosted at images.hasgeek.com"),
```
iambibhas, line 82: This error is kind of wrong given that the domain is being fetched from config.

jace: The expectation is that config will vary between production and development. We're not supporting others with hosting.

PR #648, opened 2020-05-18
Codemirror fix

vidya-ram:
1. Remove uglify for codemirror since it minified version is having errors.
2. Styling for Blockquote.
2. Cancel button for reply
4. Two lines height for comment textarea box
5. Styling fixes for comment header

dependency https://github.com/hasgeek/baseframe/pull/297

### DIFF
--- a/funnel/templates/comments.html.jinja2
+++ b/funnel/templates/comments.html.jinja2
@@ -12,22 +12,24 @@
           </button>
           <button class="js-uncollapse mui-btn mui-btn--nostyle mui--hide" aria-expanded="false" aria-label="{% trans %}Collapse comment{% endtrans %}">
             {{ faicon(icon='caret-circle-down', icon_size='subhead') }}
-          </button>&nbsp;
-          {% if comment.state.DELETED -%}
-            <span class="commenter">[{% trans %}deleted{% endtrans %}]</span>
-          {%- else -%}
-            <span class="commenter"><strong>{{ comment.user.pickername }}</strong></span>
-            {%- if comment.user == document.user %}
-              <span class="badge">Proposer</span>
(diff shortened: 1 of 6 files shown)


funnel/templates/comments.html.jinja2
```diff
@@ -12,22 +12,24 @@
...
-            <span class="mui--text-caption mui--text-accent">{{ comment.created_at|age }}</span>
-            {%- if comment.edited_at %}
-              <span class="mui--text-caption mui--text-accent">({% trans %}edited{% endtrans %} {{ comment.edited_at|age }})</span>
+          </button>
+          <div class="comment--header__details">
+            {% if comment.state.DELETED -%}
+              <span class="commenter">[{% trans %}deleted{% endtrans %}]</span>
+            {%- else -%}
+              <span class="commenter"><strong>{{ comment.user.pickername }}</strong></span>
+              {%- if comment.user == document.user %}
```
iambibhas, line 21: Unrelated to this PR - `document.user` is only available if `document` is a Proposal. We're also using `commenttree()` in project comments page where `document` is a Commentset.

jace: Comment have a `badges` set that it populates based on specific roles being available on the parent.

funnel/templates/comments.html.jinja2
```diff
@@ -12,22 +12,24 @@
...
+                <span class="badge">Proposer</span>
+              {% endif %}
+              {%- if 'crew' in project.roles_for(comment.user) %}
+                <span class="badge">Crew</span>
+              {% endif %}
```
jace, line 26: Move to the model to remove it from the template.

PR #651, opened 2020-05-19
Update form nonce for bookmarking

vidya-ram:


funnel/views/project.py
```diff
@@ -510,7 +511,8 @@ def admin(self):
                     import_tickets.queue(ticket_client.id)
             flash(
                 _(
-                    "Importing tickets from vendors...Refresh the page in about 30 seconds..."
+                    "Importing tickets from vendors...Refresh the page in about 30"
+                    " seconds..."
```
jace, line 515: Spaces should be at the end of the line.

```python
                    "Importing tickets from vendors...Refresh the page in about 30 "
                    "seconds..."
```

PR #656, opened 2020-05-19
Marking comments as spam instead of deleting them + comment badge

iambibhas:


funnel/models/commentvote.py
```diff
@@ -230,6 +232,18 @@ def title(self):
         else:
             return _("{user} commented").format(user=self.user.pickername)
 
+    @property
+    def badge(self):
```
jace, line 236: Should be `badges` returning a set. User can be both `crew` and `proposer` for instance.

funnel/models/commentvote.py
```diff
@@ -230,6 +232,19 @@ def title(self):
...
+    def badges(self):
+        badges = {}
+        if self.commentset.project is not None:
+            if 'crew' in self.commentset.project.roles_for(self.user):
+                badges.add('Crew')
+        elif self.commentset.proposal is not None:
+            if self.commentset.proposal.user == self.user:
+                badges.add('Proposer')
+            elif 'crew' in self.commentset.proposal.project.roles_for(self.user):
+                badges.add('Crew')
```
jace: Needs `_("name")`

funnel/templates/comments.html.jinja2
```diff
@@ -16,13 +17,14 @@
...
-              {%- if comment.user == document.user %}
-                <span class="badge">Proposer</span>
-              {% endif %}
-              {%- if 'crew' in project.roles_for(comment.user) %}
-                <span class="badge">Crew</span>
+              {%- if comment.badges %}
+                {% for badge in comment.badges %}
+                  <span class="badge">{{ badge }}</span>
+                {% endfor %}
               {% endif %}
```
jace, line 28: Remove `if` condition and only keep the `for`. The property is being called twice.

PR #657, opened 2020-05-19
Fix pytest fixtures and disable broken integration tests

jace:


tests/unit/lastuser_core/test_model_user_User.py
```diff
@@ -148,7 +148,7 @@ def test_user_del_email(self):
         db.session.commit()
         result1 = mr_jones.emails
         assert isinstance(result1, list)
-        result1 == [mr_jones_secondary_email, mr_jones_spare_email]
+        set(result1) == {mr_jones_secondary_email, mr_jones_spare_email}
```
iambibhas, line 151: is this supposed to be an assert?

PR #661, opened 2020-05-20
Added pagination to comment admin view

iambibhas:


funnel/templates/siteadmin_comments.html.jinja2
```diff
@@ -1,9 +1,23 @@
 {% extends 'siteadmin_layout.html.jinja2' %}
 
 {% block admincontentblock %}
-{% if comment_search_form %}
-  {{ renderform(comment_search_form, 'comment-search', action=url_for('siteadmin_comments'), submit="Search", style='horiz') }}
-{% endif %}
+<form data-parsley-validate="true" id="comment-search-form" method="GET"
+  action="/account/siteadmin/comments" accept-charset="UTF-8"
```
jace, line 5: Use `{{ url_for }}`?

funnel/templates/siteadmin_comments.html.jinja2
```diff
@@ -65,6 +79,16 @@
     {{ rendersubmit([("spam-comments-bulk", "Mark spam", '')]) }}
   </div>
 </form>
+
+<div class="mui--text-center">
+  {% for page in pages[:10] %}
+    <a
+      class="btn-pagination-item mui-btn {% if page != current_page %}mui-btn--accent{% endif %}"
+      href="{{ url_for('siteadmin_comments') }}?query={{ query }}&page={{ page }}"
+      >{{ page }}</a>
```
jace, line 88: Queries have built-in batch support. See the search views for example of how to use.

funnel/views/account.py
```diff
@@ -104,19 +103,34 @@ def siteadmin_comments(self):
...
+        # // does floor integer division, we need ceiling
+        total_pages = (total_comments // result_per_page) + 1
+        # index of the first item in current page, for slicing
+        first_item = (current_page - 1) * result_per_page
+
         return {
-            'comments': comments,
-            'comment_search_form': comment_search_form,
+            'query': query,
+            'comments': comments[first_item : first_item + result_per_page],  # NOQA
```
jace, line 131: This can be pushed into PostgreSQL using query batching.

funnel/templates/siteadmin_comments.html.jinja2
iambibhas: oh right. copied this from rendered HTML. forgot to change this.

jace: Actually this thing makes me wish for a static build step. The URL is not dynamic. The only reason to use `url_for` is to make future refactoring possible. This should be a compilation step and not a run-time step.

PR #666, opened 2020-05-21
fix presenter role for comment access

iambibhas:


funnel/models/proposal.py
```diff
@@ -502,7 +502,9 @@ def roles_for(self, actor=None, anchors=()):
         else:
             roles.add('reader')
 
-        if {'project_participant', 'speaker', 'reviewer'}.intersection(roles):
+        if (
+            {'project_participant', 'presenter', 'reviewer'}.intersection(roles)
+        ) or self.owner == actor:  # remove this check after proposal membership is implemented
             roles.add('commenter')
```
jace, line 508: Got the sake of performance:

1. `self.owner == actor` should be first
2. The fastest way to check for any of `a` being present in `b` is `not a.isdisjoint(b)` ([source](https://stackoverflow.com/a/17735466/78903)) but this is a bit ugly. Maybe we need a generic `anyin` function that does the same. Or maybe we'll get used to the ugliness.

iambibhas: #3 is a bit unconventional.

PR #667, opened 2020-05-21
Only active users for rsvp list + RSVP CSV

iambibhas:


funnel/views/project.py
```diff
@@ -460,6 +459,55 @@ def rsvp_transition(self):
...
+
+        outfile.seek(0)
+        return Response(
+            outfile.getvalue(),
+            content_type='text/csv',
+            headers=[
+                (
+                    'Content-Disposition',
+                    'attachment;filename="{filename}.csv"'.format(
+                        filename='rsvp-{project}-{state}'.format(
```
jace, line 485: Filename prefix should be `participants-`

funnel/views/project.py
```diff
@@ -460,6 +459,55 @@ def rsvp_transition(self):
...
+                            project=make_name(self.obj.title), state=state
```
jace, line 486: `project=self.obj.name`

iambibhas: can't do this because project names are mostly years. nor verbose.

jace: `profile-project`?

iambibhas: That's what make_name on title does, right? Our do you want to avoid the function call?

PR #669, opened 2020-05-21
Moderator reports

iambibhas:
This PR Adds

- A moderator report model to hold all spam comment reports
- only a comment moderator can report and review reports right now
- Mass reporting of comments
- report resolution based on reports and reviews by multiple comment moderators


funnel/forms/account.py
```diff
@@ -343,8 +343,11 @@ def validate_verification_code(self, field):
...
-        __("Report type"), validators=[forms.validators.DataRequired()]
+        __("Report type"), coerce=int, validators=[forms.validators.DataRequired()]
     )
 
     def set_queries(self):
-        self.report_type.choices = list(MODERATOR_REPORT_TYPE.items())
+        self.report_type.choices = [
+            (idx, report_type.title)
+            for idx, report_type in MODERATOR_REPORT_TYPE.items()
+        ]
```
jace, line 353: The original idea of `NameTitle` was to do this instead, to avoid exposing integer codes to the UI:

```python
self.report_type.choices = [(v.name, v.title) for v in MODERATOR_REPORT_TYPE.values()]
```

However, we don't have an API to remap from name to value. We need to revisit this.

funnel/models/commentvote.py
```diff
@@ -277,7 +277,13 @@ def delete(self):
     @state.transition(None, state.SPAM)
     def mark_spam(self):
         """
-        Mark this comment as spam.
+        Mark this comment as spam
```
jace, line 280: RST guidelines for docstrings require the first line to be a single statement that ends in a period. Counter to our own convention of not using periods.

funnel/templates/siteadmin_review_comment.html.jinja2
```diff
@@ -1,12 +1,16 @@
...
 
+{% block top_title %}
+  <h1 class="mui--text-display1"><strong>Review comments</strong></h1>
+{% endblock %}
+
 {% block admincontentblock %}
 <div class="mui--text-center">
-  <pre>{{ report.comment.message }}</pre>
-  <form action="." method="post">
+  <p>{{ report.comment.message }}</p>
```
jace, line 9: Doesn't the Markdown rendering already include `<p>` tags?

funnel/templates/siteadmin_review_comment.html.jinja2
```diff
@@ -1,12 +1,16 @@
...
+  <form action="{{ url_for('siteadmin_review_comment', report=report.uuid_b58) }}" method="post">
     {{ report_form.hidden_tag() }}
-    {% for report_type in report_form.report_type.choices %}
-    <button name="report_type" value="{{ report_type.data['value'] }}" class="mui-btn mui-btn--small mui-btn--raised">{{ report_type.data['title'] }}</button>
+    {% for value, title in report_form.report_type.choices %}
+    <button name="report_type" value="{{ value }}" class="mui-btn mui-btn--small mui-btn--raised">{{ title }}</button>
     {% endfor %}
   </form>
```
jace, line 15: Form above the message will help with Fitts' Law. Buttons will be in a consistent position, reducing cursor movement on desktop and scrolling on mobile.

funnel/templates/siteadmin_review_comment.html.jinja2
```diff
@@ -0,0 +1,31 @@
...
+{% block admincontentblock %}
+<div class="mui--text-center">
+  <form action="{{ url_for('siteadmin_review_comment', report=report.uuid_b58) }}" method="post">
+    {{ report_form.hidden_tag() }}
+    {% for value, title in report_form.report_type.choices %}
+    <button name="report_type" value="{{ value }}" class="mui-btn mui-btn--small mui-btn--accent">{{ title }}</button>
+    {% endfor %}
+  </form>
+  <div>
+    <h3>Existing reports</h3>
```
jace, line 16: `trans` tags needed

funnel/views/account.py
```diff
@@ -163,9 +168,15 @@ def siteadmin_comments_spam(self):
                 Comment.uuid_b58.in_(request.form.getlist('comment_id'))
             )
             for comment in comments:
-                comment.mark_spam()
+                comment.report_spam(reported_by=current_auth.user)
```
jace, line 171: Convention is `actor` as parameter.

funnel/views/account.py
```diff
@@ -163,9 +168,15 @@ def siteadmin_comments_spam(self):
...
             db.session.commit()
-            flash(_("Comment(s) successfully marked as spam"), category='info')
+            flash(
+                _(
+                    "Comment(s) successfully reported as spam. "
+                    "They will be removed after somebody reviews them"
```
jace, line 176: Remove second line.

funnel/views/account.py
```diff
@@ -174,6 +185,106 @@ def siteadmin_comments_spam(self):
...
+        endpoint='siteadmin_review_comment',
+        methods=['GET', 'POST'],
+    )
+    @render_with('siteadmin_review_comment.html.jinja2')
+    @requires_login
+    def siteadmin_review_comment(self, report=None):
+        if not current_auth.user.is_comment_moderator:
+            return abort(403)
+
+        report = CommentModeratorReport.query.filter_by(uuid_b58=report).one_or_404()
```
jace, line 216: But `report` is default `None`.

funnel/models/moderation.py
```diff
@@ -0,0 +1,56 @@
...
+
+__all__ = ['MODERATOR_REPORT_TYPE', 'CommentModeratorReport']
+
+
+class MODERATOR_REPORT_TYPE(LabeledEnum):  # NOQA: N801
+    OK = (0, 'ok', __("OK"))
+    SPAM = (1, 'spam', __("Spam"))
+
+
+class CommentModeratorReport(UuidMixin, NoIdMixin, db.Model):
```
jace, line 14: Using a compound primary key and adding a unique UUID column looks like a design error. One of them is not required.

funnel/views/account.py
```diff
@@ -174,6 +185,106 @@ def siteadmin_comments_spam(self):
...
+        if report.reported_by == current_auth.user:
+            flash(_("You cannot review same comment twice"), 'error')
```
jace, line 218: This should amend the existing report.

Or: allow the same comment to be reviewed multiple times, as the moderator may be in error and may need to revise their decision, while the backend should keep earlier reports for archival.

migrations/versions/954f066e0e4b_moderator_report_model.py
```diff
@@ -0,0 +1,37 @@
...
+down_revision = '047ebdac558b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'comment_moderator_report',
+        sa.Column('reported_by_id', sa.Integer(), nullable=False),
+        sa.Column('comment_id', sa.Integer(), nullable=False),
```
jace, line 24: Reorder as above. By convention user column should be `user_id`. We use other names only for secondary user column onwards.

funnel/views/account.py
iambibhas: We're not keeping existing reports once the fate of the comment is decided. This can amend previous report probably.

funnel/models/moderation.py
```diff
@@ -0,0 +1,56 @@
...
+        reported_by=reported_by, comment=self
+    ).one_or_none()
+    if report is None:
+        report = CommentModeratorReport(reported_by=reported_by, comment=self)
+        db.session.add(report)
+        db.session.commit()
+    return report
+
+
+Comment.report_spam = _report_comment
```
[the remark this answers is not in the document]
iambibhas: it'll be helpful I think.

funnel/models/moderation.py
iambibhas: Using UUID field as primary key.

funnel/models/moderation.py
```diff
@@ -0,0 +1,55 @@
...
+
+__all__ = ['MODERATOR_REPORT_TYPE', 'CommentModeratorReport']
+
+
+class MODERATOR_REPORT_TYPE(LabeledEnum):  # NOQA: N801
+    OK = (0, 'ok', __("OK"))
+    SPAM = (1, 'spam', __("Spam"))
+
+
+class CommentModeratorReport(IdMixin, UuidMixin, db.Model):
```
jace, line 14: `UuidMixin, BaseMixin`, no?

funnel/models/moderation.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    user_id = db.Column(None, db.ForeignKey('user.id'), nullable=False, index=True)
+    user = db.relationship(
+        User,
+        primaryjoin=user_id == User.id,
+        backref=db.backref('moderator_reports', cascade='all', lazy='dynamic'),
+    )
+    report_type = db.Column(
+        db.SmallInteger, nullable=False, default=MODERATOR_REPORT_TYPE.SPAM
+    )
```
jace, line 34: `reported_at` should be a distinct timestamp. Don't depend on `created_at`.

funnel/views/account.py
```diff
@@ -174,6 +181,102 @@ def siteadmin_comments_spam(self):
...
+
+        existing_reports = report.comment.moderator_reports.filter(
+            CommentModeratorReport.user != current_auth.user
+        )
+
+        report_form = ModeratorReportForm()
+        report_form.form_nonce.data = report_form.form_nonce.default()
+
+        if report_form.validate_on_submit():
+            app.logger.info('validation pass')
```
jace, line 223: `.debug`?

PR #671, opened 2020-05-22
Remove hash

vidya-ram:
1. Remove hash from navbar items
2. Update the styling of spotlight card
3. Maintain the same height for project cards
4. Increase the size of close icon of alert box
5. Add pace.js to show progress bar
6. Update the font size of project section headlines
7. Update the icon size of tabs since the tap area was small

Dependency: https://github.com/hasgeek/baseframe/pull/298

### DIFF
--- a/funnel/assets/js/app.js
+++ b/funnel/assets/js/app.js
@@ -92,7 +92,7 @@ if (
   )
 ) {
   $('.pace').addClass('pace-hide');
-  $window.onbeforeunload = function() {
+  window.onbeforeunload = function() {
     Pace.stop();
   };
 }
--- a/funnel/assets/js/app.js
+++ b/funnel/assets/js/app.js
@@ -84,11 +84,6 @@ $(() => {
   }
 });
 
-window.onbeforeunload = function() {
-  Pace.stop();
-  Pace.start();
-};
-
 if (
   navigator.userAgent.match(/(iPhone|Android)/) &&
   !(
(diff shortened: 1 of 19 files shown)


funnel/assets/js/app.js
```diff
@@ -83,3 +83,16 @@ $(() => {
     $.cookie('timezone', jstz.determine().name(), { path: '/' });
   }
 });
+
+if (
+  navigator.userAgent.match(/(iPhone|Android)/) &&
```
jace, line 88: This may need `iPad` also.

funnel/assets/js/util.js
```diff
@@ -38,8 +38,7 @@ export const Utils = {
     );
   },
   smoothScroll() {
-    $('a.js-smooth-scroll').on('click', function clickHandler(event) {
-      event.preventDefault();
+    $('a.js-smooth-scroll').on('click', function clickHandler() {
       Utils.animateScrollTo($(this.hash).offset().top);
```
jace, line 42: What does this do?

vidya-ram: Add the hash to the url for tabs. For ex: `#about`, `#talks`, `#venue` on the project page

funnel/assets/js/app.js
vidya-ram: iPad browser doesn't have a scrollbar so we can keep the pace scrollbar

PR #673, opened 2020-05-25
Project form fix

iambibhas:
- Rearranged project form fields to this order
```
- Title
- Tagline
- Location
- Timezone
- Banner image URL
- Description
```
- Added special tokens when tagging project locations

### DIFF
--- a/funnel/forms/project.py
+++ b/funnel/forms/project.py
@@ -47,9 +47,18 @@ class ProjectForm(forms.Form):
     )
     location = forms.StringField(
         __("Location"),
-        validators=[forms.validators.DataRequired(), forms.validators.Length(max=50)],
+        description=__(
+            '“Online”, “Bangalore”, “Chennai”, “Pune”, etc (without quotes)'
+        ),
+        validators=[
+            forms.validators.DataRequired(
+                __("If this project doesn’t have a fixed location, use “Online”")
+            ),
+            forms.validators.Length(
+                min=3, max=80, message=__("%(max)d characters maximum")
+            ),
+        ],
         filters=[forms.filters.strip()],
-        description=__("Eg. Bangalore, Mumbai, Pune, Online, Anywhere etc"),
     )
     timezone = forms.SelectField(
         __("Timezone"),
@@ -87,6 +96,7 @@ class ProjectForm(forms.Form):
 class ProjectLivestreamForm(forms.Form):
     livestream_urls = forms.TextListField(
         __("Livestream URLs. One per line."),
+        filters=[forms.filters.strip_each()],
         validators=[
             forms.validators.Optional(),
             forms.validators.ForEach(
--- a/funnel/forms/project.py
+++ b/funnel/forms/project.py
@@ -93,7 +93,7 @@ class ProjectForm(forms.Form):
     )
 
     def validate_location(self, field):
-        if re.findall(r'[\'"”]+', field.data):
+        if re.findall(r'["”]+', field.data):
             raise forms.ValidationError(__("Double quote not allowed in location name"))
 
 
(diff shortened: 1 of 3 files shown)


funnel/forms/project.py
```diff
@@ -39,22 +39,17 @@ class ProjectForm(forms.Form):
...
     )
-    description = forms.MarkdownField(
-        __("Project description"),
-        validators=[forms.validators.DataRequired()],
-        description=__("Landing page contents"),
+    location = forms.StringField(
+        __("Location"),
+        validators=[forms.validators.DataRequired(), forms.validators.Length(max=50)],
+        filters=[forms.filters.strip()],
+        description=__("Eg. Bangalore, Mumbai, Pune, Online, Anywhere etc"),
```
iambibhas, line 52: Could this be better?

jace: Use Hasjob’s hint and validator, but place “Online” first here (no “Anywhere”).

funnel/forms/project.py
```diff
@@ -39,22 +39,26 @@ class ProjectForm(forms.Form):
...
         description=__("One line description of the project"),
     )
-    description = forms.MarkdownField(
-        __("Project description"),
-        validators=[forms.validators.DataRequired()],
-        description=__("Landing page contents"),
+    location = forms.StringField(
+        __("Location"),
+        description=__(
+            '“Online”, “Bangalore”, “Chennai”, “Pune”, etc (without quotes)'
```
jace, line 51: I'd update this to be friendlier for the times:

`“Online” if this is online-only, else the city or region (without quotes)`

funnel/forms/project.py
```diff
@@ -39,22 +39,26 @@ class ProjectForm(forms.Form):
...
+        ),
+        validators=[
+            forms.validators.DataRequired(
+                __("If this project doesn’t have a fixed location, use “Online”")
```
jace, line 55: `If this project is online-only, use “Online”`

funnel/forms/project.py
```diff
@@ -39,22 +39,26 @@ class ProjectForm(forms.Form):
...
+            ),
+            forms.validators.Length(
+                min=3, max=80, message=__("%(max)d characters maximum")
+            ),
```
jace, line 59: A validator for quotes in the location, since people will submit `“Online”` or `"Online"` instead of `Online`.

funnel/jobs/jobs.py
```diff
@@ -43,7 +45,18 @@ def tag_locations(project_id):
...
+                url,
+                params={
+                    'q': project.location,
+                    'bias': ['IN', 'US'],
+                    'special': [
+                        __('Anywhere'),
+                        __('Home'),
+                        __('Internet'),
+                        __('Online'),
+                        __('Remote'),
```
jace, line 57: Since "Home", "Anywhere" and "Remote" are unlikely for location, keep just "Online" and "Internet".

Also, Online and Internet need non-localized English language versions in addition to localized versions since there is a high chance of a Hindi user typing "Online" instead of the Hindi word. That makes the list:

`['Online', 'Internet', __("Online"), __("Internet")]`

funnel/forms/project.py
```diff
@@ -82,11 +86,21 @@ class ProjectForm(forms.Form):
...
     )
+    description = forms.MarkdownField(
+        __("Project description"),
+        validators=[forms.validators.DataRequired()],
+        description=__("Landing page contents"),
+    )
+
+    def validate_location(self, field):
+        if re.findall(r'["”]+', field.data):
+            raise forms.ValidationError(__("Double quote not allowed in location name"))
```
jace, line 97: IMO, if using regex, always pre-compile using `re.compile` at module or class level. `re.search` will do as we don't care for _all_ instances. The regex only needs to be `["“”]` (without the `+`, but including closing quote for possible partial copy-paste)

Error message: "Quotes are not necessary in the location name"

PR #680, opened 2020-05-26
Replace strip_null with abort_null

qurbat:
Replaces `strip_null` with `abort_null`, throwing an `abort(400)` instead of stripping null bytes from user input

### DIFF
--- a/funnel/utils.py
+++ b/funnel/utils.py
@@ -3,7 +3,7 @@ import six
 import re
 import urllib.parse
 
-from flask import current_app
+from flask import current_app, abort
 
 import qrcode
 import qrcode.image.svg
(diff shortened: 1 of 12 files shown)


tests/test_utils.py
```diff
@@ -33,3 +38,8 @@ def test_split_name(self):
 
     def test_format_twitter_handle(self):
         assert format_twitter_handle("testusername") == "@testusername"
+
+    def test_null_abort(self, test_client):
+        with current_app.test_request_context('/'):
+            with pytest.raises(expected_exception=BadRequest):
+                abort_null('\x00')
```
jace, line 45: You need a second test to verify (a) it doesn't blow up given clean input, and (b) it returns the input.

funnel/utils.py
```diff
@@ -3,7 +3,7 @@
 import re
 import urllib.parse
 
-from flask import current_app
+from flask import current_app, abort
```
jace, line 6: This line has not been run through `isort`. Have you installed pre-commit?

funnel/utils.py
```diff
@@ -19,10 +19,12 @@
 # --- Utilities ---------------------------------------------------------------
 
 
-def strip_null(text):
-    # Removes null byte from given text
+def abort_null(text):
+    # Abort request if text contains nullbyte; if not, return text
```
jace, line 23: Make this a docstring.

PR #689, opened 2020-05-29
Users need to have at least one verified contact to become members or more than one to remove contacts

iambibhas:


funnel/views/membership.py
```diff
@@ -65,6 +65,18 @@ def new_member(self):
 
         if request.method == 'POST':
             if membership_form.validate_on_submit():
+                if not membership_form.user.data.email:
+                    # users without verified email cannot be members
+                    return (
+                        {
+                            'status': 'error',
+                            'message': _("User does not have a verified email address"),
```
jace, line 73: This change will conflict with the proposal we're considering: user must have one verified contact, an email address or a phone number.

I suggest making this a method on the user object itself: `has_verified_contact_info`

jace: It may also be prudent to flag this in the select2 widget itself, instead of erroring after the user submits. People will complain if names are missing, so instead show name (a) low priority, at bottom, or (b) crossed out or otherwise marked as unavailable.

jace: The verified contact info requirement should also extend to the founding member of an organization, meaning you can't create one until you've verified a phone or email.

funnel/models/user.py
```diff
@@ -27,6 +26,12 @@
 from . import BaseMixin, TSVectorType, UuidMixin, db
 from .helpers import add_search_trigger
 
+try:
+    from functools import cached_property
+except ImportError:
+    from werkzeug.utils import cached_property
```
jace, line 32: We can stick with werkzeug's cached_property. I checked the implementation and it is the same as Python 3.8's. The only difference is that werkzeug also includes a cache invalidation function, which Python doesn't. We need the invalidation occasionally, so it makes sense to keep importing from werkzeug.

funnel/models/user.py
```diff
@@ -192,6 +197,16 @@ def name(cls):  # NOQA: N805
...
         return self.status == USER_STATUS.ACTIVE
 
+    @cached_property
+    def verified_contact_count(self):
+        count = 0
+        count += len(self.emails)
+        count += len(self.phones)
+        return count
+
+    def has_verified_contact_info(self):
```
jace, line 207: This can be a property. Also, can't the test simply be `return bool(self.emails or self.phones)`

funnel/views/account.py
```diff
@@ -463,8 +463,14 @@ def remove_email(md5sum):
...
         if not useremail:
             abort(404)
-    if isinstance(useremail, UserEmail) and useremail.primary:
-        flash(_("You cannot remove your primary email address"), 'danger')
+    if (
+        isinstance(useremail, UserEmail)
+        and current_auth.user.verified_contact_count == 1
+    ):
+        flash(
+            _("You cannot remove your only verified email address or phone number"),
```
jace, line 471: "Your account requires at least one verified email address or phone number"

funnel/views/account.py
```diff
@@ -573,6 +579,15 @@ def remove_phone(number):
             # Block attempts to delete this number if verification failed.
             # It needs to be deleted in a background sweep.
             return render_redirect(url_for('account'), code=303)
+        if (
+            isinstance(userphone, UserPhone)
+            and current_auth.user.verified_contact_count == 1
+        ):
+            flash(
+                _("You cannot remove your only verified email address or phone number"),
```
jace, line 587: Same

funnel/views/membership.py
```diff
@@ -65,6 +65,20 @@ def new_member(self):
 
         if request.method == 'POST':
             if membership_form.validate_on_submit():
+                if not membership_form.user.data.has_verified_contact_info():
+                    # users without verified contact information cannot be members
+                    return (
+                        {
+                            'status': 'error',
+                            'message': _(
+                                "User does not have any verified contact information"
```
jace, line 74: "This user does not have any verified contact information. If you are able to contact them, please ask them to verify their email address or phone number"

jace: The gender-neutral "them" is going to be a problem in translation. Will require "his or her" in the target language. It's fine for English though.

funnel/views/membership.py
```diff
@@ -315,6 +329,19 @@ def new_member(self):
...
```
jace, line 338: Same as above.

funnel/views/organization.py
```diff
@@ -38,6 +44,16 @@ def index(self):
 
     @route('/new', methods=['GET', 'POST'])
     def new(self):
+        if not current_auth.user.has_verified_contact_info():
+            flash(
+                _(
+                    "You need to have at least one verified email address "
```
jace, line 50: "a" instead of "at least one", so "You need to have a verified email address…"

funnel/models/user.py
iambibhas: I'm using the count because it's cached already. so will be faster in templates.

PR #696, opened 2020-06-02
Add a 'this project is not published' alertbox

qurbat:
Add an`alertbox` for when a project is in the draft or withdrawn state

### DIFF
--- a/funnel/templates/macros.html.jinja2
+++ b/funnel/templates/macros.html.jinja2
@@ -98,6 +98,11 @@
           {{ alertbox('warning', 'This project will not be listed as it has no sessions in the schedule') }}
         </div>
       {% endif %}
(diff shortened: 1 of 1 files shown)


funnel/templates/macros.html.jinja2
```diff
@@ -98,6 +98,11 @@
           {{ alertbox('warning', 'This project will not be listed as it has no sessions in the schedule') }}
         </div>
       {% endif %}
+      {% if project.view_for('view').is_available() and not project.PUBLISHED %}
+      <div class="grid__col-sm-12">
+        {{ alertbox('warning', 'This project is currently not published.') }}
```
jace, line 103: No trailing period per style guide for captions.

PR #697, opened 2020-06-02
Add banner

vidya-ram:


funnel/models/project.py
```diff
@@ -503,13 +503,13 @@ def datelocation(self):
...
             and project.cfp_start_at is not None
             and project.cfp_start_at <= utcnow()
             and (project.cfp_end_at is None or utcnow() < project.cfp_end_at)
         ),
         lambda project: db.and_(
-            project.state.PUBLISHED,
+            Project.state.PUBLISHED,
             project.cfp_start_at.isnot(None),
             project.cfp_start_at <= db.func.utcnow(),
             db.or_(project.cfp_end_at.is_(None), db.func.utcnow() < project.cfp_end_at),
```
jace: Confused here. What is the `project` parameter in a SQL expression?

jace: (I need to go back and see what is happening in Coaster.)

PR #698, opened 2020-06-03
Set default value for 'is session a break period?' to false

qurbat:
When editing schedules, each new session by default has the 'Is this session a break period?' radio button ticked. This pull request sets the default value for `is_break` to `False`.


funnel/views/session.py
```diff
@@ -62,7 +62,7 @@ def session_form(project, proposal=None, session=None):
         del form.venue_room_id
     if request.method == 'GET':
         if not (session or proposal):
-            form.is_break.data = True
+            form.is_break.data = False
```
jace, line 65: Instead remove this block.

PR #700, opened 2020-06-03
Use baseframe date time filter

vidya-ram:


jace:
Slightly concerned by the use of custom formats. We should evaluate whether the default formats (short/medium/long) work for them. We may also allow users to customise date display from their account settings, in which case the formats will vary by user.

It's not a blocker at the moment however.

PR #704, opened 2020-06-04
Introduce blake2b hashes to replace SHA256 and MD5 (with matching digest sizes)

jace:
Also fixes a mistake in MD5 hashes, which were not using `lower(email)`.


jace:
This PR is ready by itself, but I intend to migrate UserEmail's md5sum to blake2b as well. This will affect all views that use md5 (verify email, reset password).

funnel/views/account.py
```diff
@@ -50,6 +52,20 @@
 from .sms import send_phone_verify_code
 
 
+def md5sum_or_blake2b_b58(text):
```
jace, line 55: This function is temporary until MD5 links are obsolete in a week or so. It can be removed then. We will need a more robust mechanism for casting from binary to base58 bidirectionally however, perhaps an equivalent of `uuid_b58`'s comparator in Coaster.

PR #705, opened 2020-06-04
Hide labels form if none are available

qurbat:
Hide labels form on proposal submission page if none are available in the project

### DIFF
--- a/funnel/forms/proposal.py
+++ b/funnel/forms/proposal.py
@@ -48,6 +48,8 @@ def proposal_label_form(project, proposal):
                     ],
                 ),
             )
+    if not project.labels:
+        return None
 
     form = ProposalLabelForm(
         obj=proposal.formlabels if proposal else None, meta={'csrf': False}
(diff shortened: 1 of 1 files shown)


funnel/forms/proposal.py
```diff
@@ -48,6 +48,8 @@ class ProposalLabelForm(forms.Form):
                     ],
                 ),
             )
+    if not project.labels:
+        return None
```
jace, line 52: Just `return` will do.

jace: Move this block right below the `def`. No need to define a class and iterate through labels if there are no labels.

funnel/forms/proposal.py
```diff
@@ -228,9 +230,15 @@ class ProposalForm(forms.Form):
...
 
     def set_queries(self):
-        self.formlabels.form = proposal_label_form(
+        label_form = proposal_label_form(
             project=self.edit_parent, proposal=self.edit_obj
         )
+        if label_form:
+            self.formlabels.form = proposal_label_form(
+                project=self.edit_parent, proposal=self.edit_obj
+            )
```
jace, line 239:
```python
if label_form:
    self.formlabels.form = label_form
```

Why are you calling `proposal_label_form` twice?

PR #706, opened 2020-06-05
Orgs are public by default

iambibhas:


jace:
We should also make private profile pages render to their owners, with a notice that it is private. This will replace the org's `view`, `edit` and `delete` endpoints completely, always using the profile's equivalents instead.

PR #714, opened 2020-06-08
EmailAddress model

jace:


funnel/models/email_address.py
```diff
@@ -0,0 +1,572 @@
...
+            hashlib.md5(  # NOQA: S303 # skipcq: PTC-W1003 # nosec
+                self.email_lower.encode('utf-8')
+            ).hexdigest()
+            if self.email_lower
+            else None
+        )
+
+    def __str__(self) -> str:
+        """Cast email address into a string."""
+        return str(self.email or '')
```
iambibhas, line 251: `self.email` is already a string, right? Why the `str()`?

tests/unit/models/test_email_address.py
```diff
@@ -0,0 +1,612 @@
...
+    assert ea2.email_lower == 'example@example.com'
+    assert ea2.blake2b160 == hash_map['example@example.com']
+    assert ea1.email_canonical == 'example@example.com'
+    assert ea2.blake2b160_canonical == hash_map['example@example.com']
+    assert str(ea2) == 'Example@example.com'
+    assert repr(ea2) == "EmailAddress('Example@example.com')"
+
+    # Canonical representation's hash can be distinct from regular hash
+    ea3 = EmailAddress('Example+Extra@example.com')
+    assert ea3.email == 'Example+Extra@example.com'
```
iambibhas, line 107: Why are we not using lower case by default? Email isn't case-sensitive, is it?

funnel/models/email_address.py
jace: Fixed.

tests/unit/models/test_email_address.py
jace: Email is technically case sensitive, although in practice it isn't. Our implementation preserves the user's choice of case but enforces uniqueness against the lowercase representation.

funnel/models/email_address.py
```diff
@@ -0,0 +1,623 @@
...
+        self.delivery_state_at = db.func.utcnow()
+
+    def refcount(self) -> int:
+        """Returns count of references to this EmailAddress instance"""
+        return sum(
+            len(getattr(self, backref_name)) for backref_name in self.__backrefs__
+        )
+
+    @classmethod
+    def mark_blocked(cls, email: str) -> None:
```
iambibhas, line 322: Should there be an instance method to mark itself as blocked?

funnel/models/email_address.py
```diff
@@ -0,0 +1,623 @@
...
+        Looks up all existing instances of EmailAddress with the same canonical
+        representation and amends them to forget the email address and set the
+        :attr:`is_blocked` flag.
+        """
+        for obj in cls.get_canonical(email, is_blocked=False).all():
+            obj.email = None
+            obj._is_blocked = True
+
+    @classmethod
+    def get_filter(
```
iambibhas, line 335: This one seems like a private method. We'll expect `.get()` to be used from everywhere else. Should prefix this with underscore.

funnel/models/email_address.py
jace: Can't, as all related instances need to be blocked.

funnel/models/email_address.py
jace: It's not. It's used in UserEmail and UserEmailClaim.

funnel/models/email_address.py
```diff
@@ -0,0 +1,623 @@
...
+        Optionally filtered by the :attr:`is_blocked` flag.
+        """
+        hashes = [
+            email_blake2b160_hash(result)
+            for result in canonical_email_representation(email.lower())
+        ]
+        query = cls.query.filter(cls.blake2b160_canonical.in_(hashes))
+        if is_blocked is not None:
+            query = query.filter_by(_is_blocked=is_blocked)
+        return query
```
iambibhas, line 388: Should this be `.all()`? or returning query is fine?

jace: I'm moving away from enforcing `.all()` in `get*` methods.

1. The generator interface is a seamless replacement for a list
2. Additional filters can be applied if required

We are discouraging use of filters in views, but not in related models.

funnel/models/email_address.py
jace: This makes unblocking even harder. Given an email address, we can only find one variant to unblock. There is no way to unblock a `+sub-address` variant without explicitly knowing the variant.

PR #717, opened 2020-06-09
Send a confirmation email upon registering 

qurbat:


funnel/jobs/jobs.py
```diff
@@ -13,9 +13,12 @@
...
 
 @rq.job('funnel')
-def send_mail_async(sender, to, body, subject):
+def send_mail_async(sender, to, body, subject, file=None):
     with app.app_context():
-        send_mail(sender, to, body, subject)
+        if file is not None:
+            send_mail(sender, to, body, file, subject)
+        else:
+            send_mail(sender, to, body, subject)
```
iambibhas, line 21: Check if it's ok to pass `file=None` to `send_mail`, then you dont need this if-else block.

funnel/templates/project_registered_email.html.jinja2
```diff
@@ -0,0 +1,7 @@
+Hi {{ current_auth.user.fullname.split()[0] }},
```
iambibhas, line 1: It's better if you can add a property called `first_name` in User models and do this there. Or just use fullname.

funnel/views/helpers.py
```diff
@@ -192,10 +192,12 @@ def url_join(base, url=''):
     return urljoin(base, url)
 
 
-def send_mail(sender, to, body, subject):
+def send_mail(sender, to, body, subject, file=None):
     msg = Message(sender=sender, subject=subject, recipients=[to])
     msg.body = body
     msg.html = markdown(msg.body)  # FIXME: This does not include HTML head/body tags
+    if file is not None:
```
iambibhas, line 199: Same comment as the async function.

funnel/views/project.py
```diff
@@ -571,7 +580,10 @@ def deregister(self):
     @requires_login
     @requires_roles({'concierge'})
     def rsvp_list(self):
-        return {'project': self.obj, 'statuses': RSVP_STATUS}
+        return {
+            'project': self.obj,
+            'going_rsvps': self.obj.rsvps_with(RSVP_STATUS.YES),
```
iambibhas, line 585: pull master in this branch.

funnel/views/schedule.py
```diff
@@ -124,6 +124,24 @@ def schedule_data(project, with_slots=True, scheduled_sessions=None):
     return schedule
 
 
+def schedule_ical(sched):
+    cal = Calendar()
+    cal.add('prodid', "-//HasGeek//NONSGML Funnel//EN")
```
iambibhas, line 129: Do we need to mention Funnel here?

funnel/templates/project_registered_email.html.jinja2
jace: This should remain `fullname`. Do not assume the first part is the given name.

PR #724, opened 2020-06-11
Anyone can report spam

iambibhas:


funnel/templates/comments.html.jinja2
```diff
@@ -41,7 +41,7 @@
               data-delete-url="{{ comment.url_for('delete') }}"
               data-edit-url="{{ comment.url_for('edit') }}"
             {% endif %}
-            {% if current_auth.user.is_comment_moderator %}
+            {% if comment.view_for('report_spam').is_available() %}
```
jace, line 44: This will be slow inside a loop as it inits the view class. If the reporting feature access is by user, set a flag outside the loop. However, if the feature is per comment, then it needs to be a feature function.

iambibhas: I can skip the `is_available()` check I think. Any logged in user can report. so only checking `current_auth.user` should be fine.

PR #731, opened 2020-06-17
Update account form UI and add password strength check and indicator to UI

vidya-ram:


funnel/views/account.py
```diff
@@ -488,14 +489,7 @@ def change_password():
...
         )
-    return render_form(
-        form=form,
-        title=_("Change password"),
-        formid='changepassword',
-        submit=_("Change password"),
-        cancel_url=url_for('account'),
-        ajax=True,
-    )
+    return render_template('change_password.html.jinja2', form=form, title=_("Change password"), formid='changepassword', ref_id='form-changepassword')
```
iambibhas, line 492: How did this skip pre-commit checks?

funnel/views/login.py
```diff
@@ -317,7 +320,7 @@ def register():
         db.session.commit()
         flash(_("You are now one of us. Welcome aboard!"), category='success')
         return redirect(get_next_url(session=True), code=303)
-    return render_template('signup_form.html.jinja2', form=form, login_registry=login_registry)
+    return render_template('signup_form.html.jinja2', form=form, login_registry=login_registry, title=_("Register account"),  formid='registeraccount', ref_id='form-registeraccount')
```
iambibhas, line 323: same as this.

funnel/templates/password_login_form.html.jinja2
```diff
@@ -1,7 +1,7 @@
 {% from "baseframe/forms.html.jinja2" import renderfield, rendersubmit %}
 
-{% macro passwordlogin(loginform, markup) %}
-  <form id="form-passwordlogin" method="POST" class="mui-form mui-form--margins" accept-charset="UTF-8">
+{% macro passwordlogin(loginform, ref_id, markup) %}
```
jace, line 3: Drop the markup parameter. It's used for the forgot password link, but that link should not be in field helper text as it can't be translated. It has to move out of the form.

funnel/views/account.py
```diff
@@ -373,7 +388,7 @@ def account_edit(newprofile=False):
             formid='account_edit',
             submit=_("Save changes"),
             ajax=True,
-            cancel_url=url_for('account') if not newprofile else None,
+            template='account_formlayout.html.jinja2',
```
jace, line 391: Didn't this parameter change to `layout_template`? It should.

funnel/views/account.py
```diff
@@ -491,10 +506,10 @@ def change_password():
     return render_form(
         form=form,
         title=_("Change password"),
-        formid='changepassword',
+        formid='password-change',
         submit=_("Change password"),
-        cancel_url=url_for('account'),
```
jace, line 496: User should be able to cancel this operation.

jace: Mobile layout has an X button. If web layout also does, this is fine.

funnel/views/account.py
```diff
@@ -647,7 +663,7 @@ def verify_email(email_hash):
         ),
         formid="email_verify",
         submit=_("Send"),
-        cancel_url=url_for('account'),
```
jace, line 650: Same as above.

funnel/templates/account_merge.html.jinja2
```diff
@@ -29,11 +27,7 @@
   </ul>
 {% endmacro %}
 
-{% block contentwrapper %}
-  <div class="alert alert--error cookies-required-alert alert--dismissable">
-    <a class="alert__close" href="javascript:void(0);" aria-label="close">{{ faicon(icon='times', icon_size='title') }}</a>
-    <p class="alert__text">{% trans %}Cookies are required to perform this operation. Please enable cookies in your browser’s settings and reload this page.{% endtrans %}</p>
-  </div>
+{% block loginboxcontent %}
```
jace, line 30: Looks like this block's name is too specific for one use case. May need a better name, like "focuscontent".

funnel/views/account.py
```diff
@@ -73,18 +74,31 @@ def blake2b_b58(text):
...
+                'is_weak': bool(failed_tests),
+                'failed_tests': [repr(t) for t in failed_tests],
+            },
+        }
+    else:
+        return (
+            {
+                'status': 'error',
+                'error_code': 'policy_form_error',
+                'error_description': _("Something went wrong. Please try again"),
```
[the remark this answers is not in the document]
iambibhas: This is in case of CSRF/Nonce error. "Please reload and try again?"

funnel/views/account.py
vidya-ram: Add X to web layout

PR #733, opened 2020-06-18
EmailAddress improvements

jace:


tests/unit/models/test_email_address.py
```diff
@@ -356,14 +357,25 @@ def test_email_address_delivery_state(clean_db):
...
     assert ea.delivery_state.ACTIVE
     assert str(ea.delivery_state_at) == str(db.func.utcnow())
 
+    # Active state cannot be downgraded to normal state
+    with pytest.raises(StateTransitionError):
+        ea.mark_sent()
+
+    # The proper protocol is to call mark_sent() only when it's available:
+    if ea.mark_sent.is_available:
+        ea.mark_sent()
```
iambibhas, line 377: I think the check should be whether `mark_sent()` is available or not. The lines above this block calls `mark_sent()` so we know it's available.

jace: Isn't that what the check is doing?

jace: I'm adding an `assert False` to confirm that the branch isn't being called, as there is no test otherwise.

jace: Oh, "check" meaning `assert` instead of `if`. Changing it to `assert not ea.mark_sent.is_available`

funnel/templates/email_account_reset.html.jinja2
```diff
@@ -0,0 +1,12 @@
+{% extends "email_layout.html.jinja2" %}
+{% block content -%}
+
+  <p>{% trans %}Hello {{ fullname }},{% endtrans %}</p>
```
vidya-ram, line 4: Does translation work in this case?
`{% trans name=user.fullname %}Hello {{ name }},{% endtrans %}`

funnel/templates/proposal_comment_to_proposer_email.md.jinja2
```diff
@@ -1,6 +0,0 @@
-**{{ current_auth.user.pickername }}** replied to you on **{{ proposal.title
-}}**.
-
-{{ comment.message }}
-
-View it on [funnel]({{link}})
```
vidya-ram, line 6: We are not sending email when user comments on a proposal?

funnel/templates/email_account_reset.html.jinja2
jace: Both should work, but I'll check again.

funnel/templates/proposal_comment_to_proposer_email.md.jinja2
jace: It was removed last month due to spam.

funnel/templates/email_account_reset.html.jinja2
jace: This does work. I'm merging this PR.

PR #739, opened 2020-06-25
Seo fixes

vidya-ram:
1. Fix schema
- 1. location of schema for projects without venue
- 2. Add eventAttendanceMode
- 3. Add organizer
- 4. remove invalid offer
2. Add profile name to project page title

### DIFF
--- a/funnel/templates/project.html.jinja2
+++ b/funnel/templates/project.html.jinja2
@@ -1,4 +1,5 @@
 {% extends "layout.html.jinja2" %}
+{% set title_suffix = project.profile.title %}
 {% from "baseframe/forms.html.jinja2" import ajaxform %}
 {% from "baseframe/components.html.jinja2" import faicon %}
 {%- from "macros.html.jinja2" import project_header %}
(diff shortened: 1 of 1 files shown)


funnel/templates/project.html.jinja2
```diff
@@ -1,4 +1,5 @@
 {% extends "layout.html.jinja2" %}
+{% set title_suffix = project.profile.title %}
```
jace, line 2: This is not being used. Also, it's usually a prefix.

funnel/templates/project.html.jinja2
```diff
@@ -17,6 +18,8 @@
       "name": {{ project.title|tojson }},
       "url": {{ project.url_for(_external=true)|tojson }},
       "startDate": {{ project.schedule_start_at_localized|tojson }},
+      "endDate": {{ project.schedule_end_at_localized|tojson }},
+      "eventStatus": "https://schema.org/EventScheduled",
```
jace, line 22: I think this needs an `if` condition that checks for `schedule_start_at` being null.

funnel/templates/project.html.jinja2
vidya-ram: We have defined it in layout
https://github.com/hasgeek/funnel/blob/master/funnel/templates/layout.html.jinja2#L5

funnel/templates/project.html.jinja2
vidya-ram: We are checking it on line 13

PR #745, opened 2020-06-29
Create robots.txt

qurbat:
`Disallow` paths that don't provide indexing value

### DIFF
--- a/funnel/static/robots.txt
+++ b/funnel/static/robots.txt
@@ -2,5 +2,5 @@
 # www.google.com/support/webmasters/bin/answer.py?hl=en&answer=156449
 
 User-agent: *
-Disallow: /account/
-Disallow: /login/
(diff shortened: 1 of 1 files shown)


funnel/static/robots.txt
```diff
@@ -0,0 +1,6 @@
+# www.robotstxt.org/
+# www.google.com/support/webmasters/bin/answer.py?hl=en&answer=156449
+
+User-agent: *
+Disallow: /account/
+Disallow: /login/
```
jace, line 6: Does this also block these URLs without a trailing slash? Both `/account` and `/login` are valid URLs.

qurbat: Updated, now it should match both `/login` and `/login/github` for instance. See: https://developers.google.com/search/reference/robots_txt

PR #749, opened 2020-07-01
Lazy load past projects on home and profile page

vidya-ram:


funnel/models/project.py
```diff
@@ -921,7 +921,9 @@ def all_unsorted(cls, legacy=None):
         """
         Return currently active events, not sorted.
         """
-        projects = cls.query.filter(cls.state.PUBLISHED)
+        from .venue import Venue
```
jace, line 924: This should be a tail import. Importing here will do an import every time the function is called.

Some of the other files have tail imports, for reference.

funnel/views/index.py
```diff
@@ -107,14 +99,31 @@ def whoami():
         return jsonify(message="Hmm, so who _are_ you?", code=401)
 
 
-@app.route('/json')
-def all_projects_json():
+@app.route('/past.json')
+@requestargs(('page', int), ('per_page', int))
+def past_projects_json(page=1, per_page=10):
```
jace, line 104: These should be the same parameter names as in /search (I haven't checked)

iambibhas: They are same.

PR #754, opened 2020-07-06
Sitemap generation

qurbat:
Generate a sitemap consisting of public profiles, published projects, and available proposals. The view is cached for a period of two hours once generated.

### DIFF
--- a/funnel/views/index.py
+++ b/funnel/views/index.py
@@ -176,6 +176,30 @@ def opensearch():
     return render_template('opensearch.xml.jinja2')
 
 
+@app.route('/sitemap.xml')
+def sitemap():
+    sitemapxml = (
+        '<?xml version="1.0" encoding="UTF-8"?>\n'
+        '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">\n'
+    )
+    projects = []
+    for project in Project.query.all():
+        projects.append(project.id)
+        sitemapxml += (
+            '  <url>\n'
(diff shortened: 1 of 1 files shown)


funnel/views/index.py
```diff
@@ -176,6 +177,60 @@ def opensearch():
...
+                '    <lastmod>{updated_at}</lastmod>\n'
+                '    <changefreq>monthly</changefreq>\n'
+                '  </url>\n'.format(
+                    url=project.url_for(_external=True),
+                    updated_at=project.updated_at.isoformat(),
+                )
+            )
+    proposals = []
+    for proposal in Proposal.query.all():
+        if not proposal.state.DELETED:
```
iambibhas, line 217: Also have to exclude `DRAFT` proposals. Not sure about `CANCELLED` and `REJECTED`.

funnel/views/index.py
```diff
@@ -176,6 +177,60 @@ def opensearch():
...
+                '    <lastmod>{updated_at}</lastmod>\n'
+                '    <changefreq>monthly</changefreq>\n'
+                '  </url>\n'.format(
+                    url=profile.url_for(_external=True),
+                    updated_at=profile.updated_at.isoformat(),
+                )
+            )
+    projects = []
+    for project in Project.query.all():
+        if project.state.PUBLISHED:
```
iambibhas, line 203: Don't do this. Add this condition as a filter in the query right above it.

jace:
There is a new implementation in #935, so I'm closing this.

PR #756, opened 2020-07-08
Update profile page UI

vidya-ram:
Screen Shot 2020-07-08 at 9 23 06 PM
Screen Shot 2020-07-08 at 9 23 11 PM

<hr>

Screen Shot 2020-07-08 at 9 22 33 PM
Screen Shot 2020-07-08 at 9 22 53 PM

### DIFF
--- a/funnel/models/profile.py
+++ b/funnel/models/profile.py
@@ -121,6 +121,7 @@ class Profile(UuidMixin, BaseMixin, db.Model):
     __roles__ = {
         'all': {
             'read': {
+                'features',
                 'urls',
                 'uuid_b58',
                 'name',
@@ -130,6 +131,7 @@ class Profile(UuidMixin, BaseMixin, db.Model):
                 'user',
                 'organization',
                 'banner_image_url',
+                'forms',
             },
             'call': {'url_for'},
         }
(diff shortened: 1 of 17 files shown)


funnel/models/profile.py
```diff
@@ -121,6 +121,7 @@ class Profile(UuidMixin, BaseMixin, db.Model):
     __roles__ = {
         'all': {
             'read': {
+                'features',
```
jace, line 124: This should be in `call`

funnel/models/profile.py
```diff
@@ -130,6 +131,7 @@ class Profile(UuidMixin, BaseMixin, db.Model):
                 'user',
                 'organization',
                 'banner_image_url',
+                'forms',
```
jace, line 134: Should be in `call`

funnel/views/profile.py
```diff
@@ -22,6 +22,11 @@
 from .mixins import ProfileViewMixin
 
 
+@Profile.features('new_project')
+def feature_profile_new_project(obj):
+    return bool(obj.state.PUBLIC)
```
jace, line 27: Also test if the user has access. Remove that test from the template and do it here.

PR #759, opened 2020-07-14
Change password strength calculations to zxcvbn library

vidya-ram:
Dependency https://github.com/hasgeek/baseframe/pull/307

### DIFF
--- a/funnel/views/account.py
+++ b/funnel/views/account.py
@@ -82,7 +82,7 @@ def password_policy_check():
         tested_password = password_policy.test_password(
             policy_form.candidate.data,
             user_inputs=(
-                [current_auth.user.fullname, current_auth.user.email]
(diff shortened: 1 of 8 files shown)


funnel/views/account.py
```diff
@@ -79,14 +79,29 @@ def password_policy_check():
     policy_form.form_nonce.data = policy_form.form_nonce.default()
 
     if policy_form.validate_on_submit():
-        tested_password = password_policy.password(policy_form.candidate.data)
-        failed_tests = tested_password.test()
+        tested_password = password_policy.test_password(
+            policy_form.candidate.data,
+            user_inputs=(
+                [current_auth.user.fullname, current_auth.user.email]
```
jace, line 85: Shouldn't this be `str(current_auth.user.email)`? Also, it may be None. Also, email claims.

PR #771, opened 2020-07-20
Don't autocomplete on just '@'

jace:


funnel/models/user.py
```diff
@@ -447,7 +447,7 @@ def autocomplete(cls, query):
         # and doesn't use an index in PostgreSQL. There's a functional index for lower()
         # defined above in __table_args__ that also applies to LIKE lower(val) queries.
 
-        if like_query == '%':
+        if like_query == '%' or like_query == '@%':
```
jace, line 450: `if like_query in ('%', '@%'):`

funnel/models/user.py
```diff
@@ -521,6 +521,8 @@ def autocomplete(cls, query):
                 .union(base_users)
                 .all()
             )
+        elif query == '@':
+            return []
```
jace, line 525: This should be unnecessary given the change in `like_query` above.

tests/unit/lastuser_core/test_model_user_User.py
```diff
@@ -287,6 +287,12 @@ def test_user_autocomplete(self):
         assert query_for_piglet == [piglet]
         query_for_crusoe = models.User.autocomplete(queries[2])
         assert query_for_crusoe == [crusoe]
+        query_at_character = models.User.autocomplete('@[')
+        assert query_at_character == []  # Test for empty searches
+        query_square_brackets = models.User.autocomplete('[[]]')
+        assert query_square_brackets == []
+        query_for_crusoe_with_square_brackets = models.User.autocomplete(queries[3])
+        assert query_for_crusoe_with_square_brackets == [crusoe]
```
jace, line 295: Assignment to variable is not required. You can simply do `assert function_call() == result`

PR #778, opened 2020-07-23
UI changes for login session box

vidya-ram:


StephanieBr:
- [ ] Font size should be 16 & 14px on desktop and 14 & 12px on mobile. 

- [ ] Align icon to text.  Screenshot 2020-07-23 at 1 48 21 PM

- [ ] Device name should be outside brackets and OS should be inside brackets . Screenshot 2020-07-23 at 1 54 00 PM

- [ ] Cannot see the IP address and service provider info anymore. Looks like this   - Screenshot 2020-07-23 at 1 59 27 PM

PR #779, opened 2020-07-23
Project updates

iambibhas:


funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+
+        return render_form(
+            form=post_form,
+            title=_("Post an update"),
+            submit=_("Save & preview"),
+            cancel_url=self.obj.url_for(),
+        )
+
+
+ProjectPostView.init_app(app)
```
jace, line 87: Talkfunnel version missing.

funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@Post.features('publish')
+def post_publishable(obj):
+    return obj.state.DRAFT and 'editor' in obj.roles_for(current_auth.user)
+
+
+@Post.views('main')
+@route('/<profile>/<project>/updates/<url_name_uuid_b58>')
```
jace, line 96: `url_name_uuid_b58` here should be `post`.

funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+class ProjectPostDetailsView(UrlForView, ModelView):
+    __decorators__ = [legacy_redirect]
+    model = Post
+    route_model_map = {
+        'profile': 'project.profile.name',
+        'project': 'project.name',
+        'url_name_uuid_b58': 'url_name_uuid_b58',
+    }
+
+    def loader(self, profile, project, url_name_uuid_b58):
```
jace, line 106: Similarly, `post`

funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@Post.features('publish')
+def post_publishable(obj):
+    return obj.state.DRAFT and 'editor' in obj.roles_for(current_auth.user)
+
+
+@Post.views('main')
+@route('/<profile>/<project>/updates/<url_name_uuid_b58>')
+class ProjectPostDetailsView(UrlForView, ModelView):
```
jace, line 97: `UrlChangeCheck` base class needed

funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+            .one_or_404()
+        )
+
+        post = (
+            self.model.query.join(Project)
+            .filter(
+                Project.id == project.id, Post.url_name_uuid_b58 == url_name_uuid_b58,
+            )
+            .one_or_404()
+        )
```
jace, line 119: These can be a single query.

funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+
+        if post.visibility_state.RESTRICTED and not (
+            project.current_roles.participant or project.current_roles.crew
+        ):
+            abort(403)
```
jace, line 124: Don't abort, just hide the body and ask the user to login or become a participant. This is a crucial requirement for upselling participation.

funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+
+        g.profile = post.project.profile
+        return post
+
+    @route('', methods=['GET'])
+    @render_with('project_post_details.html.jinja2')
+    @requires_roles({'reader'})
+    def view(self):
+        project_save_form = SavedProjectForm()
```
jace, line 133: Why does the Post view need to be aware of what forms the Project template requires? This should be coming out of a base class. Have a `ProjectFeatureMixin` or `ProjectHeaderMixin` class that provides a method that all these views can call. Pass these to the template namespaced: `'project_forms': self.project_header_forms()`. Let the template render using `project_forms.save` instead of `project_save_form`.

funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+        post_form = PostForm(obj=self.obj)
+
+        if post_form.validate_on_submit():
+            post_form.populate_obj(self.obj)
+            db.session.commit()
+            flash(_("The update has been edited"), 'success')
+            if self.obj.state.DRAFT:
+                return redirect(self.obj.url_for())
+            else:
+                return redirect(self.obj.project.url_for('posts'))
```
jace, line 176: Not required. Always redirect to permalink

funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+    @requires_roles({'editor'})
+    def edit(self):
+        post_form = PostForm(obj=self.obj)
+
+        if post_form.validate_on_submit():
+            post_form.populate_obj(self.obj)
+            db.session.commit()
+            flash(_("The update has been edited"), 'success')
+            if self.obj.state.DRAFT:
+                return redirect(self.obj.url_for())
```
jace, line 174: Redirect after POST must use code 303.

funnel/views/post.py
```diff
@@ -0,0 +1,208 @@
...
+        if delete_form.validate_on_submit():
+            self.obj.delete(actor=current_auth.user)
+            db.session.commit()
+            flash(_("The update has been deleted"), 'success')
+            return redirect(self.obj.project.url_for('posts'))
+
+        return render_form(
+            form=delete_form,
+            title=_("Delete update"),
+            message=_("Do you really wish to delete this post ‘{title}’? ").format(
```
jace, line 200: Needs better language.

funnel/views/project.py
```diff
@@ -199,6 +199,11 @@ def feature_project_comment_new(obj):
     return obj.current_roles.participant is True
 
 
+@Project.features('post_new')
```
jace, line 202: `post_update` (feature is called "update", while "post" is verb)

funnel/views/post.py
iambibhas: If we make it a single query, then it joins the Profile as well and no updates are returned because for project posts, profile is none.

jace: Incorrect join condition then.

PR #780, opened 2020-07-24
Split out login helpers into a separate login_session.py

jace:
Also:
* Set SameSite attribute on cookies
* Move the view-ish `user_session.access()` method into a view function
* Move user_session accessed_at processing into a post-request function that runs after the request has returned a response


jace:
Merging since there are no issues reported.

PR #782, opened 2020-07-27
Improved password strength validator and added PasswordCreateForm

iambibhas:


funnel/forms/account.py
```diff
@@ -35,35 +36,39 @@
...
+            message = _(
+                "This password is too simple. Add complexity by making it longer and using "
+                "a mix of upper and lower case letters, numbers and symbols"
+            )
+        self.message = message
+
+    def __call__(self, form, field):
+        user_inputs = []
+        for field_name in self.user_input_fields:
+            if hasattr(form, field_name) and getattr(form, field_name) is not None:
```
jace, line 52: Don't use `hasattr` if the form is expected to declare its fields. This will hide bugs.

funnel/forms/account.py
```diff
@@ -35,35 +36,39 @@
...
+        # rendering into UI
+        form.password_strength = float(tested_password['score'])
+        # No test failures? All good then
+        if not tested_password['is_weak']:
+            return
+        # Tell the user to make up a better password
+        raise forms.validators.StopValidation(
+            tested_password['warning']
+            if tested_password['warning']
+            else ' '.join(tested_password['suggestions'])
```
jace, line 68: IMO `'\n'` or even `'<br>'` (but the whole thing will need a `Markup()` wrapper and each suggestion will need `escape()`).

funnel/forms/account.py
```diff
@@ -35,35 +36,39 @@
...
-    raise forms.validators.StopValidation(
-        tested_password['warning']
-        if tested_password['warning']
-        else ' '.join(tested_password['suggestions'])
-        if tested_password['suggestions']
-        else _(
-            "This password is too simple. Add complexity by making it longer and using "
-            "a mix of upper and lower case letters, numbers and symbols"
+class PasswordStrengthValidator:
+    def __init__(self, user_input_fields=[], message=None) -> None:
```
jace, line 40: Default value should be `()` or `None`, never `[]`.

funnel/views/account.py
```diff
@@ -576,9 +576,7 @@ def confirm_email(email_hash, secret):
 @requires_login
 def change_password():
     if not current_auth.user.pw_hash:
-        form = PasswordResetForm()
-        form.edit_user = current_auth.user
-        del form.username
+        form = PasswordCreateForm()
```
jace, line 579: Doesn't this need `edit_user`? It has to check candidate against the user's fullname/username.

funnel/views/account.py
```diff
@@ -576,9 +576,7 @@ def confirm_email(email_hash, secret):
...
     else:
         form = PasswordChangeForm()
         form.edit_user = current_auth.user
```
jace, line 582: Since this is repeated here, it could be moved out of the `if` block.

PR #783, opened 2020-07-28
Add boxoffice_data to project call. Change post title to link.

vidya-ram:


funnel/templates/macros.html.jinja2
```diff
@@ -828,7 +828,7 @@
             </div>
           </div>
         </div>
-        <p class="mui--text-headline mui--text-bold">{{ post.title }}</p>
+        <p class="mui--text-headline mui--text-bold"><a class="mui--text-dark" :href="post.urls.project_view">{{ post.title }}</a></p>
```
jace, line 831: Also needs a link from the date.

vidya-ram: I have added link to update and date https://github.com/hasgeek/funnel/pull/783/files#diff-a5041c530595e2f582d11e4983d836b2R818

PR #796, opened 2020-07-31
Add version tag to static images

vidya-ram:


jace:
This will break hardcoded references to the URLs. Better to add `?v=2` to URLs.

PR #797, opened 2020-07-31
Make webmanifest a Py dict for JSON validation

jace:
Also repeat the manifest tag in uppercase to bypass adblocker filters.


jace:
There was an error in the HTML comment that I've fixed and deployed.

PR #804, opened 2020-08-03
fixed reload loop for reviews

iambibhas:


funnel/views/account.py
```diff
@@ -320,12 +320,12 @@ def siteadmin_review_comment(self, report):
...
-            flash(_("You cannot review same comment twice"), 'error')
-            return redirect(url_for('siteadmin_review_comments_random'))
 
+        all_user_reports = db.session.query(CommentModeratorReport.id).filter_by(
+            user_id=current_auth.user.id
+        )
         existing_reports = report.comment.moderator_reports.filter(
-            CommentModeratorReport.user != current_auth.user
+            ~CommentModeratorReport.id.in_(all_user_reports)
         )
```
jace, line 329: This is same query.

funnel/views/account.py
```diff
@@ -351,7 +358,8 @@ def siteadmin_review_comment(self, report):
                 if most_common_two[0][0] == MODERATOR_REPORT_TYPE.SPAM:
```
jace, line 358: This needs to be a namedtuple.

jace (approved):

PR #809, opened 2020-08-05
Switch email availability validation to EmailAddress

jace:


funnel/forms/account.py
```diff
@@ -84,6 +86,44 @@ def __call__(self, form, field):
...
+            "This account is for you as an individual. We’ll make one for your organization later"
+        ),
+        validators=[forms.validators.DataRequired(), forms.validators.Length(max=80)],
+    )
+    email = forms.EmailField(
+        __("Email address"),
+        validators=[
+            forms.validators.DataRequired(),
+            EmailAddressAvailable(),
+            forms.validators.ValidEmail(),
```
jace, line 105: ValidEmail does a DNS check that EmailAddressAvailable does not, but it should.

PR #814, opened 2020-08-06
Added pagination to user projects view

iambibhas:


iambibhas:
Closing this. This didn't come up as a requirement. I created this thinking it'll be needed. But now it's become stale because it's not. When needed, I'll add these back again.

PR #815, opened 2020-08-06
User avatar UI

vidya-ram:


funnel/templates/contacts.html.jinja2
```diff
@@ -72,7 +72,11 @@
                     {% for contact in daterow.contacts.all() %}
                       <div class="user">
                         <div class="user__box">
-                          <div class="user__box__gravatar user__box__gravatar--initials">{{ contact.participant.fullname|initials }}</div>
+                          {% if contact.participant.user and contact.participant.user.avatar %}
+                            <img class="user__box__gravatar" src="{{ contact.participant.user.avatar }}"/>
```
jace, line 76: This should use the user_avatar macro as well.

funnel/templates/contacts.html.jinja2
```diff
@@ -72,7 +72,11 @@
...
+                          {%- else %}
+                            <div class="user__box__gravatar user__box__gravatar--initials">{{ contact.participant.fullname|initials }}</div>
```
jace, line 78: Participant model should also get an avatar property that returns the user avatar if user is present. This if block can be removed then, replaced with `{{ user_avatar(participant) }}`

funnel/templates/macros.html.jinja2
```diff
@@ -954,3 +955,11 @@
     </tbody>
   </table>
 {%- endmacro %}
+
+{%- macro user_avatar(user, css_class='') %}
+  {% if user.profile and user.profile.logo_url.url %}
```
jace, line 960: Use `user.avatar` instead, so it can also work for Participant model.

migrations/versions/99e5e7b54104_drop_user_avatar_field.py
```diff
@@ -0,0 +1,26 @@
+"""drop user avatar field
+
+Revision ID: 99e5e7b54104
+Revises: dcd0870c24cc
```
iambibhas, line 4: Put `ea1ea3b0ff95` here.

PR #816, opened 2020-08-07
Use meta-refresh redirect to work around browser bugs with setting cookies on 30x redirects

qurbat:


funnel/views/login.py
```diff
@@ -526,10 +527,13 @@ def login_service_postcallback(service, userdata):
 
     if 'merge_buid' in session:
         return set_loginmethod_cookie(
-            redirect(url_for('account_merge', next=login_next), code=303), service
+            Response(render_template('meta_refresh.html.jinja2', url=login_next)), service,
```
jace, line 530: Redirect to account_merge

PR #846, opened 2020-08-26
Include Organizer name in iCal

qurbat:
Insert profile name as value for the `ORGANIZER` field in each event in the schedule iCal

### DIFF
--- a/funnel/views/schedule.py
+++ b/funnel/views/schedule.py
@@ -153,7 +153,9 @@ def session_ical(session):
 
     event = Event()
     event.add('summary', session.title)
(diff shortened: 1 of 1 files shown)


funnel/views/schedule.py
```diff
@@ -153,6 +153,9 @@ def session_ical(session):
 
     event = Event()
     event.add('summary', session.title)
+    organizer = vCalAddress('MAILTO:no-reply@hasgeek.com')
```
jace, line 156: Use `f"MAILTO:no-reply@{current_app.config['DEFAULT_DOMAIN']}"`

PR #850, opened 2020-08-27
Process email recipients to pass Flask-Mailman's sanitizer check

iambibhas:


funnel/transports/email.py
```diff
@@ -55,6 +56,37 @@ def jsonld_confirm_action(description, url, title):
     }
 
 
+def sanitize_emailaddr(recipient):
+    formatted = None
```
jace, line 60: Don't need if all pathways create this variable.

funnel/transports/email.py
```diff
@@ -55,6 +56,37 @@ def jsonld_confirm_action(description, url, title):
...
+        formatted = formataddr(recipient.fullname, str(recipient.email))
+    elif isinstance(recipient, tuple):
+        formatted = formataddr(recipient)
+    elif isinstance(recipient, str):
+        formatted = recipient
+    else:
+        raise ValueError(
+            "Not a valid email format. Provide either a User object, or"
+            " a tuple of (realname, email) or"
+            " a preformatted string e.g. '<\"Name\" email>'"
```
jace, line 71: `" a preformatted string with Name <email>"

funnel/transports/email.py
```diff
@@ -55,6 +56,37 @@ def jsonld_confirm_action(description, url, title):
...
+    if not email_address:
+        raise ValueError('No email address to sanitize')
+
+    try:
+        # try to sanitize the address to check
+        sanitize_address((realname, email_address), 'utf-8')
+    except ValueError:
+        # `realname` is too long, call this function again but
+        # truncate realname by 1 character
+        return sanitize_emailaddr((realname[:-1], email_address))
```
jace, line 84: Don't do recursion. Instead:
```python
	while True:
		try:
			result = sanitize_address((realname, email_address), 'utf-8')
			break
		except ValueError:
			realname = realname[:-1]
	return result
```

funnel/transports/email.py
```diff
@@ -55,6 +56,37 @@ def jsonld_confirm_action(description, url, title):
...
+
+    # `realname` and `addr` are valid, return formatted string
+    return formataddr((realname, email_address))
```
jace, line 87: Second return?

jace (changes requested):

PR #852, opened 2020-08-28
Cache just the home page. Fix caching of offline page

vidya-ram:


funnel/templates/logout_browser_data.html.jinja2
```diff
@@ -0,0 +1,17 @@
+{% extends "layout.html.jinja2" %}
+{% block title %}{% trans %}Logout{% endtrans %}{% endblock %}
+
```
jace, line 3: This needs a bit of content explaining that the logout is in progress. We have a similar case in `meta_refresh.html.jinja2`.

jace (changes requested):

funnel/templates/logout_browser_data.html.jinja2
jace: This helps, but I was referring to the body. Users will see a blank page for a second or so. We might as well show "Logging out…" in the middle of the screen for that second.

jace (approved):
This needs a logout message in the body of the page (not just title). Otherwise all okay.

PR #858, opened 2020-08-30
Moar notifications

jace:


funnel/models/moderation.py
```diff
@@ -47,16 +53,24 @@ def get_one(cls, exclude_user=None):
...
 
-Comment.report_spam = _report_comment
+@reopen(Comment)
+class Comment:
+    def report_spam(self, actor):
+        report = CommentModeratorReport.query.filter_by(
+            user=actor, comment=self
+        ).one_or_none()
+        if report is None:
+            report = CommentModeratorReport(user=actor, comment=self)
```
jace, line 73: Make this `CommentModeratorReport.submit(...)` and let that class deal with the deduplication.

funnel/templates/notifications/comment_report_received_email.html.jinja2
```diff
@@ -0,0 +1,12 @@
+{% extends "notifications/layout_email.html.jinja2" %}
+{% block content -%}
+
+<p>{% trans actor=view.actor.fullname %}Hi {{ actor }},{% endtrans %}</p>
+
+<p>
+    {% trans url=url_for('siteadmin_review_comment', report=view.report.uuid_b58) -%}
+    A comment has been reported as spam. <a href="{{ url }}">Please review</a>.
+    {%- endtrans %}
+</p>
```
jace, line 10: This is a use case for adding the actions layer to notifications. I'll spec it out along with the Telegram transport, but we could do a couple quick fixes here:

1. Include a copy of the comment. (This may be a bad idea if it triggers spam detection in the email client.)

2. Use a button bar in the template. `<p class="btn-bar"><a class="btn">`

funnel/views/commentvote_notification.py
```diff
@@ -0,0 +1,28 @@
...
+
+    reason = __("You are receiving this because you are a site admin.")
+
+    def web(self):
+        return render_template(
+            'notifications/comment_report_received_web.html.jinja2', view=self
+        )
+
+    def email_subject(self):
+        return _("Comment reported as spam")
```
jace, line 23: Following convention of emojis in all email, maybe 💩

funnel/views/commentvote_notification.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def email_content(self):
+        return render_template(
+            'notifications/comment_report_received_email.html.jinja2', view=self
+        )
```
jace, line 28: SMS template is needed.

funnel/models/notification_types.py
```diff
@@ -134,30 +129,45 @@ class ProjectStartingNotification(ProfileIsParent, Notification):
...
+    category = notification_categories.participant
     title = __("When my proposal receives a comment")
     exclude_actor = True
 
     document_model = Proposal
     fragment_model = Comment
     # Note: These roles must be available on Comment, not Proposal. Roles come from
     # fragment if present, document if not.
-    roles = ['presenter', 'replied_to_commenter']
+    roles = ['presenter', 'replied_to_commenter']  # FIXME: Role may still be 'speaker'
```
iambibhas, line 161: `presenter` seems right https://github.com/hasgeek/funnel/blob/d2111a9efde8d6ced07205f761dc0f5f407a5b53/funnel/models/proposal_membership.py#L98-L108

PR #862, opened 2020-09-02
UI fixes

vidya-ram:
1. Use platform-native share(web api)
2. Fix bug of lazy loading on notification feed page. 
When the site updates page is loaded on a large screen, only the first page of notifications are shown. Second page is not loaded as there is no scroll activity, but a spinner appears permanently
3. Fix the padding of sub pages of project.
4. Change notification preferences choice to wrap on mobile
5. Fix the size and position of home icon on mobile and desktop
6. Change calendar strip to be full width

### DIFF
--- a/funnel/assets/js/update.js
+++ b/funnel/assets/js/update.js
@@ -12,12 +12,13 @@ const Updates = {
     updateTemplate,
     isEditor,
     headerHeight,
+    addReadMore,
   }) {
     Vue.use(VS2);
 
     const updateUI = Vue.component('update', {
       template: updateTemplate,
-      props: ['update', 'iseditor'],
+      props: ['update', 'iseditor', 'addreadmore'],
       data() {
         return {
           truncated: true,
@@ -71,6 +72,7 @@ const Updates = {
           updates: updates.length > 0 ? updates : '',
           isEditor,
           headerHeight,
+          addReadMore,
         };
       },
       mounted() {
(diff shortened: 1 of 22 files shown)


funnel/static/css/app.css
```diff
@@ -1615,58 +1617,6 @@ a.loginbutton.hidden,
   padding-right: 0;
 }
 
-.tabs {
```
vidya-ram, line 1618: Moved to Basframe

funnel/templates/project.html.jinja2
```diff
@@ -69,6 +69,41 @@
 {% block contenthead %}{% endblock %}
 
 {% block baseheadline %}
+  {% if current_auth.user.is_site_editor %}
```
vidya-ram, line 72: Don't need feature project form in all sub pages of project

funnel/assets/js/update.js
```diff
@@ -12,12 +12,13 @@ const Updates = {
...
     isEditor,
     headerHeight,
+    addReadMore,
   }) {
     Vue.use(VS2);
 
     const updateUI = Vue.component('update', {
       template: updateTemplate,
-      props: ['update', 'iseditor'],
+      props: ['update', 'iseditor', 'addreadmore'],
```
jace, line 21: This isn't case sensitive?

jace (approved):

funnel/assets/js/update.js
vidya-ram: vue template variables(props) don't allow camel case

PR #881, opened 2020-09-08
Comment report resolution no longer delete reports

iambibhas:


funnel/models/moderation.py
```diff
@@ -36,9 +36,17 @@ class CommentModeratorReport(UuidMixin, BaseMixin, db.Model):
     reported_at = db.Column(
         db.TIMESTAMP(timezone=True), default=db.func.utcnow(), nullable=False
     )
+    resolved_at = db.Column(db.TIMESTAMP(timezone=True), nullable=True)
```
jace, line 39: Needs `index=True`

funnel/models/site_membership.py
```diff
@@ -95,3 +95,9 @@ class User:
         lambda self: self.active_site_membership
         and 'site_editor' in self.active_site_membership.offered_roles
     )
+
+    # site_admin means user has one or more of above roles
+    is_site_admin = property(
+        lambda self: self.active_site_membership
+        and 'site_admin' in self.active_site_membership.offered_roles
+    )
```
jace, line 103: All of these need to be rewritten as regular functions

migrations/versions/98d5b16208b8_comment_moderator_report_resolved_at.py
```diff
@@ -0,0 +1,27 @@
...
+down_revision = '2d2797b91909'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        'comment_moderator_report',
+        sa.Column('resolved_at', sa.TIMESTAMP(timezone=True), nullable=True),
+    )
```
jace, line 23: Index needed here too.

jace (changes requested):

PR #894, opened 2020-09-15
Sms cleanup

anandvenkatanarayanan:
This PR adds the following: 
1. Introduces Twilio SDK into the code line for sending SMS and Validating incoming webhooks. 
2. Unit Tests using Twilio SDK for sending SMS via Encrypted Test SID, ACCOUNT and FROM Keys. 

Validating incoming message notifications has not been done as it is turning out to be very difficult and has to be tested in production only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ env:
   global:
     - PGVER=12
     - PGPORT=5433
+    - secure: VSk63d0fSpVr5HNKORE9QJ01BoRkE4PyiADMnO6n7ka0TULzeIyCoPmwNlwaSPi3UounssdLUsR9SOPUwg8FLPBiYoHoTqxaL2y6dVJcP7F1uW8ofJ3M3+edOHfjY/txkktQ36os0pXXFukSzVDajA4J/vZ2A9Pj8nnqmF5siJc=
+    - secure: bi2i66oahTdm00psMe6FuTRVmTubcqZms1nm2UUrllLhALRfJDcT7boBsIkM/pSEHCI76yVVHCQxAL9ouEu0kBlCV9aCCPh0MAAGSVn+LE7ru0U76C9Yoivok5wDJpXo+zUo+RPYdn/VGlY6XI1nAZgur3ZjnkkgUp8dKhcNoHw=
+    - secure: ZmRtFNNRZkk1kOkPCV5jmMuXnestL8tyVA9Wk3TPCIqYsRC1Cgb21aDNlrWOyPuLb2OvGGy2DRlQVLDsHaNTyP0dgYNdoUmr2QEMqmZmrvJAmD6Qw4ibpe5e7hHDhtomDwrtoPeny3JpwWo9EXWm0LLYFfKeQI2uBKkZD603uvY=
 services:
   - redis-server
   - postgresql
@@ -28,11 +31,11 @@ install:
 before_script:
   - sudo -- sh -c "echo '127.0.0.1  funnel.travis.local' >> /etc/hosts"
   - psql -c 'create database funnel_testing;' -U postgres
-  - ./manage.py dbconfig | sudo -u postgres psql funnel_testing
+  - './manage.py dbconfig | sudo -u postgres psql funnel_testing'
   - python -c "import six; print('Six:', six.__version__)"
 script:
-  - ./runtests.sh
-  - ./runfrontendtests.sh
+  - './runtests.sh'
+  - './runfrontendtests.sh'
 after_success:
   - coveralls
 notifications:
--- a/funnel/transports/sms.py
+++ b/funnel/transports/sms.py
@@ -1,5 +1,13 @@
 """Support functions for sending a short text message."""
 
+from dataclasses import dataclass, field
+from typing import Optional
+
+from flask import url_for
+
+from dataclasses_json import config, dataclass_json
+from twilio.base.exceptions import TwilioRestException
+from twilio.rest import Client
 import requests
 
 from .. import app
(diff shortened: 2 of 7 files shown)


funnel/transports/sms.py
```diff
@@ -9,65 +17,131 @@
...
+        sender = app.config['SMS_TWILIO_FROM']
+
+        # Send (This uses the routing API to deliver SMS via a Low Latency Location).
+        # See https://www.twilio.com/docs/global-infrastructure/edge-locations
+        client = Client(account, token)
+
+        # We need a callback url, if it was not generated already.
+        if (not self.twilio_callback) and self.callback:
+            with app.app_context():
+                self.twilio_callback = url_for('process_twilio_event', _external=True)
```
jace, line 126: Don't use `with app.app_context()` when you already have a context (such as when this is called from a view). Database sessions are linked to app context, so creating one will invalidate the existing database session and all ORM objects loaded from it, plus the ongoing DB transaction. The calling function will be unable to use the database after this.

Always use `with app.app_context()` as the _outermost_ construct. We learnt this the hard way a few months ago, resulting in the refactoring in ff99ff22c8a7e7cb4b524a4f45d3482f7ccd7efa.

jace: You can use `app_url_for` here, but it's not needed. The problem is in the tests, not here.

.travis.yml
```diff
@@ -12,6 +12,9 @@ env:
   global:
     - PGVER=12
     - PGPORT=5433
+    - secure: VSk63d0fSpVr5HNKORE9QJ01BoRkE4PyiADMnO6n7ka0TULzeIyCoPmwNlwaSPi3UounssdLUsR9SOPUwg8FLPBiYoHoTqxaL2y6dVJcP7F1uW8ofJ3M3+edOHfjY/txkktQ36os0pXXFukSzVDajA4J/vZ2A9Pj8nnqmF5siJc=
+    - secure: bi2i66oahTdm00psMe6FuTRVmTubcqZms1nm2UUrllLhALRfJDcT7boBsIkM/pSEHCI76yVVHCQxAL9ouEu0kBlCV9aCCPh0MAAGSVn+LE7ru0U76C9Yoivok5wDJpXo+zUo+RPYdn/VGlY6XI1nAZgur3ZjnkkgUp8dKhcNoHw=
+    - secure: ZmRtFNNRZkk1kOkPCV5jmMuXnestL8tyVA9Wk3TPCIqYsRC1Cgb21aDNlrWOyPuLb2OvGGy2DRlQVLDsHaNTyP0dgYNdoUmr2QEMqmZmrvJAmD6Qw4ibpe5e7hHDhtomDwrtoPeny3JpwWo9EXWm0LLYFfKeQI2uBKkZD603uvY=
```
jace, line 17: These three need a comment with the name of the environment variable, for future maintainability. If the secrets change, we need to know which line to replace.

funnel/transports/sms.py
```diff
@@ -9,65 +17,131 @@
...
+    msg_status: Optional[str] = field(metadata=config(field_name='MessageStatus'))
+    to: str = field(metadata=config(field_name='To'))
+    msg_sid: Optional[str] = field(metadata=config(field_name='MessageSid'))
+    account_sid: str = field(metadata=config(field_name='AccountSid'))
+    sender: str = field(metadata=config(field_name='From'))
+    api_version: Optional[str] = field(metadata=config(field_name='ApiVersion'))
+
+
+class SmsSender:
+    """ Manages Sending SMS to different providers. """
```
jace, line 51: Trailing spaces (both sides)

funnel/transports/sms.py
```diff
@@ -9,65 +17,131 @@
...
-        raise TransportTransactionError("Twilio API error", r.status_code, r.text)
-    except requests.ConnectionError:
-        raise TransportConnectionError("Twilio not reachable")
+
+    @staticmethod
+    def _route_via_exotel(phone_number: str, message: str) -> str:
+        """
+        Route the SMS via Twilio. This is done only for messages that can't be delivered by exotel.
+        :param phone_number: Phone Number
+        :param message: Message to deliver to Phone number.
```
jace, line 78: Sphinx (documentation generator) supports type annotations in `param` tags. Syntax: `:param type name: doc`

We don't use Sphinx to document Funnel, but we should. Also, this _may_ be unnecessary if Sphinx reads type annotations from the function signature itself. I haven't checked for this since moving to Py3.

jace: Confirmed, Sphinx gets it from annotations. This fix is not necessary. https://stackoverflow.com/questions/34647966/how-to-express-multiple-types-for-a-single-parameter-or-a-return-value-in-docstr

funnel/views/sms_events.py
```diff
@@ -0,0 +1,89 @@
...
+            self.url = url_for('process_twilio_event', _external=True)
+            self.validator = RequestValidator(app.config['SMS_TWILIO_TOKEN'])
+
+        # Let us now validate and reject if signatures mismatch. See this:
+        # https://www.twilio.com/docs/usage/security#validating-requests
+        return self.validator.validate(self.url, message, signature)
+
+
+# Unexposed global validator
+twilio_validator = TwilioMessageValidator()
```
jace, line 40: This singleton is going to trip someone in future. Classes are expected to be mutable, so someone is going to write to `self` some day and not notice anything wrong when testing in a single-threaded environment. I made this exact mistake with the SNS validator class.

If immutability is desired, use a module with functions instead of a class. Mutation will then require prominent `nonlocal var` or `global var` declarations.

funnel/views/sms_events.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@app.route('/api/1/sms/twilio_event', methods=['POST'])
+@render_with(template=None, json=True)
+def process_twilio_event():
+    """ Processes SMS callback events from Twilio """
+
+    # Register the fact that we got a Twilio SMS event.
+    # If there are too many rejects, then most likely a hack attempt.
+    statsd.incr('sms_message.twilio_event.received')
```
jace, line 50: Let's prefix these with `phone_number.` for the forthcoming `phone_number` table. So `phone_number.sms.twilio_event.received`

funnel/views/sms_events.py
```diff
@@ -0,0 +1,89 @@
...
+        return {'status': 'error', 'error': 'missing_signature'}, 400
+
+    # Get the JSON message
+    message = request.get_json(force=True, silent=True)
+    if not message:
+        statsd.incr('sms_address.twilio_event.rejected')
+        return {'status': 'error', 'error': 'not_json'}, 400
+
+    # Needs conversion to Dict
+    payload = json.loads(message)
```
jace, line 65: `message` is already a dict because of the `request.get_json()` call above.

funnel/views/sms_events.py
```diff
@@ -0,0 +1,89 @@
...
+    # Check if we find twilio headers and if not reject it
+    signature = request.headers.get('X-Twilio-Signature')
+    if not signature:
+        statsd.incr('sms_message.twilio_event.rejected')
+        return {'status': 'error', 'error': 'missing_signature'}, 400
+
+    # Get the JSON message
+    message = request.get_json(force=True, silent=True)
+    if not message:
+        statsd.incr('sms_address.twilio_event.rejected')
```
jace, line 61: `sms_message` or `sms_address`? Best to consolidate into `phone_number.sms`

funnel/views/sms_events.py
```diff
@@ -0,0 +1,89 @@
...
+    if sms_response.msg_status == 'queued':
+        sms_db_model.status = SMS_STATUS.QUEUED
+    elif sms_response.msg_status == 'failed':
+        sms_db_model.status = SMS_STATUS.FAILED
+    elif sms_response.msg_status == 'delivered':
+        sms_db_model.status = SMS_STATUS.DELIVERED
+    elif sms_response.msg_status == 'sent':
+        sms_db_model.status = SMS_STATUS.PENDING
+    else:
+        sms_db_model.status = SMS_STATUS.UNKNOWN
```
jace, line 84: Prefix this with a `# FIXME` as this is a temp-replacement for the missing PhoneNumber model. SMSMessage is meant to store outgoing messages.

tests/unit/transports/test_sms.py
```diff
@@ -0,0 +1,104 @@
...
+    TWILIO_INVALID_TARGET = "+15005550001"
+    TWILIO_CANT_ROUTE = "+15005550002"
+    TWILIO_NO_SMS_SERVICE = "+15005550009"
+
+    # Dummy Message
+    MESSAGE = "Test Message"
+
+    def test_twilio_success(self):
+        """ Test if Message sending is a success. """
+        sender = SmsSender(callback=False)
```
jace, line 24: This `callback` parameter is not used in the singleton. Also, since it's a singleton, you must test with the existing singleton for tests to match runtime expectations

tests/unit/transports/test_sms.py
```diff
@@ -0,0 +1,104 @@
...
+    # https://www.twilio.com/docs/iam/test-credentials
+    TWILIO_CLEAN_TARGET = "+15005550010"
+    TWILIO_INVALID_TARGET = "+15005550001"
+    TWILIO_CANT_ROUTE = "+15005550002"
+    TWILIO_NO_SMS_SERVICE = "+15005550009"
+
+    # Dummy Message
+    MESSAGE = "Test Message"
+
+    def test_twilio_success(self):
```
jace, line 22: These tests must take `test_client` as parameter, and will get `app.app_context()` automatically applied. You don't have to use `test_client` within the test itself.

tests/unit/transports/test_sms.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Target Numbers (Test Only). See this
+    # https://www.twilio.com/docs/iam/test-credentials
+    TWILIO_CLEAN_TARGET = "+15005550010"
+    TWILIO_INVALID_TARGET = "+15005550001"
+    TWILIO_CANT_ROUTE = "+15005550002"
+    TWILIO_NO_SMS_SERVICE = "+15005550009"
+
+    # Dummy Message
+    MESSAGE = "Test Message"
```
jace, line 20: The class is unnecessary. Make these module-level constants.

jace (changes requested):

funnel/transports/base.py
```diff
@@ -43,4 +45,16 @@ def init():
     ):
         platform_transports['sms'] = True
 
+    # We need to set Callback URL for SMS notifications here as there are two different providers
+    # (Exotel and Twilio) and both of them need different URLs as the schema is very different.
+    with app.app_context():
+        app.config['SMS_TWILIO_CALLBACK'] = url_for(
+            'process_twilio_event', _external=True, _method="POST"
```
jace, line 52: Our string quote convention is single quotes for identifiers and double quotes for human-facing text. In this case, it should be `'POST'` in single quotes.

funnel/transports/sms.py
```diff
@@ -9,10 +15,60 @@
...
+    """
+    if phone_number.startswith('+91'):
+        return _route_via_exotel(phone_number, message)
+    elif phone_number.startswith('+'):
+        return _route_via_twilio(phone_number, message, callback)
+    raise TransportRecipientError("No service provider available for this recipient")
 
 
-def send_via_exotel(phone_number, message):
+def _route_via_exotel(phone_number: str, message: str) -> str:
```
jace, line 64: It's okay to expose these for external use, in case a specific carrier is desired. The older names `send_via_exotel` and `send_via_twilio` work for that.

funnel/transports/sms.py
```diff
@@ -9,10 +15,60 @@
...
+    """
+    Route the SMS via Twilio. This is done only for messages that can't be delivered by exotel.
```
jace, line 66: Docstring to be updated.

funnel/transports/sms.py
```diff
@@ -38,36 +94,37 @@ def send_via_exotel(phone_number, message):
...
-            },
-        )
-        if r.status_code in (200, 201):
-            # All good
-            jsonresponse = r.json()
-            return jsonresponse.get('sid')
-        raise TransportTransactionError("Twilio API error", r.status_code, r.text)
-    except requests.ConnectionError:
-        raise TransportConnectionError("Twilio not reachable")
+def _route_via_twilio(phone: str, message: str, callback: bool = False) -> str:
```
jace, line 97: Callback is necessary for tracking deliverability, so it shouldn't be default False

jace (changes requested):

funnel/views/sms_events.py
```diff
@@ -0,0 +1,125 @@
...
+@render_with(template=None, json=True)
+def process_exotel_event(secret_token: str):
+    """Process SMS callback event from Exotel."""
+
+    # Register the fact that we got a Exotel SMS event.
+    # If there are too many rejects, then most likely a hack attempt.
+    statsd.incr('phone_number.sms.exotel_event.received')
+
+    # We just need to verify the token first.
+    if not validate_exotel_token(secret_token):
```
jace, line 81: Should be `validate_exotel_token(secret_token, request.form.get('To'))`. Use `form.get('To')` instead of `form['To']` because the presence of the field has not been validated yet.

tests/unit/transports/test_sms.py
```diff
@@ -0,0 +1,95 @@
...
+
+# Target Numbers (Test Only). See this
+# https://www.twilio.com/docs/iam/test-credentials
+TWILIO_CLEAN_TARGET = "+15005550010"
+TWILIO_INVALID_TARGET = "+15005550001"
+TWILIO_CANT_ROUTE = "+15005550002"
+TWILIO_NO_SMS_SERVICE = "+15005550009"
+
+# Exotel Numbers to test. They are just made up numbers.
+EXOTEL_TO = "+919999999999"
```
jace, line 21: We can't do this unless Exotel gives us a test API. Instead, mock the request to fake the API call. Use [requests-mock](https://requests-mock.readthedocs.io/en/latest/pytest.html)

jace: Sample use of requests-mock is in `tests/unit/forms/test_project_forms.py`

jace (changes requested):

funnel/transports/sms.py
```diff
@@ -1,29 +1,76 @@
...
+    :param callback: Whether to request a status callback
+    :return: Transaction id
+    """
     sid = app.config['SMS_EXOTEL_SID']
     token = app.config['SMS_EXOTEL_TOKEN']
+    payload = {'From': app.config['SMS_EXOTEL_FROM'], 'To': phone, 'Body': message}
+    if callback:
+        nonce = make_exotel_token(str(uuid.uuid1()))
+        payload['StatusCallback'] = url_for(
+            'process_exotel_event', _external=True, _method='POST', secret_token=nonce
```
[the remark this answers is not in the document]
jace: Py3 does not seem to optimize. You can see it does create a new name in the namespace, even for single use variables:

```python
>>> import dis
>>> code = '''
... import os
... a = os.urandom(1)
... print(a)
... '''
>>> dis.dis(code)
  2           0 LOAD_CONST               0 (0)
              2 LOAD_CONST               1 (None)
              4 IMPORT_NAME              0 (os)
              6 STORE_NAME               0 (os)

  3           8 LOAD_NAME                0 (os)
             10 LOAD_METHOD              1 (urandom)
             12 LOAD_CONST               2 (1)
             14 CALL_METHOD              1
             16 STORE_NAME               2 (a)

  4          18 LOAD_NAME                3 (print)
             20 LOAD_NAME                2 (a)
             22 CALL_FUNCTION            1
             24 POP_TOP
             26 LOAD_CONST               1 (None)
             28 RETURN_VALUE
>>> 
```

tests/unit/transports/test_sms.py
anandvenkatanarayanan: A monkey patch of the requests object via mock was sufficient as there was no need to test responses.

funnel/views/sms_events.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+@app.route('/api/1/sms/twilio_event', methods=['POST'])
+@render_with(template=None, json=True)
+def process_twilio_event():
+    """Process SMS callback event from Twilio."""
+
+    # Register the fact that we got a Twilio SMS event.
+    # If there are too many rejects, then most likely a hack attempt.
+    statsd.incr('phone_number.sms.twilio_event.received')
```
jace, line 20: We don't need to record these events, because all endpoints are logged by default (in Baseframe).

jace (changes requested):

funnel/views/sms_events.py
anandvenkatanarayanan: We did the same in Email also. The idea is to use Kibana's dashboard and group these events together than trying to cobble it from various other sources for correlating them.

PR #905, opened 2020-09-21
Added undo for proposal cancel

iambibhas:


funnel/models/proposal.py
```diff
@@ -423,6 +423,17 @@ def reject(self):
     def cancel(self):
         pass
 
+    @with_roles(call={'creator'})
+    @state.transition(
+        state.CANCELLED,
+        state.SUBMITTED,
+        title=__("Undo Cancel"),
```
jace, line 430: Sentence case: `Undo cancel`

funnel/models/proposal.py
```diff
@@ -423,6 +423,17 @@ def reject(self):
...
+        message=__("This proposal has been submitted"),
```
jace, line 431: `This proposal's cancellation has been reversed.`

jace (changes requested):

PR #911, opened 2020-09-25
Logging JSON decode error while parsing video API

iambibhas:


funnel/models/video.py
```diff
@@ -1,6 +1,7 @@
 import urllib.parse
 
 from pytz import utc
+from simplejson import JSONDecodeError
```
jace, line 4: `requests` uses simplejson?

jace (changes requested):
This PR needs tests, or at least reference documentation on how `requests` behaves when there is a JSON decode error.

funnel/models/video.py
iambibhas: It uses `simplejson` if available, or else `json`.

jace (approved):

PR #923, opened 2020-10-05
Update email_address.py

qurbat:
Fix for "subsection not found" error that occurs when an incorrect value is passed to `/account/email/<hash>/verify`


jace:
There's a proper fix in #928.

PR #924, opened 2020-10-05
XHR support for requires_sudo

jace:
A page with a `@requires_sudo` decorator may be loaded in an XHR request, which inhibits rendering a password dialog.

1. If the page is loaded as XHR HTML, which is typical with legacy jQuery Form endpoints, it will render a HTML+JS redirect to itself, thereby loading the page full screen.

2. If the page is expecting a JSON response, it will return `{"status": "error", "error": "requires_sudo",  "error_description": "User friendly message"}`. The caller must then redirect the user to the `account_sudo` endpoint, with a `next` parameter that points back to the current page and dialog. This endpoint is at `/account/sudo`.

3. If sudo mode is currently active (authentication was performed in the last 15 minutes), the request is passed through uninterrupted.

The first commit in this draft PR includes this infrastructure. The membership JS however fails to account for an error response when removing a member. @vidya-ram this needs to be fixed in JS.

### DIFF
--- a/funnel/assets/js/util.js
+++ b/funnel/assets/js/util.js
@@ -186,8 +186,13 @@ export const Utils = {
       if (response.status === 500) {
         errorMsg =
           'An internal server error occurred. Our support team has been notified and will investigate.';
+      } else if (
+        response.status === 422 &&
+        response.responseJSON.error === 'requires_sudo'
+      ) {
+        window.location.href = `${window.Hasgeek.config.accountSudo}?next=${window.location.href}`;
       } else {
-        errorMsg = JSON.parse(response.responseText).error_description;
+        errorMsg = response.responseJSON.error_description;
       }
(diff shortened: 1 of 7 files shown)


funnel/assets/js/util.js
```diff
@@ -186,8 +186,13 @@ export const Utils = {
       if (response.status === 500) {
         errorMsg =
           'An internal server error occurred. Our support team has been notified and will investigate.';
+      } else if (
+        response.status === 422 &&
+        response.responseJSON.error === 'requires_sudo'
+      ) {
+        window.location.href = `${window.Hasgeek.config.accountSudo}?next=${window.location.href}`;
```
jace, line 193: Will `next` be properly escaped? It needs to include an encoded fragment identifier so that the dialog can be reloaded on return.

PR #925, opened 2020-10-06
Schedule related fixes 

vidya-ram:
1. Show last session date on past project list. 
2. Scroll to latest date on schedule page. 
3. Don't show room title if empty on schedule page.
4. Fix the room tabs active border colour.

### DIFF
--- a/funnel/assets/js/schedule_view.js
+++ b/funnel/assets/js/schedule_view.js
@@ -21,6 +21,8 @@ const Schedule = {
           height: $(window).height(),
           modalHtml: '',
           headerHeight: '',
+          mobileHeader: schedule.config.mobileHeader,
+          desktopHeader: schedule.config.desktopHeader,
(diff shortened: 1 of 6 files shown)


funnel/assets/js/schedule_view.js
```diff
@@ -207,7 +207,10 @@ const Schedule = {
           } else {
             // Scroll page to schedule table
             Utils.animateScrollTo(
-              $(schedule.config.parentContainer).offset().top
+              $(schedule.config.parentContainer)
+                .find('.schedule__date')
+                .last()
+                .offset().top - this.headerHeight
```
jace, line 213: This should jump to _first_ future date, not last date. Some projects may have future dates added in advance.

jace (changes requested):
One concern with the scroll to next date.

jace (approved):

PR #928, opened 2020-10-07
Drop support for 128-bit blake2b hash in UserEmailClaim

jace:
128-bit hashes were used for about a week in June 2020, during the transition from MD5 to 160-bit Blake2b hashes.

Also checks for ValueError from invalid hashes in all views that accept email_hash.


iambibhas (approved):
👍

PR #934, opened 2020-10-13
Increase translation coverage

qurbat:
Some parts of the funnel codebase include strings and text that currently do not account for i18n. This pull request is meant to increase translation coverage for Python strings in the project, as well as coverage (wherever missing) for `{% trans %}`  tags in Jinja templates.


jace (approved):
Merge this and start a new PR for whatever new turns up.

PR #944, opened 2020-10-20
Add mypy and pydocstyle to pre-commit workflow, with fixes

jace:


funnel/views/email_events.py
```diff
@@ -42,27 +46,29 @@ def _email_address(address: str) -> EmailAddress:
         return email_address
 
     def bounce(self, ses_event: SesEvent) -> None:
-        for bounced in ses_event.bounce.bounced_recipients:
+        ses_event_bounce = cast(SesBounce, ses_event.bounce)
```
jace, line 49: Replace `cast` with an `assert not None`. Casts are risky for type safety as typos referring to the wrong type won't be caught. However, Bandit will complain about use of `assert`, so we must either disable that (globally, per line), or write an `assert_not_none` function that raises `AssertionError` instead of using an `assert` statement.

Since function calls are an expensive runtime solution to a static type safety system, the best option may be to use a plain `assert` and tell Bandit to take a hike.

PR #945, opened 2020-10-21
Typing and validation with mypy for models

jace:


funnel/models/commentvote.py
```diff
@@ -96,7 +100,7 @@ class Vote(NoIdMixin, db.Model):
     votedown = db.Column(db.Boolean, default=False, nullable=False)
 
     @classmethod
-    def migrate_user(cls, old_user, new_user):
+    def migrate_user(cls, old_user: User, new_user: User) -> Optional[Iterable[str]]:
```
jace, line 103: `migrate_user` optionally returns a list of tables that have been migrated. This return type here is confusing. It should be defined in a single place (a types module?) as `OptionalMigratedTables`.

PR #949, opened 2020-10-28
Fix scope for image model for imgee test

iambibhas:


tests/unit/models/test_helpers.py
```diff
@@ -146,19 +149,29 @@ def new_foobar(self):
...
 
 
-def test_imgeetype(test_client, test_db):
+@pytest.fixture(scope='session')
+def image_models():
+    class MyImageModel(db.Model):
+        id = db.Column(db.Integer, primary_key=True)  # NOQA: A003
+        image_url = db.Column(ImgeeType)
+
+    db.create_all()
```
jace, line 158: This should not be necessary as `test_db_structure` makes the came call. Try placing the fixture before `test_db` when using it.

PR #952, opened 2020-10-29
Translate JS strings

vidya-ram:


funnel/assets/Makefile
```diff
@@ -6,3 +6,12 @@ assets:
...
 assetsonly:
 	npm run build
+
+source_dir = ../translations
+target_dir = ../static/translations
+baseframe_dir := $(shell python ../../manage.py baseframe_translations_path)
+
+babel:
+	@mkdir -p $(target_dir)
+	ls $(source_dir) | egrep '[[:lower:]]{2}_[[:upper:]]{2}' | xargs -I % sh -c 'mkdir -p $(target_dir)/% && ./node_modules/.bin/po2json --pretty $(source_dir)/%/LC_MESSAGES/messages.po $(target_dir)/%/messages.json'
```
jace, line 16: macOS has BSD `grep` by default, with different regex syntax. Is this why you're using `egrep` here?

funnel/assets/js/app.js
```diff
@@ -1,8 +1,12 @@
 /* global jstz, Pace */
 
 import { Utils, ScrollActiveMenu, LazyloadImg } from './util';
+import i18n from 'gettext.js';
 
 $(() => {
+  window.Hasgeek.config.availableLanguages = {
+    hi: 'hi_IN',
```
jace, line 8: Also an explicit `hi_IN: 'hi_IN'` in case the browser gives you that.

funnel/assets/js/util.js
```diff
@@ -328,6 +328,36 @@ export const Utils = {
...
+    window.i18n = window.i18n();
+    let lang =
+      window.navigator.userLanguage ||
+      window.navigator.language ||
+      window.navigator.browserLanguage ||
+      window.navigator.userLanguage ||
+      window.navigator.systemLanguage;
+    let langShortForm = lang.substring(0, 2);
+    let translatedLang =
+      window.Hasgeek.config.availableLanguages[langShortForm];
```
jace, line 342: This is okay for now, when only Hindi is supported, but we will have another one coming soon: English (India) vs English (United States), primarily to change date formatting. This code needs more robust language matching.

funnel/templates/updates.html.jinja2
```diff
@@ -2,6 +2,7 @@
 {% set title_suffix = project.title %}
 {% from "baseframe/forms.html.jinja2" import renderfield, widgetscripts %}
 {%- from "macros.html.jinja2" import project_header, updates_template, useravatar %}
+{%- from "js/update.js" import updates_template %}
```
jace, line 5: Jinja2 JavaScript templates must be named `.js.jinja2`. This template seems to be a mix of HTML and JS, so I suggest splitting it up into separate template files and importing as needed.

funnel/assets/js/util.js
jace: This may be a good solution: https://stackoverflow.com/questions/28483144/i18n-access-locale-resolution-logic-in-javascript

funnel/assets/js/gettext.js
```diff
@@ -101,11 +98,13 @@ class Gettext {
         console.error("Plural forms don't match");
       }
 
+      var msgstr = '';
       if (num <= 1) {
-        return msgidCatalog[1].replace('%d', num);
+        msgstr = sprintf(msgidCatalog[1], { num: num });
       } else {
-        return msgidCatalog[2].replace('%d', num);
+        msgstr = sprintf(msgidCatalog[2], { num: num });
```
vidya-ram, line 105: Isn't this msgidCatalog[0]

iambibhas: `[0]` is `msgid_plural`. Check the comments on the top.

funnel/assets/Makefile
iambibhas: `grep` doesn't seem to support the `{2}` part. Hence used `egrep` here.

babel.cfg
```diff
@@ -1,3 +1,8 @@
 [python: **.py]
-[jinja2: **/templates/**.jinja2]
+[jinja2: **/templates/**.xml.jinja2]
 extensions=jinja2.ext.autoescape,jinja2.ext.with_,jinja2.ext.do
+[jinja2: **/templates/**.html.jinja2]
+extensions=jinja2.ext.autoescape,jinja2.ext.with_,jinja2.ext.do,webassets.ext.jinja2.AssetsExtension
+[javascript: **/templates/js/**.js.jinja2]
```
jace, line 6: The `js/` folder is unnecessary.

funnel/assets/js/util.js
```diff
@@ -166,13 +173,13 @@ export const Utils = {
...
+      // Defined these strings in project_countdown macro in calendar_snippet.js.jinja2
+      let dayText = ['Today', 'Tomorrow', 'Day after', 'In %d days'];
       // Show number of days on the widget only if it is less than 32 days
       if (counting >= 0 && counting < 3) {
-        monthElem.text(dayText[counting]);
+        monthElem.text(window.gettext(dayText[counting]));
       } else if (counting > 2 && counting < 32) {
-        let daysRemainingTxt = `In ${counting} days`;
-        monthElem.text(daysRemainingTxt);
+        monthElem.text(window.gettext(dayText[3], counting));
```
jace, line 182: Looks like this will mean the source texts will not be detected by `make babel` as there is no gettext wrapper on them. Are these strings hardcoded into the `pot` file, or are there copies elsewhere?

funnel/templates/index.html.jinja2
```diff
@@ -1,7 +1,8 @@
 {% extends "layout.html.jinja2" %}
 {% block title %}{{ config['SITE_TITLE'] }}{% endblock %}
 {% from "baseframe/components.html.jinja2" import faicon %}
-{%- from "macros.html.jinja2" import calendarwidget, saveprojectform, featured_section, upcoming_section, open_cfp_section, all_projects_section, past_projects_section %}
+{%- from "macros.html.jinja2" import calendarwidget, saveprojectform, featured_section, upcoming_section, open_cfp_section, all_projects_section %}
+{%- from "js/past_projects.js.jinja2" import past_projects_section %}
```
jace, line 5: These templates can be moved back into the parent folder.

Subfolders have a lookup cost in Jinja2 because Flask blueprints also appear like subfolders (like the `baseframe/components` above). Every time a template is imported from a subfolder, Jinja2 will scan through _both_ blueprints and filesystem folders looking for a match.

funnel/templates/js/calendar_snippet.js.jinja2
```diff
@@ -0,0 +1,8 @@
+{% macro project_countdown() %}
+  {% raw %}
+    <span>{{ gettext('Today') }}</span>
+    <span>{{ gettext('Tomorrow') }}</span>
+    <span>{{ gettext('Day after') }}</span>
+    <span>{{ gettext('In %d days', 3) }}</span>
+  {% endraw %}
+{% endmacro %}
```
jace, line 8: If this template is not used anywhere, it may be accidentally removed in future. I've had a few PRs already this year that removed unused templates that I happened to notice.

jace: Translation-only templates can be kept in a special file named `translations_only.js.jinja2`, with a comment on each string pointing to where it is used.

funnel/templates/js/server_error.js.jinja2
```diff
@@ -0,0 +1,6 @@
+{% macro error_txt() %}
+  {% raw %}
+    <span>{{ gettext('An internal server error occurred. Our support team has been notified and will investigate.') }}</span>
+    <span>{{ gettext('Unable to connect. Please reload and try again.') }}</span>
```
jace, line 4: I think this one should use a link to reload the page. Avoid use of the words "please" and "kindly" -- these are Indianisms. This text reads better as "Unable to connect. Check connection and tap to reload"

funnel/assets/Makefile
jace: `grep -e` or `grep -E` should do that, but the default BSD grep on macOS may not support it.

funnel/assets/js/util.js
jace: Found it in the calendar snippet template.

funnel/templates/index.html.jinja2
jace: (It is possible this extra lookup cost is negligible. Worth measuring.)

babel.cfg
iambibhas: Made is explicit because it's where all the JS templates are.

PR #969, opened 2020-11-25
Don't return Form objects from JSON views

iambibhas:
- [ ] Remove `csrf_form` from the `project_header` macro

### DIFF
--- a/funnel/templates/updates.html.jinja2
+++ b/funnel/templates/updates.html.jinja2
@@ -22,7 +22,7 @@
       <a href="{{ project.url_for() }}" aria-label="{% trans %}Back to the project{% endtrans %}" class="mui--text-dark mobile-nav__icon" data-action="Back to the {{ project.title }}(videos page)">{{ faicon(icon='arrow-left', icon_size='title') }}</a><span class="mui--text-dark mobile-nav__headline">{% trans %}Updates{% endtrans %}</span>
     </div>
   </div>
-  {{ project_header(project, csrf_form, class='mui--hidden-xs mui--hidden-sm', current_page='updates', pinned=false) }}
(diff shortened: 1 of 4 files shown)


funnel/templates/updates.html.jinja2
```diff
@@ -22,7 +22,7 @@
       <a href="{{ project.url_for() }}" aria-label="{% trans %}Back to the project{% endtrans %}" class="mui--text-dark mobile-nav__icon" data-action="Back to the {{ project.title }}(videos page)">{{ faicon(icon='arrow-left', icon_size='title') }}</a><span class="mui--text-dark mobile-nav__headline">{% trans %}Updates{% endtrans %}</span>
     </div>
   </div>
-  {{ project_header(project, csrf_form, class='mui--hidden-xs mui--hidden-sm', current_page='updates', pinned=false) }}
+  {{ project_header(project, current_view.CsrfForm(), class='mui--hidden-xs mui--hidden-sm', current_page='updates', pinned=false) }}
```
jace, line 25: Baseframe already publishes a `csrf_token()` global. This parameter can be removed and replaced with a macro in Baseframe's `components.html.jinja2`. See: https://github.com/hasgeek/baseframe/blob/9be0dbe371854c29dec0b1d9cabf4550928795db/baseframe/templates/baseframe/mui/baseframe.html.jinja2#L24-L27

funnel/views/mixins.py
```diff
@@ -31,6 +31,7 @@ class ProjectViewMixin(object):
     model = Project
     route_model_map = {'profile': 'profile.name', 'project': 'name'}
     SavedProjectForm = SavedProjectForm
+    CsrfForm = forms.Form
```
jace, line 34: This can be removed in favour of the macro.

jace (changes requested):

jace:
Forms are now JSON-renderable, so this PR is no longer necessary.

PR #972, opened 2020-12-02
Moved csrf tags to a separate macro

iambibhas:


jace (approved):
This is fine, but I think `csrf_tag` belongs in Baseframe. It can be reproduced there in `components.html.jinja2` and migrated here in a separate PR.

PR #976, opened 2020-12-03
Tests and fixes for merge_users in memberships and notifications

jace:


funnel/models/membership.py
```diff
@@ -230,8 +248,68 @@ def replace(self, actor: User, **roles: Dict[str, Any]) -> ImmutableMembershipMi
...
+            # If both records are invites or neither is an invite, use existing records
+            this = self
+
+        role_columns = {}
+        for column in this.__data_columns__:
+            column_value = getattr(this, column)
+            if not column_value:
+                # Replace falsy values with value from the other record. This may need
+                # a more robust mechanism in the future if there are multi-value columns
+                column_value = getattr(other, column)
```
iambibhas, line 277: What is `this` has a role that `other` does not? We keep the one from `this`?

jace: For boolean columns, yes. Behaviour is undefined for non-boolean columns. If we create one, we'll need to revise this method.

PR #979, opened 2020-12-08
Simplified Proposal body content (resolves #437)

### LINKED ISSUE #437: Proposal templates and structure
Picking up from #403, proposals currently contain a number of content fields:

- Abstract ("Objective" prior to #436)
- Outline ("Description" prior to #436)
- Bio
- Requirements
- Slides
- ~~Preview video~~ (removed in #572)
- Links

This is unnecessarily verbose and reflects past thinking where workflow was hardcoded. We've moved away from this approach, notably with the implementation of labels in #313. The content of the above fields should be merged into a single Markdown field named Body, with Markdown headlines inserted for the erstwhile columns.

New columns:

* `body_text`/`body_html` (Markdown; contains the full proposal)
* `description` (Text; contains an auto-generated description: the first paragraph of the body with HTML stripped. However the user may optionally edit this description)
* `custom_description` (Boolean; indicates that the user customised the description and it should not be auto-generated when the body is edited)

In addition, project editors should be able to define one or more proposal templates, giving proposers a blueprint to start with. Templates can be in the form of:

1. A `proposal_template` (`_text`/`_html`) column in the Project model which is copied to the `body` column in a new proposal, or
2. Proposals can get a new state named "Template" that is only available to editors. When a proposer starts the proposal submission process, they get the option of picking up a template.

Both these options potentially conflict with #327 (unsubmitted proposals) since they will require drafts to be composed starting from within a project rather than independently.

jace:
This PR replaces the multiple fields used in proposals so far (abstract, outline, etc) with a single "body" field. Existing proposals are migrated by converting the fields to Markdown headers.

This PR _does not_ delete legacy fields from the database as a transition period is prudent. A second PR following this must delete them. Additionally, a decision has to be made on handling existing data for three deprecated fields: email, phone and location.

### DIFF
--- a/funnel/assets/cypress/fixtures/proposal.json
+++ b/funnel/assets/cypress/fixtures/proposal.json
@@ -1,12 +1,7 @@
 {
   "title": "End to end testing",
-  "abstract": "This talk is about e2e testing in frontend using Cypress.",
-  "outline": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. \nUt enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.",
-  "slides": "https://hasgeek.com/about",
+  "content": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. \nUt enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.",
   "preview_video": "https://www.youtube.com/watch?v=SLlSDI-N8iM",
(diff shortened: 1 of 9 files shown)


jace:
@vidya-ram This PR needs a couple more things:

1. Cypress tests have to be updated for the single `body` field.
2. Front-end must retrieve comments from the `/comments` endpoint and not directly from the proposal URL as that is breaking browser cache.

jace:
Additional functionality needed in this PR:

1. Proposal list must exclude templates (`filter_by(template=False)`)
2. Project editors must be able to set the "template" flag
3. Project editors must be able to set the "featured" flag
4. Featured proposals should be listed on the project page
5. "New proposal" GET action must use the template's body, if one only is available
6. If there are multiple templates, "New proposal" must ask the user to choose, and must reflect the choice by using the `new?template=<suuid>` URL

jace:
Maybe template management can be a separate PR as the priority is the use case for testimonials.

PR #992, opened 2020-12-18
Add size parameter to imgee url

vidya-ram:


funnel/templates/macros.html.jinja2
```diff
@@ -125,7 +125,7 @@
         <div class="project-banner__profile-details">
           {%- if project.profile.logo_url.url %}
             <a href="{{ project.profile.url_for() }}" class="project-banner__profile-details__logo-wrapper" data-action="View {{ project.profile.title }}(project page)">
-              <img class="project-banner__profile-details__logo_wrapper__logo" src="{{ project.profile.logo_url }}" alt="{{ project.profile.title }}"/>
+              <img class="project-banner__profile-details__logo_wrapper__logo" src="{{ project.profile.logo_url }}?size=80" alt="{{ project.profile.title }}"/>
```
jace, line 128: Jinja2 calls should use `{{ project.profile.logo_url.resize(80) }}` @iambibhas please confirm the method name and parameters.

This is required because the URL may use Imgee's S3 bucket instead of directly being an Imgee URL. The ImgeeField is smarter about knowing when it can resize and when it can't.

funnel/templates/macros.html.jinja2
```diff
@@ -853,10 +853,17 @@
...
 {%- endmacro %}
 
-{%- macro useravatar(user, css_class='', add_profile_link=true) %}
+{%- macro useravatar(user, css_class='', add_profile_link=true,  size='medium') %}
+  {% if size == 'big' %}
+    {%- set imgsize = '160' %}
+  {%- elif size == 'medium' %}
+    {%- set imgsize = '80' %}
+  {%- else %}
+    {%- set imgsize = '48' %}
```
jace, line 862: These should be integers when using the resize method.

jace (changes requested):

jace (approved):

funnel/templates/macros.html.jinja2
iambibhas: Profile model was not using `ImgeeField`. `resize(width, [height])` is available to all `ImgeeField` instances.

tests/unit/models/test_helpers.py
```diff
@@ -177,5 +177,6 @@ def test_imgeetype(test_client, test_db, image_models):
     test_db.session.commit()
     assert m2.image_url.url == valid_url
     assert m2.image_url.resize(120, 100).url == valid_url + "?size=120x100"
+    assert m2.image_url.resize(120).url == valid_url + "?size=120"
```
jace, line 180: Add a test for a URL that already has a query string in it.

PR #998, opened 2021-01-04
Replaced "refresh" with "reload" in messages

### LINKED ISSUE #995: Replace "Refresh" with "Reload" in all messages
https://github.com/hasgeek/funnel/pull/977#discussion_r549219408

iambibhas:
Fixes #995.


funnel/views/membership.py
```diff
@@ -200,8 +200,8 @@ def edit(self):
                         {
                             'status': 'error',
                             'error_description': _(
-                                "This member’s record was edited elsewhere."
-                                " Please refresh the page"
+                                "This member’s record was edited elsewhere. "
+                                "Reload the page"
```
jace, line 204: I think we should stick with the convention of space on the next line.

1. Trailing space often looks like a typing error.
2. Preceding space makes it clearer that this string is joined to the previous string.

jace (approved):
One small change.

Issue #1014, opened 2021-01-30
Complement updates with promotions


jace:
The Update model is meant for posting updates within a project as new information becomes available – such as a Zoom meeting link or a YouTube video.

However, updates are also frequently used to promote newer projects. This usage is spammy as multiple past projects get the same update content. We need a distinct variant of updates to make this work properly.

1. Instead of posting multiple updates in a past project, post a single update in the new project.
2. Mark this update as a promotion, and select the past projects it should be promoted in.
3. This update will now appear in the notifications for past project participants, but as it is a single update, it will only appear once, without spamming.
4. It will (should?) not appear in the new project's updates.

This feature can be implemented using either a new Promotion data model, or by reusing the existing Update model and adding a flag. The flag-based approach will make implementation and maintenance simpler.

There is a pending problem with authorization: editors can write updates and edit all updates within the same project, but to promote to another project's participants, they will need authorization from an editor there. While one user may be an editor in both places, another editor may not be, so there has to be an explicit authorization token linked to the editor who authorized. This will need a "promotion authorization" model with both "requested" and "granted" flags and corresponding UI.

jace:
As clarified in #1015, promotion authorization requires the `promoter` role (formerly `concierge`), not `editor`. This will mean that updates can now be written by both editors and promoters.

jace:
If the Update model is to be reused for promotions, it will need a distinguishing flag and a support model.

1. `UPDATE_STATE` currently has `DRAFT`, `PUBLISHED` and `DELETED`. It could use an additional `PROMOTED` state to indicate it's published as a promotion. However, a deleted promotion will have no indication of what it was, and a draft can't indicate what it is meant to be. Also, an update can't simultaneously appear in this and other projects. We _may_ instead need a distinct `is_promotion` flag.

2. A new `UpdatePromotion` support model should link an update with other projects. It will need these columns:

    * `update_id` and `project_id` as the join columns
    * `requested_by` and `requested_at` (non-nullable)
    * `granted_by` and `granted_at`
    * `revoked_by` and `revoked_at`
    * `is_active` column property (defined as `granted_at is not None and revoked_at is None`)
    * Note: the existence of an `UpdatePromotion` cannot be used to determine if an update is a promotion as it is an inefficient query

3. A `PromotedUpdateNotification` to show updates to participants in these projects.

    * Document: `Update` (or document is `Update.project` and fragment is `Update` to prevent dupe notifications when the update is re-promoted)
[...]

PR #1038, opened 2021-02-16
Promoted updates, implements #1014

jace:


jace:
This has been abandoned in #1014.

Issue #933, opened 2020-10-09
Account deletion feature


iambibhas:
We discussed it and decided that we won't actually delete any content from the site, but when users opt-in to "delete" their accounts, we'll replace their personally identifiable data like names with `[deleted]`, remove emails and disable their accounts. This ticket lists all the things needed to accomplish this - 

1. Replace user's full name with `[deleted]`
2. release the username, replace with something random
3. delete emails linked to the account (need more detailed steps)
4. disable account so nobody can login with it
5. what else?

jace:
All _personal_ data of a user has to be deleted. This includes:

1. UserEmail/UserEmailClaim
2. UserPhone/UserPhoneClaim
3. Profile (no random usernames)

This will cascade to anything owned by a profile (currently Project), and presents a problem there as projects are collaborative spaces and hold the data of other users. The first cut of user deletion will only work when the profile is empty.

We have a separate "convert user account into organization" ticket for orgs that registered as users. That feature will have to be used to appoint collaborators and release a profile from the user account.

jace:
As a destructive action, this also needs safeguards. One of:

1. Temporary suspension that the user can undo if they change their mind (Twitter has this)
2. Actual delete, but send the user a signed packet of data that can be restored with one upload (somewhat dangerous due to possible conflicts when data is exclusive, like an email address)
3. Gating against hijacks, such as requiring an email or SMS-based confirmation
4. Backend-only Locked state on a user account, preventing deletion attempts

PR #1051, opened 2021-03-19
Hiding comment content if user is suspended or deleted

iambibhas:
> Kiran Jonnalagadda, [17.03.21 11:34]
> @m0ntypy comments view does not exclude comments where the user is suspended. It should mark the user as such.
> 
> Kiran Jonnalagadda, [17.03.21 11:35]
> I've suspended a persistent spammer's account. Comments are still showing. https://hasgeek.com/thegoaproject/design/sub/project-defy-design-education-for-yourself-TUeVDbExzF9DQ3PcfE3jEE#c-PUU72uf7C4Ko8e2nTQHcB6

### DIFF
--- a/funnel/models/user.py
+++ b/funnel/models/user.py
@@ -809,8 +809,12 @@ class DuckTypeUser(RoleMixin):
         self.fullname = self.title = self.pickername = representation
 
     def __str__(self) -> str:
-        """Represent user account as a string."""
-        return self.pickername
+        """
+        Represent user account as a string.
+
+        `str()` because `__str__()` needs to return a string object.
+        """
+        return str(self.pickername)
 
 
 deleted_user = DuckTypeUser(__("[deleted]"))
--- a/funnel/models/commentvote.py
+++ b/funnel/models/commentvote.py
@@ -260,9 +260,9 @@ class Comment(UuidMixin, BaseMixin, db.Model):
     def user(self) -> Union[User, DuckTypeUser]:
         return (
             deleted_user
-            if self.state.DELETED
+            if self.state.DELETED or self._user.state.DELETED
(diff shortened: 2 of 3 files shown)


funnel/models/user.py
```diff
@@ -809,8 +809,12 @@ def __init__(self, representation: str) -> None:
         self.fullname = self.title = self.pickername = representation
 
     def __str__(self) -> str:
-        """Represent user account as a string."""
-        return self.pickername
+        """
+        Represent user account as a string.
+
+        `str()` because `__str__()` needs to return a string object.
```
jace, line 815: When is `pickername` not a str? Also, this should be comment, not docstring.

iambibhas: By default it's `LazyString` because we create model instance like this `deleted_user = DuckTypeUser(__("[deleted]"))`.

iambibhas:
> Kiran Jonnalagadda, [19.03.21 17:08]
> This PR needs a product spec, because its functionality conflicts with the opening comment of this other pending issue: https://github.com/hasgeek/funnel/issues/933
> 
> Kiran Jonnalagadda, [19.03.21 17:13]
> Quick product spec: we need a UserModeratorReport model similar to CommentModeratorReport. The report should be necessary for any account suspension. If the report determines that the user has been suspended for spamming, only then should comments from that user be hidden.
> 
> There is partial support for this feature already in the SiteMembership model's is_user_moderator flag.
> 
> Kiran Jonnalagadda, [19.03.21 17:21]
> The other option is to change the spec in 933 and hide content from suspended and deleted users. This will have a cascading effect on projects and proposals, so we'll have to decide how much to hide.

PR #1067, opened 2021-03-30
Added Project.has_featured_proposals

iambibhas:


funnel/models/proposal.py
```diff
@@ -733,3 +733,11 @@ def proposals_by_confirmation(self):
             .order_by(db.desc('created_at'))
             .all(),
         }
+
+    @property
```
jace, line 737: Make this a column property or a cached property.

tests/unit/models/test_project.py
```diff
@@ -282,3 +282,31 @@ def test_project_rename(
     assert new_redirect is not None
     assert new_redirect == redirect
     assert new_redirect.project == new_project2
+
+
+def test_project_featured_proposal(db_session, user_twoflower, new_project):
```
jace, line 287: Use one of the new project fixtures. Have to deprecate the old fixtures. Also, in this case it would be Rincewind or Ponder Stibbons making a submission. Twoflower is for the user account tests.

jace (changes requested):

tests/unit/models/test_project.py
jace: The fixture could also use Twoflower making a submission to the Ankh Morpork expo (a fixture), because that is the actual story in the first book: Twoflower visits as the very first tourist.

iambibhas: When I'm using any of the expo projects, I get this error - 

`Can't resolve value for column profile.id on object <Profile at 0x10a6088e0>; no value has been set for this column`

Likely because the profile is not committed before being used to create the project - 

https://github.com/hasgeek/funnel/blob/project-featured-proposals/tests/conftest.py#L372:L372

jace: Add `db_session.flush()` in the project fixture's start

jace (approved):

PR #1073, opened 2021-04-05
Add Vega lite support

vidya-ram:


jace (changes requested):
Support for Vega and other embeds is in four places: Profile, Project, Proposal and Session. These should go into multi-use JS and CSS.

Issue #1052, opened 2021-03-19
Surfacing unread updates


jace:
Since the introduction of a notifications feed in #605 and #835, a new problem has emerged. A user may not check their web notifications for a while, run up a backlog, open the feed later and get the top few marked as read, and then have unread notifications buried down the feed. The unread dot will then refuse to go away, but the notifications are not visible.

To fix this, there should be a way to (a) show only unread notifications, a toggle, or (b) unread notifications should appear at the top of the feed. The latter does not introduce new UI elements and may be preferred.

However, the code that paginates the feed items depends on their order being stable, which is currently ensured by using chronological order. If unread items are shown first and then marked as read before the second page is loaded, the order won't be stable.

To get stable ordering, a full-feed snapshot needs to be made before the unread marker is updated. (Instead of full-feed, it could be snapshot that includes all items up to the unread items' dates.) However, this snapshot mechanism itself will be somewhat expensive and wasted if there are no unread updates.

Another option is to eschew the snapshot and borrow a mechanism from Hasjob:

1. Pagination is by timestamp, not page number. The client sends the timestamp of the bottom-most notification and asks for all notifications older than it.
2. The client keeps track of out-of-order items (in this case, unread items that have been marked read; in Hasjob's case, pinned jobs) and asks for them to be excluded from the results.
3. In addition, instead of refreshing the first page every few minutes, the client can also send a request for notifications newer than a stored timestamp and insert them at the top, not touching the rest of the page.

jace:
In point 2 above, the client does not need to send a list for exclusion. It could get a full page back and strip out the items that are already on the page, making de-duplication a client-side concern.

jace:
Another option: have an `/updates/unread` endpoint, that is the same front-end, but only sends unread updates.

jace:
#1077 introduces a separate endpoint for unread updates. It is a temporary measure for users complaining about being unable to clear unread updates. For a proper resolution:

1. We accept this approach, and add it to the UI, or
2. We implement the original idea in this issue, showing unread updates first.

jace:
The unread endpoint is unreliable when there is pagination. By the time the second page loads, the first page will be marked as read, so the page will have shifted.

PR #1077, opened 2021-04-07
Unread updates endpoint (for #1052)

jace:


vidya-ram (approved):
Translations will have to be run for the new page title

PR #1084, opened 2021-04-12
Misc pre-commit related fixes

iambibhas:
Due to several pre-commit hook updates, pre-commit checks have been failing for sometime. These fixes solve that.

### DIFF
--- a/funnel/registry.py
+++ b/funnel/registry.py
@@ -104,7 +104,7 @@ class ResourceRegistry(OrderedDict):
                 try:
                     result = f(authtoken, args, request.files)
                     response = jsonify({'status': 'ok', 'result': result})
-                except Exception as exception:
+                except Exception as exception:  # noqa: B902
                     exception_catchall.send(exception)
                     response = jsonify(
                         {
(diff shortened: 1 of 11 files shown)


funnel/views/venue.py
```diff
@@ -74,11 +74,8 @@ def update_venue_settings(self):
                         room_obj.bgcolor = room['color'].lstrip('#')
                         room_obj.seq = room['seq']
                         db.session.add(room_obj)
-        try:
-            db.session.commit()
-            return {'status': True}
-        except Exception as e:
-            return {'error': str(e)}, 400
+        db.session.commit()
+        return {'status': True}
```
iambibhas, line 78: Don't know what exception was being raised here. Removing it till it throws an exception so that we can add a proper handler.

jace (approved):

funnel/registry.py
```diff
@@ -104,7 +104,7 @@ def decorated_function():
                 try:
                     result = f(authtoken, args, request.files)
                     response = jsonify({'status': 'ok', 'result': result})
-                except Exception as exception:
+                except Exception as exception:  # noqa: B902
```
jace, line 107: We're using `# NOQA` everywhere (upper case) as convention.

PR #1085, opened 2021-04-12
Remove JSON and CSV endpoints (resolves #974)

### LINKED ISSUE #974: The JSON and CSV endpoints on Profile and Project are obsolete and should be removed
They were originally used for the native app and for sync with Eventframe. There is no current usage for them. They return too much data and aren't efficient queries. If they are needed again, (a) the main endpoints now support JSON rendering, and (b) a GraphQL API is more appropriate.

jace:
PR #1065 missed the actual endpoints.


iambibhas (approved):
LGTM

PR #1093, opened 2021-04-14
Added endpoints to remove project/profile images

iambibhas:


funnel/views/profile.py
```diff
@@ -301,14 +311,34 @@ def edit_logo_url(self):
...
+    @route('remove_logo', methods=['POST'])
+    @render_with(json=True)
+    @requires_login
+    @requires_roles({'admin'})
+    def remove_logo(self):
+        form = self.CsrfForm()
+        if form.validate_on_submit():
+            self.obj.logo_url = None
+            db.session.commit()
+            flash(_("The banner image has been removed from this profile"), 'info')
```
jace, line 323: Flash messages should not be used when the outcome is obvious to the user.

funnel/views/profile.py
```diff
@@ -301,14 +311,34 @@ def edit_logo_url(self):
...
+            return render_redirect(self.obj.url_for(), code=303)
+        else:
+            current_app.logger.error(
+                "CSRF form validation error when removing profile logo."
+            )
+            flash(_("There was an issue removing the logo. Try again."), 'error')
```
jace, line 329: Do not use ambiguous statements like "There was an issue". This outcome can happen when the user was tricked into making a submission via XSS, which means they don't know what action was just attempted. The text could be:

"Were you trying to remove the logo? Try again to confirm."

This still needs some tweaking because it contains "you" and the phrase "try again to confirm" sounds wrong.

funnel/views/profile.py
```diff
@@ -334,6 +364,25 @@ def edit_banner_image_url(self):
...
 
+    @route('remove_banner', methods=['POST'])
+    @requires_login
+    @requires_roles({'admin'})
+    def remove_banner(self):
+        form = self.CsrfForm()
+        if form.validate_on_submit():
+            self.obj.banner_image_url = None
+            db.session.commit()
+            flash(_("The banner image has been removed from this profile"), 'info')
```
jace, line 375: Same. Not required.

funnel/views/profile.py
```diff
@@ -334,6 +364,25 @@ def edit_banner_image_url(self):
...
+            return render_redirect(self.obj.url_for(), code=303)
+        else:
+            current_app.logger.error(
+                "CSRF form validation error when removing profile banner."
+            )
+            flash(
+                _("There was an issue removing the banner image. Try again."), 'error'
```
jace, line 382: As above.

funnel/views/project.py
```diff
@@ -421,6 +423,26 @@ def edit_banner(self):
...
+            self.obj.bg_image = None
+            db.session.commit()
+            flash(_("The banner image has been removed"), 'info')
+            return render_redirect(self.obj.url_for(), code=303)
+        else:
+            current_app.logger.error(
+                "CSRF form validation error when removing project banner."
+            )
+            flash(
+                _("There was an issue removing the banner image. Try again."), 'error'
```
jace, line 442: As above.

jace (changes requested):
UI text needs better phrasing or is not required.

funnel/views/profile.py
iambibhas: I've used the sentence you mentioned. Any improvement you can think of? We check for CSRF failure through the codebase. We dont really have a uniform error for them.

funnel/views/profile.py
```diff
@@ -301,14 +311,35 @@ def edit_logo_url(self):
...
+        if form.validate_on_submit():
+            self.obj.logo_url = None
+            db.session.commit()
+            return render_redirect(self.obj.url_for(), code=303)
+        else:
+            current_app.logger.error(
+                "CSRF form validation error when removing profile logo."
+            )
+            flash(
+                _("Were you trying to remove the logo? Try again to confirm."), 'error'
```
jace, line 329: No trailing period in "confirm." (Few more places below.)

jace (approved):
No trailing period in captions. Otherwise okay.

PR #1105, opened 2021-04-21
Profiles have URLs even if not public

jace:


funnel/models/user.py
```diff
@@ -86,11 +86,7 @@ def avatar(self) -> str:
     @with_roles(read={'all'})  # type: ignore[misc]
     @property
     def profile_url(self) -> Optional[str]:
-        # Use a cast because mypy does not recognise that self.has_public_profile
-        # already asserts `self.profile is not None`
-        return (
-            cast(Profile, self.profile).url_for() if self.has_public_profile else None
-        )
+        return Profile.url_for() if self.profile is not None else None
```
iambibhas: What are scenarios where a user/organization may not have a profile?

funnel/models/user.py
```diff
@@ -70,27 +70,26 @@ def validate_name_candidate(self, name: str) -> Optional[str]:
...
     @with_roles(read={'all'})  # type: ignore[misc]
     @property
     def profile_url(self) -> Optional[str]:
-        # Use a cast because mypy does not recognise that self.has_public_profile
-        # already asserts `self.profile is not None`
-        return (
-            cast(Profile, self.profile).url_for() if self.has_public_profile else None
-        )
+        profile = self.profile
+        return profile.url_for() if profile is not None else None
```
iambibhas, line 92: What are scenarios where a user/organization may not have a profile?

jace: Organizations always have a username. Optional for users.

PR #1122, opened 2021-04-30
Switch to proposal memberships

jace:


funnel/static/sass/_project.scss
```diff
@@ -169,7 +169,7 @@
       position: absolute;
       bottom: $mui-grid-padding;
       right: $mui-grid-padding;
-      z-index: 3;
+      z-index: 2;
```
jace, line 172: @vidya-ram is this intended?

PR #1123, opened 2021-04-30
Add  prev and next navigation for Update

vidya-ram:
1. Add `All submission`, `Previous`, `Next` navigation to update page
2. Change all heading font size from `mui--text-display1` to `mui--text-headline`
3. Avoid call getprev() twice, wrap it in a with for proposal page
4. Add cancel button to proposal submission form


funnel/models/update.py
```diff
@@ -310,6 +310,26 @@ def all_published_public(cls) -> Query:
...
+            )
+            .order_by(Update.number.asc())
+            .first()
+        )
+
+    def getprev(self):
+        return (
+            Update.query.filter(
+                Update.project == self.project,
+                Update.number < self.number,
```
jace, line 327: Both queries need an extra condition: `Update.state.PUBLISHED`, like so:

```python
Update.query.filter(
    Update.project == self.project,
    Update.state.PUBLISHED,
    Update.number > self.number
)
```

funnel/templates/update_details.html.jinja2
```diff
@@ -7,11 +7,22 @@
 {% block mobile_header_title %}{% endblock %}
 
 {% block left_col %}
+  <div class="flex-wrapper flex-wrapper--center flex-wrapper--space-between nav-btn-wrapper">
+    <a class="mui--text-dark mui--text-body2 nounderline" href="{{ project.url_for('updates') }}">{{ faicon(icon='chevron-double-left-solid', icon_size='body2', css_class='mui--align-middle btn-margin-right') }}{% trans %}All submissions{% endtrans %}</a>
+    <div>
+      {%- if update.getprev() -%}
+        <a href="{{ update.getprev().url_for() }}" class="mui--text-body2 mui--text-capitalize mui--text-dark nounderline prev-btn">{{ faicon(icon=prev_icon, icon_size='subhead', baseline=false, css_class='mui--align-middle btn-margin-right') }}{% trans %}Previous{% endtrans %}</a>
+      {%- endif -%}
```
jace, line 15: Because the query hits the database, calling `getprev()` twice will hit the database twice. Wrap this in a `with`:

```jinja2
{%- with prev_update = update.getprev() %}{% if prev_update %}
...
```

funnel/views/update.py
```diff
@@ -184,6 +184,22 @@ def delete(self):
...
+        else:
+            return redirect(self.obj.project.url_for())
+
+    @route('prev')
+    def prev(self):
+        prevobj = self.obj.getprev()
+        if prevobj:
+            return redirect(prevobj.url_for())
+        else:
+            return redirect(self.obj.project.url_for())
```
jace, line 201: This routes are not required, as we're rendering with the actual URL of the next and previous update.

jace (changes requested):

jace (approved):

PR #1129, opened 2021-05-05
Remove trailing periods from UI strings

jace:


funnel/templates/formlayout.html.jinja2
```diff
@@ -44,7 +44,7 @@
         var waitingForResponse = false;
         var lastSavedData = '';
 
-        $('input[name="form.revision"]').val() ? $('#autosave-msg').text('These changes have not been published yet.') : '';
+        $('input[name="form.revision"]').val() ? $('#autosave-msg').text(gettext('These changes have not been published yet')) : '';
```
jace, line 47: @vidya-ram Does this line work?

funnel/templates/ticket_event.html.jinja2
```diff
@@ -114,7 +114,7 @@
       if ($('#badge_form select.field-badge_printed').val() === '') {
         return false;
       } else {
-        return window.confirm('Are you sure? Selected action will apply to all listed participants.');
+        return window.confirm(gettext('Are you sure? Selected action will apply to all listed participants'));
```
jace, line 117: @vidya-ram One more.

jace (commented):
@vidya-ram A couple `gettext` calls need your attention.

1. Is the `gettext` function available in these templates?
2. These calls are not getting picked up by Babel because the templates do not have `.js.jinja2` extensions. Where should these calls be relocated for translations to work?

PR #1133, opened 2021-05-07
Add dropdown menu for account

vidya-ram:


jace (changes requested):
Backend changes are inefficient and need to be rewritten.

jace (changes requested):

PR #1137, opened 2021-05-12
Remove namespace from AuthClient; deprecate client notifications

jace:
Also remove code that processes resources, as the supporting models were removed in commit 4b3804507e26cc41042bf7da12a641f15b682ad7 as part of hasgeek/lastuser#382 when merging Lastuser info Funnel.


jace:
CSRF validation is now required in testing. The OAuth2 integration test fails without it. However, this appears to have broken some Cypress tests, and they need to be fixed.

PR #1140, opened 2021-05-13
Added github action for backend tests

iambibhas:
Tried adding a github action for checking test coverage. But Github actions can only output pass/fail. So while it's possible to check whether the coverage is above or below a given number, it's not possible to check whether the coverage has gone up or down compared to what it was. It can't replace what coverall used to do. So renamed this one to backend test action. So we can technically drop travis from running backend tests now. We have to check coverage manually.


jace:
We now have working backend tests via #1460, so closing this.

PR #1151, opened 2021-05-20
Fix project form and follow text

jace:
Date validators are fixed, but it turns out autosave is broken after a server-side form validation failure. This was not noticed until now because all prior validators had client-side versions. The datetime fields do not have client-side validators.

Current status: autosave does not run at all after the form is rendered in a POST request.


jace:
In addition, `formlayout.html.jinja2` has text in embedded JS without `gettext()` wrappers.

PR #1188, opened 2021-06-13
Shortlink implementation (resolves #1174)

### LINKED ISSUE #1174: Short URL generator
A short URL generator was added in #843 for SMS unsubscribe links. However, there is a recurring requirement for short links elsewhere, with various desirable properties, with the option of picking a commercial service or building a custom implementation.

A subset of these characteristics will be easy to build. The full set will raise complexity considerably:

* [x] SMS notification links, where character length matters. Fewer is better.
    * A short domain is required. At the moment the shortest we have is `has.gy`, which has the nice property of referencing the brand name. A shorter one will be better and could be an option in the future, subject to domain cost.
    * Minimal character length in the path. 4 chars ideal (as in #843), 5-6 okay too, but beyond that is hard on SMS.
* [x] Unguessable ids, to protect from enumeration and reduce risk of personal information appearing in the long URL. However, this conflicts with the short id expectation.
* [x] Custom ids as an option (default is random ids), for branded shares.
* [x] Don't hand over auto-generated short link (mandatory for SMS) metadata to a third party service where the user cannot consent to such sharing. Our privacy policy will effectively be a superset of that service's policy.
* [ ] Apart from SMS, use in email and webshare, to hide ugly UTM tracking tags. This will require the short URL to work for page preview.
* [ ] Unique URLs per share, to track number of shares and traffic from that share (via utm tags added to the long URL).
* [ ] Analytics on the short URL itself (not deferred to Google Analytics for destination traffic). Track bots vs browsers, open counts, periods of activity, etc. This is effectively tracking every open, like a typical short URL service.
* [ ] Replace all outbound links with short links to get analytics on outbound traffic. This is a debatable use as web ping can be used instead, and is already used by web analytics to track clicks.
* [ ] Automatic short links in the form of `domain.tld/uuid-as-base58` where there is no backing database entry. Instead, this causes a search through tables for a matching document, and generates a URL for it. An additional character can be used to identify the table or share criteria (web, email, etc). This is similar to how YouTube's short URLs at `youtu.be` use the same ids as the regular URL. This has the advantage of being cheaper to construct, for scenarios where the URL may not be used at all.

Based on these expectations, if we do an internal implementation, the short URL table will need the following:

1. Primary key, with options:
   1. Serial integer (usual)
   2. UUID, for distributed generation as the service goes
   3. Random bigint (instead of UUID), which has the advantage of serving as integer representation of the short URL – with the caveat the int-str-int converter must be fast and portable across architectures
   4. String pkey, to contain the short URL itself. Will force the same string to appear as fkey in the analytics table
2. Optional: Custom short URL code, indexed unique. Required only if custom URLs are desired, or if the random bigint option is not suitable
3. Long URL
4. Optional: Canonical Long URL, created by either stripping query parameters (risky) or fetching the page and looking for a `link rel=canonical` tag (slow). Useful for analytics, to measure traffic to the destination across short links
5. Optional: document table and document uuid, similar to the `Notification` model, replacing the long URL and allowing it to be reconstructed. This will make analytics simpler for internal destinations, but add overhead when external links are typical.
6. Optional: user id, tracking which user created the link (for webshare), to identify high value sharers, but at risk to privacy
7. `created_at` (standard), to know when the link was created, presumably closely tied to when it was shared

jace:


tests/conftest.py
```diff
@@ -75,10 +75,6 @@ def restart_savepoint(session, transaction_in):
 @pytest.fixture(autouse=True)
 def client(request, db_session):
     """Provide a test client."""
-    if 'noclient' in request.keywords:
-        # To use this, annotate a test with:
-        # @pytest.mark.noclient
-        return None
```
jace, line 81: This doesn't work because if there is a `yield` statement in a fixture, there can't be a `return` or a second `yield`. This fixture will have to be rewritten without using `yield` to allow it to disable itself.

PR #1196, opened 2021-06-16
Send verify email template for user with unverified email

vidya-ram:


jace:
This PR is no longer necessary as the verification system is changing:

1. Redis-backed OTP instead of a database model. UserPhoneClaim is gone in #1414, UserEmailClaim is going next.
2. UserEmail will get a verified bool (likely via a verified_at datetime) and past instances of UserEmailClaim will be merged into UserEmail.

PR #1200, opened 2021-06-17
Sending notification if a spam report was created, and fixed tests

iambibhas:


instance/testing.py
```diff
@@ -25,7 +25,7 @@
 RECAPTCHA_PUBLIC_KEY = environ.get('RECAPTCHA_PUBLIC_KEY')
 RECAPTCHA_PRIVATE_KEY = environ.get('RECAPTCHA_PRIVATE_KEY')
 RECAPTCHA_OPTIONS = ''
-WTF_CSRF_ENABLED = True
+WTF_CSRF_ENABLED = False
```
jace, line 28: This should be `True`.

jace (changes requested):

jace (approved):

PR #1237, opened 2021-07-14
Update node modules, webpack

jace:


vidya-ram (approved):
Downgraded ractive to 0.8
All cypress tests are passing now. Wasn't able to test ticket sync tickets.

PR #1253, opened 2021-08-04
Exclude suspended users from search

jace:


vidya-ram (approved):

jace:
Marking a profile as private is a lossy action as the original state cannot be restored when the user account is unsuspended.

Instead, the profile should have a column property that reflects the user's active state.

In addition, organizations should also have active/suspended states for use by moderators.

PR #1276, opened 2021-09-22
Support to add collaborators to submission

vidya-ram:


funnel/static/js/scripts.js
```diff
@@ -39,7 +39,7 @@ window.Hasgeek.Config = {
 
 function activate_widgets() {
   // Activate codemirror on all textareas with class='markdown'
-  $('textarea.markdown').each(function () {
+  $('textarea.markdown:not([style*="display: none"]').each(function () {
```
vidya-ram, line 42: activate widget fn is called again on new collaborator form in the modal. This is to avoid reinitialising codemirror enabled textarea (description field in the form)

funnel/static/js/scripts.js
```diff
@@ -67,6 +67,10 @@ function activate_widgets() {
       }, 300);
     });
   });
+
+  $(document).on('select2:open', function () {
+    document.querySelector('.select2-search__field').focus();
```
vidya-ram, line 72: Upgrade to jquery 3.6 select2 autofocus isn't working. This is to fix that problem.
https://github.com/select2/select2/issues/5993

funnel/assets/js/labels.js
```diff
@@ -19,37 +21,9 @@ $(() => {
...
-            }
-          } else {
-            errorMsg = window.Hasgeek.Config.errorMsg.networkError;
-          }
-          window.toastr.error(errorMsg);
-        },
-      });
-    }
-  });
+  const msg = 'Are you sure you want to %s?';
```
jace, line 24: This needs gettext

jace: Also, translation variables must be for nouns, not phrases.

funnel/assets/js/submission_form.js
```diff
@@ -98,4 +114,10 @@ $(() => {
         removeLineBreaks($(event.currentTarget).val())
       );
     });
+
+  const msg = 'Are you sure you want to remove %s?';
```
jace, line 118: gettext

funnel/static/js/scripts.js
jace: This should be a comment in the code, for anyone who has to maintain this against a future jQuery/Select2 release.

jace (approved):
Minor issues

PR #1298, opened 2021-11-29
Additional config for Docker and AWS hosting

jace:


jace:
@vivekdurai The dev setup in Docker is incomplete because of a peculiarity in how our workflow has developed over the years. The funnel repo is not the root folder, but is instead a subfolder in a larger workspace:

* `hasgeek` – workspace folder
    * `.venv` – virtual environment used for all repos within the `hasgeek` folder
    * `funnel` – this repo
    * `hasjob` – our other project, independent from funnel but sharing the workspace
    * `coaster` (lib) – contains a framework we built on top of Flask
    * `baseframe` (lib) – contains shared UI elements between `funnel` and `hasjob`
    * `flask-babelhg` (lib) – supports translations scattered across multiple repos (libs + app)
    * `flask-lastuser` (lib) – auth management library used by `hasjob` to acquire a user from `funnel`

The libraries – `coaster`, `baseframe` and `flask-lastuser` are often modified in sync with the client apps that use them, and changes are committed only after the functionality works as expected in the client app. To make this work, we use `python setup.py develop` within the venv so that the venv refers directly to the repository.

We need the equivalent of this `python setup.py develop` inside Docker.

jace:
@vivekdurai issue 2: having postgres inside the container in dev will mean that the database is destroyed when the image is rebuilt -- even for dependency upgrades. Having to dump and restore db contents is not optimal. Postgres should be contained for tests but not dev.

fabfile.py
```diff
@@ -0,0 +1,80 @@
+"""Fabric deploy script."""
```
jace, line 1: I think we can delete this file as we're not using Fabric at this time.

instance/testing.py
```diff
@@ -8,9 +8,9 @@
 SECRET_KEYS = ['testkey']  # nosec
 LASTUSER_SECRET_KEYS = ['testkey']  # nosec
 SITE_TITLE = 'Hasgeek'
-SQLALCHEMY_DATABASE_URI = 'postgresql://localhost/funnel_testing'
+SQLALCHEMY_DATABASE_URI = 'postgresql://postgres@postgres:5432/funnel_testing'
 SQLALCHEMY_BINDS = {
-    'geoname': 'postgresql://localhost/geoname_testing',
+    'geoname': 'postgresql://postgres@postgres:5432/geoname_testing',
```
jace, line 13: This shouldn't be using the `postgres` superuser account.

instance/testing.py
```diff
@@ -76,9 +76,9 @@
 SMS_DLT_ENTITY_ID = environ.get('SMS_DLT_ENTITY_ID')
 
 #: Twilio support for non-indian numbers
-SMS_TWILIO_SID = environ.get('SMS_TWILIO_SID')
-SMS_TWILIO_TOKEN = environ.get('SMS_TWILIO_TOKEN')
-SMS_TWILIO_FROM = environ.get('SMS_TWILIO_FROM')
+# SMS_TWILIO_SID = environ.get('SMS_TWILIO_SID')
+# SMS_TWILIO_TOKEN = environ.get('SMS_TWILIO_TOKEN')
+# SMS_TWILIO_FROM = environ.get('SMS_TWILIO_FROM')
```
jace, line 81: All `environ.get` lines can be deleted, as that happens automatically if the env var is named `FLASK_*`

docker/entrypoints/dev.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+if [ "$(psql -XtA -U postgres -h $DB_HOST funnel -c "select count(*) from information_schema.tables where table_schema = 'public';")" = "0" ]; then
+    flask dbcreate
+    flask db stamp
+else
+    flask db upgrade head
+fi
```
jace, line 8: I'm still uncomfortable with this. It should be explicit, not implicit, as it's risky to do this when in the middle of writing a migration.

miteshashar: So, we just remove the else part. We can always run `docker compose exec -it dev flask db upgrade head` manually.

docker/compose/services.yml
```diff
@@ -0,0 +1,21 @@
...
+    logging:
+      driver: json-file
+      options:
+        max-size: 200k
+        max-file: 10
+    extra_hosts:
+      - 'funnel.test:127.0.0.1'
+      - 'f.test:127.0.0.1'
+    environment:
+      - OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES
```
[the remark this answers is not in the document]
miteshashar: Not even when the container is running in macOS?

miteshashar (approved):

PR #1305, opened 2021-12-20
Use role access for ticket sync

jace:


jace:
This PR does not specify access rights for the `project_usher` role. It’ll need a follow-up PR to restore access.

PR #1306, opened 2021-12-22
Reduce banner image size of spotlight and upcoming cards on homepage

vidya-ram:


funnel/templates/js/search.js.jinja2
```diff
@@ -21,7 +21,7 @@
                     <a href="{{ url }}" aria-label="{{ title }}" class="card card--upcoming clickable-card" data-ga="View project">
                       <div class="card__image-wrapper">
                         {{#if obj.bg_image }}
-                          <img class="card__image" src="{{ obj.bg_image }}?size=1500" alt="{{ title }}"/>
+                          <img class="card__image" src="{{ obj.bg_image }}?size=400" alt="{{ title }}"/>
```
jace, line 24: It's an Imgee field, so it'll accept `{{ obj.bg_image.resize(400) }}`

jace (changes requested):
One small change

funnel/templates/js/search.js.jinja2
vidya-ram: It is ractive JS template and resize call is not working

PR #1332, opened 2022-02-08
Show latest event on schedule page

vidya-ram:


funnel/assets/sass/schedule.scss
```diff
@@ -173,7 +173,7 @@
           overflow: auto;
           .schedule__row__column {
             position: relative;
-            background-image: image-url('schedule-box-background.png');
+            background-image: url('../../img/schedule-box-background.png');
```
jace, line 176: Doesn't Sass have a path in config?

funnel/views/schedule.py
```diff
@@ -109,7 +109,7 @@ def schedule_data(
         else:
             data[day] = {}
     schedule = []
-    for day in sorted(data):
+    for day in sorted(data, reverse=True):
```
jace, line 112: This will show future dates before present day. What we need is to hide past dates until the user asks to see the past.

PR #1339, opened 2022-02-24
Updated registry.py

djamg:
Added a method in LoginProviderRegistry to show the services which have the flag 'at_login' set to 'True' and fixed a minor bug.

### DIFF
--- a/funnel/registry.py
+++ b/funnel/registry.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from collections import OrderedDict
 from functools import wraps
-from typing import List, Optional
+from typing import List, Optional, Tuple
 import re
 
 from flask import Response, abort, jsonify, request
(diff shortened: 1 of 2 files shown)


funnel/registry.py
```diff
@@ -197,7 +204,8 @@ class LoginProvider:
     icon = None
     #: Login form, if required
     form = None
-    #: This service's usernames are typically used for addressing with @username
+    #: This service's usernames are typically
+    # used for addressing with @username
```
jace, line 208: This line needs to use `#:`. The colon is used by Sphinx for documentation of whatever is on the following line, since objects in Python can't have docstrings, only modules, classes and files.

In general, wrap comments as close to 88th column as possible. When we find a comment reformatter, it'll use a similar rule.

jace (changes requested):
One formatting fix and then good to merge.

djamg:
Made changes

PR #1345, opened 2022-03-07
Update the button styling and fix avatar colour in the spa templates

vidya-ram:
1. Don't show button when submission is disabled. Remove strikethrough button style
2. Add button style for disabled button. Use it to show when ticket sales is closed.
3. Add full width get tickets button on project page
4. Add avatar_color_code to enable avatar colour for jinja template


funnel/views/account.py
```diff
@@ -155,6 +156,17 @@ def recent_organization_memberships(
...
 
+@User.views('avatar_color_code', cached_property=True)
+def avatar_color_code(obj):
+    # Return an int from 1 to avatar_color_code from the initials of the given string
+    initial = initials(obj.title)
+    parts = initial.split()
+    string_total = ord(parts[0][0])
+    if len(parts) > 1:
+        string_total += ord(parts[0][1])
+    return string_total % no_avatar_colours or no_avatar_colours
```
jace, line 167: This is slightly more efficient if we reproduce the Baseframe filter here and replace the split+join+split with a single split:

```python
    if obj.title:
        parts = obj.title.split()
        if len(parts) > 1:
            total = ord(parts[0][0]) + ord(parts[-1][0])
        else:
            total = ord(parts[0][0])
    else:
        total = 0
    return total % no_avatar_colours or no_avatar_colours
```

funnel/views/helpers.py
```diff
@@ -42,6 +42,9 @@
 
 nocache_expires = utc.localize(datetime(1990, 1, 1))
 
+# Six avatar colours defined in _variable.scss
+no_avatar_colours = 6
```
jace, line 46: In the code above I kept reading it as "no" instead of "number". This variable should be named "avatar_color_count" (using American English for variable names).

funnel/views/organization.py
```diff
@@ -49,6 +51,17 @@ def people_and_teams(obj):
...
+def avatar_color_code(obj):
+    # Return an int from 1 to avatar_color_code from the initials of the given string
+    initial = initials(obj.title)
+    parts = initial.split()
+    string_total = ord(parts[0][0])
+    if len(parts) > 1:
+        string_total += ord(parts[0][1])
+    return string_total % no_avatar_colours or no_avatar_colours
+
+
```
jace, line 64: This is not required as the version in account.py with multiple decorators handles it.

funnel/views/profile.py
```diff
@@ -66,6 +67,17 @@ def template_switcher(templateargs):
...
+def avatar_color_code(obj):
+    # Return an int from 1 to avatar_color_code from the initials of the given string
+    initial = initials(obj.title)
+    parts = initial.split()
+    string_total = ord(parts[0][0])
+    if len(parts) > 1:
+        string_total += ord(parts[0][1])
+    return string_total % no_avatar_colours or no_avatar_colours
+
+
```
jace, line 80: Also not required.

jace (changes requested):
The Python code can be simplified a bit.

funnel/views/account.py
jace: This code effectively takes a modulus from 0 to 5, and then replaces 0 with 6 to get a 1-6 scale. I think it will be cleaner to read if we use 0-5 scale everywhere including in the CSS and JavaScript version.

PR #1367, opened 2022-05-12
Adds csrf_token as a fixture

djamg:


tests/conftest.py
```diff
@@ -823,3 +823,9 @@ def new_proposal(db_session, new_user, new_project):
     db_session.add(proposal)
     db_session.commit()
     return proposal
+
+
+@pytest.fixture
+def csrf_token(client):
+    csrf_token = client.get('/api/baseframe/1/csrf/refresh').get_data(as_text=True)
```
jace, line 830: This can be just `return client.get` without a variable assignment.

jace (approved):
Looks fine. One possible change for aesthetics.

PR #1373, opened 2022-05-16
flake8-print has changed namespace from T0* to T2*

jace:


jace:
Changed flake8-print==5.0.0, introduced in #1369.

PR #1397, opened 2022-06-21
Captured Vimeo request exception and added tests for it

djamg:


funnel/views/video.py
```diff
@@ -136,8 +136,13 @@ def video_property(obj: VideoMixin) -> Optional[VideoData]:
                 )
 
                 video_url = f'/videos/{obj.video_id}'
-                vimeo_resp = vimeo_client.get(video_url)
-                # vimeo_resp = requests.get(video_url, timeout=30)
+                try:
+                    vimeo_resp = vimeo_client.get(video_url)
+                    # vimeo_resp = requests.get(video_url, timeout=30)
+                except requests.exceptions.RequestException as exc:
+                    current_app.logger.error("Vimeo API request error: %s", repr(exc))
```
jace, line 143: Isn't pre-commit complaining about placing the exception in a log? Also, this could be `%r` and just `exc`.

jace (changes requested):

funnel/views/video.py
djamg: pre-commit did not catch it. The same block exists for youtube.

jace (approved):

PR #1402, opened 2022-06-30
Require phone number to allow content creation

jace:


jace:
image

Comment submit is broken as the JS doesn't recognise the error from the backend. Backend wants user to verify their phone number first.

PR #1427, opened 2022-07-19
Fix djlint issues

vidya-ram:


funnel/templates/403.html.jinja2
```diff
@@ -1,13 +1,13 @@
 {% extends "layout.html.jinja2" %}
-{% block title %}403 Forbidden{% endblock %}
+{% block title %}403 Forbidden{% endblock title %}
```
jace, line 2: Trans tag needed in this and other error templates.

jace (changes requested):

PR #1440, opened 2022-08-04
Make user parameter mandatory for ProfileForm

djamg:


funnel/forms/organization.py
```diff
@@ -6,7 +6,7 @@
 
 from flask import Markup, url_for
 
-from baseframe import _, __, forms
+from baseframe import _, __, current_auth, forms
```
jace, line 9: Import current_app from flask.

funnel/forms/organization.py
```diff
@@ -69,7 +69,10 @@ def validate_name(self, field) -> None:
             # from existing name, or has only changed case. This is a validation pass.
             return
         if reason == 'user':
-            if self.user.username and field.data.lower() == self.user.username.lower():
+            if (
+                current_auth.user.username
+                and field.data.lower() == current_auth.user.username.lower()
```
jace, line 74: Don't access current_auth in a form. The form is supposed to be a data validator that can be used anywhere. Instead, look for uses of the form in views and ensure the form is being given a user.

jace (changes requested):

funnel/forms/profile.py
```diff
@@ -24,8 +24,9 @@ class ProfileForm(OrganizationForm):
     A `profile` keyword argument is necessary for the ImgeeField.
     """
 
-    __expects__ = ('profile',)
+    __expects__ = ('profile', 'user')
     profile: Profile
+    user: User
```
jace, line 29: Get user from profile.user. It's not available when the form is used for an organization.

jace (changes requested):

jace (approved):

PR #1447, opened 2022-08-17
Replace jquery ajax with window.fetch

vidya-ram:


funnel/assets/js/comments.js
```diff
@@ -299,13 +310,17 @@ const Comments = {
...
+                  );
+                  fetch(lastSeenUrl, {
+                    method: 'POST',
+                    headers: {
+                      Accept: 'application/json',
+                      'Content-Type': 'application/x-www-form-urlencoded',
+                      'X-Requested-With': 'XMLHttpRequest',
                     },
-                    dataType: 'json',
+                    body: `csrf_token=${encodeURIComponent(csrfToken)}`,
```
jace, line 323: Would it help if we start accepting JSON submissions in all these endpoints? I've been working on a change in Coaster. We need one more in WTForms, which can be done in Baseframe.

funnel/assets/js/membership.js
```diff
@@ -56,25 +56,27 @@ const Membership = {
...
+              headers: {
+                Accept: 'application/json',
+                'X-Requested-With': 'XMLHttpRequest',
+              },
+            }).catch(Form.handleFetchNetworkError);
+            if (response && response.ok) {
+              const data = await response.json();
+              if (data) {
                 const vueFormHtml = data.form;
                 app.memberForm = vueFormHtml.replace(/\bscript\b/g, 'script2');
```
jace, line 74: I didn't see this earlier. Why is the word `script` being replaced with `script2`?

funnel/assets/js/scan_badge.js
```diff
@@ -34,61 +34,69 @@ const badgeScan = {
...
-
-      checkinAttendee(qrcode) {
+      async checkinAttendee(qrcode) {
         this.set({
           scanning: true,
           showModal: true,
         });
         $('#status-msg').modal('show');
-        const url = checkinApiUrl.replace('puk', qrcode.substring(0, 8));
+        const url = `${checkinApiUrl.replace('puk', qrcode.substring(0, 8))}?t`;
```
jace, line 43: What is the `?t` suffix for?

funnel/assets/js/schedule_view.js
```diff
@@ -120,17 +124,18 @@ const Schedule = {
...
-              success: (sessionHtml) => {
-                this.openModal(sessionHtml, currentPage, pageDetails);
-              },
-              error(response) {
-                const errorMsg = Form.getResponseError(response);
-                window.toastr.error(errorMsg);
+            const response = await fetch(activeSession.modal_url, {
+              headers: {
+                Accept: 'application/x.html+json',
+                'X-Requested-With': 'XMLHttpRequest',
```
jace, line 130: This header should not be needed when asking for `application/x.html+json`.

jace (changes requested):
Mostly okay. A couple improvements:

1. We can drop the `X-Requested-With: XMLHttpRequest` header entirely
2. Backend can change to accept JSON submissions instead of URL-encoded forms.

funnel/assets/js/membership.js
vidya-ram: This is for the script tag in the fetched html to work in vue framework
https://github.com/taoeffect/vue-script2

funnel/assets/js/scan_badge.js
vidya-ram: Checkin set to true

jace: We should be using the thorough regex used here: https://github.com/taoeffect/vue-script2#writing-script-instead-of-script2-using-script2ify

funnel/assets/js/membership.js
jace: We should be using the thorough regex used here: https://github.com/taoeffect/vue-script2#writing-script-instead-of-script2-using-script2ify. My version:

`/(?:</?script([\s>]))/gm`

PR #1453, opened 2022-08-23
Use available webdriver or skip/fail tests

djamg:


tests/conftest.py
```diff
@@ -827,6 +843,9 @@ def splinter_driver_kwargs(splinter_webdriver):
     """Disable certification verification for webdriver."""
     from selenium import webdriver
 
+    if not check_webdrivers():
+        pytest.skip('webdriver not found')
```
jace, line 847: How does this choose an available webdriver?

tests/conftest.py
```diff
@@ -35,6 +44,13 @@
 # --- Pytest config --------------------------------------------------------------------
 
 
+def check_webdrivers():
+    for driver in WEBDRIVERS.values():
+        if shutil.which(driver):
+            return True
+    return False
```
jace, line 51: This should be the `splinter_webdriver` fixture. Change body to:

```python
@pytest.fixture(scope='session')
def splinter_webdriver():
    for driver, executable in WEBDRIVERS.items():
        if shutil.which(executable):
            return driver
    pytest.skip("No webdrivers available")
```

Next task: make this compatible with a CLI override using `pytest --splinter-webdriver=whatever`.

jace (changes requested):

tests/conftest.py
djamg: Splinter does it for us. It chooses the driver based on availability. Should there be one more layer before it checks?

jace: It doesn't, which is why this PR was needed. It tries to use `firefox` even if there is no geckodriver binary.

jace: This is how you default to the driver specified on the CLI:

https://github.com/pytest-dev/pytest-splinter/blob/master/pytest_splinter/plugin.py#L95-L98

jace (approved):

PR #1455, opened 2022-08-23
Markdown parser using markdown-it-py

miteshashar:


funnel/utils/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Utility functions."""
+# flake8: noqa
+
+from .markdown import markdown
```
jace, line 4: Always do `import *` in `__init__.py`, and declare an `__all__` list in the respective modules if you don't want to export everything.

funnel/utils/markdown/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""Markdown parser using markdown-it-py."""
+# flake8: noqa
+
+from .base import markdown
```
jace, line 4: Same here, `import *`

funnel/utils/markdown/base.py
```diff
@@ -0,0 +1,137 @@
...
+from typing_extensions import Protocol
+
+from coaster.utils import make_name
+from coaster.utils.text import VALID_TAGS, normalize_spaces_multiline, sanitize_html
+
+MARKDOWN_HTML_TAGS = deepcopy(VALID_TAGS)
+
+
+# --- Typing Protocol for markdown-it-py plugins ---------------------------------------
+class MarkdownItPluginProtocol(Protocol):  # pylint: disable=too-few-public-methods
```
jace, line 18: This disable should not be necessary as it's already in the global list in `pyproject.toml`.

funnel/utils/markdown/base.py
```diff
@@ -0,0 +1,137 @@
...
+#       Port from https://github.com/markdown-it/markdown-it-sup
+#       Port from https://github.com/markdown-it/markdown-it-ins
+# TODO: emoji, (mark => highlight, inlinehilite)
+#       Port from https://github.com/markdown-it/markdown-it-emoji
+#       Port from https://github.com/markdown-it/markdown-it-mark
+#       Evaluate:
+#       https://www.npmjs.com/search?q=highlight%20keywords%3Amarkdown-it-plugin
+
+
+default_markdown_extensions_html: Mapping[str, Type[MarkdownItPluginProtocol]] = {
```
jace, line 39: The `Type[]` wrapper should not be necessary. You can use `Dict[str, MarkdownItPluginProtocol]` as this is an explicit dict and not an abstract Mapping.

jace (changes requested):

miteshashar (commented):
Need resolution on this to continue merging.

PR #1464, opened 2022-09-07
Use htmx instead jqueryform. Replace ajax with fetch

vidya-ram:


funnel/assets/js/account_form.js
```diff
@@ -0,0 +1,187 @@
...
+
+  async function checkUsernameAvailability(field) {
+    if (!waitingForUsernameResponse && $(field).val()) {
+      waitingForUsernameResponse = true;
+      const response = await fetch(accountUsernameUrl, {
+        method: 'POST',
+        headers: {
+          Accept: 'application/json',
+          'Content-Type': 'application/x-www-form-urlencoded',
+          'X-Requested-With': 'XMLHttpRequest',
```
jace, line 34: Can we drop the XHR header and change to JSON submit in this PR?

funnel/assets/js/account_form.js
```diff
@@ -0,0 +1,187 @@
...
+        }
+      } else {
+        waitingForUsernameResponse = false;
+        Form.getFetchError(response);
+      }
+    }
+  }
+
+  async function checkPasswordStrength(field) {
+    // FIXME: This function is getting called multiple times for same candidate
```
jace, line 73: Sounds like endpoint needs to set cache headers.

funnel/templates/forms.html.jinja2
```diff
@@ -250,6 +243,7 @@
 {%- endmacro %}
 
 {% macro tinymce3_editor(field, script=true) -%}
+  //Incase textarea uses tinymce, include `jquery.tinymce` in ext_requires in __init__.py
```
jace, line 246: This should be a Jinja2 comment using `{#- … #}`

funnel/templates/forms.html.jinja2
```diff
@@ -289,6 +283,7 @@
 {% endmacro %}
 
 {% macro tinymce4_editor(field, script=true) -%}
+  //Incase textarea uses tinymce, include `jquery.tinymce` in ext_requires in __init__.py
```
jace, line 286: Same

funnel/templates/forms.html.jinja2
jace: Also, the JS comment here is outside the JS `<script>` tag below.

jace: In fact we can remove this TinyMCE3 support entirely. It was used on Hasjob which has since upgraded to TinyMCE4. Version 3 was last used in Kharcha, which has been shutdown for several years.

runfrontendtests.sh
```diff
@@ -9,7 +9,7 @@ python -m tests.cypress.frontend_tests_initdb
 flask run -p 3002 --no-reload --debugger 2>&1 1>/tmp/funnel-server.log & echo $! > /tmp/funnel-server.pid
 function killserver() {
     kill $(cat /tmp/funnel-server.pid)
-    python -m tests.cypress.frontend_tests_dropdb
+    python -m frontend_tests_dropdb
```
jace, line 12: This won't work unless the file has moved to root folder, or the shell script has moved into the Cypress folder.

jace (changes requested):

funnel/assets/js/account_form.js
vidya-ram: Debouncing is present, will that suffice?
https://github.com/hasgeek/funnel/blob/main/funnel/templates/account_formlayout.html.jinja2#L84

runfrontendtests.sh
vidya-ram: In the next line, we cd into the cypress folder.

funnel/assets/js/account_form.js
vidya-ram: Changed

funnel/templates/forms.html.jinja2
vidya-ram: Removed

runfrontendtests.sh
jace: Ah, we should `cd` out before calling this module then. This function itself has no context for where it is being called from, so the assumption is that it's still in the home folder.

PR #1468, opened 2022-09-14
Improve homepage performance

vidya-ram:
Switch Vue with htmx(light weight JS framework) to load past project section.


funnel/templates/past_projects_section.html.jinja2
```diff
@@ -0,0 +1,20 @@
+{% for project in past_projects %}
+  <tr>
+    <td data-th="Date">
```
jace, line 3: Is `gettext` being called in JS? Else this should have a `{% trans %}` right here.

funnel/templates/past_projects_section.html.jinja2
```diff
@@ -0,0 +1,20 @@
...
+      <p class="mui--text-subhead zero-bottom-margin">{{ project.datetime }}</p>
+    </td>
+    <td data-th="Event name" class="past-events-table__title">
```
jace, line 6: Should be "Project" (pending a better template).

funnel/views/index.py
```diff
@@ -128,19 +128,17 @@ def home(self) -> ReturnRenderWith:
 IndexView.init_app(app)
 
 
-@app.route('/past.json')
+@app.route('/past')
```
jace, line 131: This will block 'past' as a username, so it either has to be in the reserved list in `models/helpers.py`, or the URL must have a character that is not valid in usernames (like a `.`)

jace (changes requested):
Some minor cleanup needed.

funnel/views/index.py
vidya-ram: Changed it to past_projects

funnel/views/profile.py
```diff
@@ -239,18 +239,17 @@ def user_proposals(self) -> ReturnRenderWith:
...
-    def past_projects_json(self, page: int = 1, per_page: int = 10) -> ReturnView:
+    @render_with('past_projects_section.html.jinja2')
+    def past_projects(self, page: int = 1, per_page: int = 10) -> ReturnView:
         projects = self.obj.listed_projects.order_by(None)
         past_projects = projects.filter(Project.state.PAST).order_by(
             Project.start_at.desc()
         )
         pagination = past_projects.paginate(page=page, per_page=per_page)
         return {
-            'status': 'ok',
```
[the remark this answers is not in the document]
vidya-ram: Added it back

PR #1472, opened 2022-09-16
Enhancements to markmap

miteshashar:
Markmap: Disable pan, enable autoFit and decrease fitRatio options. Aspect-ratio and viewport-based height adjustment for markmap svg element.

### DIFF
--- a/funnel/assets/js/utils/markmap.js
+++ b/funnel/assets/js/utils/markmap.js
@@ -24,7 +24,17 @@ const MarkmapEmbed = {
           asset += 1;
           if (window.markmap) {
             window.markmap = {
-              autoLoader: { manual: true },
+              autoLoader: {
+                manual: true,
+                onReady: () => {
+                  window.markmap.Markmap.defaultOptions = {
+                    ...window.markmap.Markmap.defaultOptions,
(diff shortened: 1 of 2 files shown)


funnel/assets/js/utils/markmap.js
```diff
@@ -3,6 +3,11 @@ const MarkmapEmbed = {
     $('.language-markmap').each(function embedMarkmap() {
       $(this).addClass('embed-added');
       $(this).find('code').addClass('markmap');
+      $(this).after(
+        $(
+          '<div>Click the empty white connectors <strong>&#9675;</strong> to close/collapse a node &amp; the filled coloured connectors <strong>&#9679;</strong> to open/expand a node.</div>'
```
jace, line 8: This needs `gettext` plus a term change from "click" to "tap" on mobile. IMO, best to not insert text into user content, and let the user figure this out.

jace (commented):
The change is okay. The caption is problematic because it's inserted into user content.

funnel/assets/js/utils/markmap.js
miteshashar: Got it

PR #1480, opened 2022-09-19
Markdown parser phase 2

### LINKED ISSUE #1483: Restructure markdown config profiles & plugin registry
We need to revamp the existing config mapping system to a simplified dictionary-based config profile structure.

We need to move the plugin registry from being class-based to a simplified dictionary structure.

Identified requirements:

The utils hierarchy should have no awareness of specific models. So, we save config profiles mapped to functional identifiers.
We use a layering principle where inner layers don't know anything about outer layers and only export APIs to them.

Hierarchy we should keep in mind when structuring our code:
```
utils → models → forms → registry → loginproviders → transports → views
```

### LINKED ISSUE #1486: Overhaul markdown tests for new code structure
Overhaul markdown tests to run with the new restructured config profile & plugin registry system.

Some considerations in discussion already are:

- [How we can do better diffs in pytest's console output.](https://eshlox.net/2020/04/07/better-diffs-in-pytest)

### LINKED ISSUE #1490: Decouple debug-rendering logic from markdown tests
The current markdown tests have debug-related rendering logic tightly coupled with the tests. These are deeply ingrained in the markdown test helpers too.
Image

So, testing output should be a pure check/validation and the content being checked should not be tampered with when the actual tests run. Hence, the debugging bells and whistles should be encapsulated only to be used during debug purposes.

- [x] Fixture for unified diff
- [x] Separate the functionality that generates output.html for comparision to be included in the above fixture or if not possible, another fixture.

### LINKED ISSUE #1493: markdown-it-py plugins for ins/del/sup/sub/mark
- Write plugins for:
  - [x] `++` » `<ins>`
  - [x] `~~` » `<del>` (Suppress default behaviour, converting to `<s>`)
  - [x] `^` » `<sup>`
  - [x] `~` » `<sub>`
  - [x] `==` » `<mark>`
- [x] Write tests for all the above plugins

### LINKED ISSUE #1496: markdown-it-py plugin for embeddable fenced blocks
- [x] Fenced blocks being used for rendering embeddable diagrams, mindmaps, flowcharts, etc need to be changed to accept syntax similar to that of [MyST roles & directives](https://myst-parser.readthedocs.io/en/latest/syntax/roles-and-directives.html). 
- [x] This change needs to be handled on the front-end too by using the same css classes for activation.
- [x] Write tests

### LINKED ISSUE #1506: Store markdown test case results in multiline form
Store the markdown test case results in the test case `.toml` files in multiline form, for better readability.

Ref: https://github.com/hasgeek/funnel/pull/1480#discussion_r1000238700

### LINKED ISSUE #1515: Introduce prismjs
The new markdown parser already renders out-of-the-box HTML5 style markup (with the `language-*` class) for fenced blocks.

prismjs activates styling for `language-*` blocks with it's autoloader, which we will utilise.

miteshashar:
Resolves #1483 
Resolves #1486 
Resolves #1490 
Resolves #1493
Resolves #1496 
Resolves #1506
Resolves #1515


funnel/models/helpers.py
```diff
@@ -588,3 +592,127 @@ def process_bind_param(self, value, dialect):
...
+    :param str group: Defer column group
+    :param kwargs: Additional column options, passed to SQLAlchemy's column constructor
+    """
+    # Construct a custom subclass of MarkdownComposite and set the markdown processor
+    # and processor options on it. We'll pass this class to SQLAlchemy's composite
+    # constructor.
+    class CustomMarkdownComposite(MarkdownCompositeNative):
+        pass
+
+    CustomMarkdownComposite.profile = profile
```
jace, line 707: Instead of doing it here, make those subclasses at top-level and make a `Dict[str, type]`. It'll make debugging easier, as currently it will resolve as `markdown_column_native.<locals>`.

funnel/models/venue.py
```diff
@@ -25,7 +32,9 @@ class Venue(UuidMixin, BaseScopedNameMixin, CoordinatesMixin, db.Model):
         grants_via={None: project_child_role_map},
     )
     parent = sa.orm.synonym('project')
-    description = MarkdownColumn('description', default='', nullable=False)
+    description = MarkdownColumnNative(  # type: ignore[has-type]
+        'description', profile='basic', default='', nullable=False
```
jace, line 36: Should `profile='basic'` be the default value?

funnel/utils/markdown/base.py
```diff
@@ -32,63 +33,88 @@
...
+        try:
+            _profile = profiles[profile]
+        except KeyError as exc:
+            raise KeyError(
+                f'Wrong markdown config profile "{profile}". Check name.'
+            ) from exc
+    elif isinstance(profile, dict):
+        _profile = profile
+    else:
+        raise TypeError('Wrong type - profile has to be either str or dict')
```
jace, line 69: IMO, skip `dict` type. If the columns aren't using it, this will be redundant code.

funnel/utils/markdown/base.py
```diff
@@ -32,63 +33,88 @@
...
+
+    args = _profile.get('args', ())
+
+    md = MarkdownIt(*args)
+
+    funnel_config = _profile.get('funnel_config', {})
```
jace, line 75: What is this? Needs clearer documentation.

funnel/utils/markdown/base.py
```diff
@@ -32,63 +33,88 @@
...
+
+    if md.linkify is not None:
+        md.linkify.set({'fuzzy_link': False, 'fuzzy_email': False})
+
+    for action in ['enableOnly', 'enable', 'disable']:
+        if action in funnel_config:
+            getattr(md, action)(funnel_config[action])
```
jace, line 82: This effectively means `funnel_config` is not a TypedDict as there is no strong link between what it contains and what happens in this function. This could be better.

funnel/utils/markdown/base.py
```diff
@@ -32,63 +33,88 @@
...
+
+    for e in _profile.get('plugins', []):
```
jace, line 84: I think profile should also be a `dataclass` with default values. Avoid all uses of `.get`.

funnel/utils/markdown/presets.py
```diff
@@ -0,0 +1,13 @@
...
+
+__all__ = []  # type: ignore[var-annotated]
+
+
+class custom_profile:  # noqa: N801
+    @staticmethod
+    def make():
+        config = gfm_like.make()
+        # ... options to process single line
+        return config
```
jace, line 13: If profile is a dataclass, you can simply subclass it.

funnel/utils/markdown/profiles.py
```diff
@@ -0,0 +1,71 @@
...
+#     ),
+#     'funnel_config' : { # Optional
+#         'enable': List = [],
+#         'disable': List = [],
+#         'enableOnly': List = [],
+#         'render_with': str = 'render'
+#     }
+# }
+
+profiles: Dict[str, Dict[str, Any]] = {
```
jace, line 50: Turn each profile into a `dataclass` and make a dict that refers to each.

funnel/views/api/markdown.py
```diff
@@ -17,7 +17,7 @@ def markdown_preview() -> ReturnView:
     # mtype = request.form.get('type')
     text = request.form.get('text')
 
-    html = markdown(text)
+    html = markdown(text, 'document')
```
jace, line 20: This should now use `mtype`. Maybe rename it to `profile` in the API call arguments.

tests/data/markdown/basic.toml
```diff
@@ -76,8 +76,8 @@ ___
...
...link <a href=\"https://github.com/hasgeek\">https://github.com/hasgeek</a> (will autoconvert if linkify is on)</p>\n<hr />\n<h2 id=\"h:markdown-it-typography\">Markdown-it typography <a class=\"header-anchor\" href=\"#h:markdown-it-typography\">¶</a></h2>\n<p>The results of the below depend on the typographer options enabled by us in the <code>markdown-it-py</code> parser, if <code>typographer=True</code> has been passed to it.</p>\n<p><strong>The below should convert if <code>replacements</code> has been enabled.</strong></p>\n<p>(c) (C) (r) (R) (tm) (TM) (p) (P) +-<br />\ntest.. test... test..... test?..... test!....<br />\n!!!!!! ???? ,,  -- ---</p>\n<p><strong>The below should convert the quotes if <code>smartquotes</code> has been enabled.</strong></p>\n<p>&quot;Smartypants, double quotes&quot; and 'single quotes'</p>\n<hr />\n<h2 id=\"h:blockquotes\">Blockquotes <a class=\"header-anchor\" href=\"#h:blockquotes\">¶</a></h2>\n<blockquote>\n<p>Blockquotes can also be nested...</p>\n<blockquote>\n<p>...by using additional greater-than signs right next to each other...</p>\n<blockquote>\n<p>...or with spaces between arrows.</p>\n</blockquote>\n</blockquote>\n</blockquote>\n<hr />"
```
jace, line 83: These should use multiline strings for readability.

jace (changes requested):

funnel/models/venue.py
miteshashar: It's not a default value. This specifies `Venue.description` shall use the `basic` profile.

funnel/utils/markdown/base.py
miteshashar: Now we are not doing dataclasses, right?

tests/data/markdown/basic.toml
miteshashar: Can I post an issue for this and resolve it separately?

miteshashar: As discussed, this will be addressed in #1506

funnel/models/helpers.py
```diff
@@ -588,3 +592,127 @@ def process_bind_param(self, value, dialect):
...
+        return cls(value)
+
+
+def markdown_column_native(
+    name: str,
+    deferred: bool = False,
+    group: Optional[str] = None,
+    profile: str = 'basic',
+    **kwargs,
+) -> composite:
```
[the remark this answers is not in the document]
miteshashar: Looks like SQLAlchemy has not defined composite as a generic type. So, not making this change.

vidya-ram (changes requested):

funnel/assets/js/utils/markmap.js
```diff
@@ -1,10 +1,34 @@
...
+      });
+  },
+  resizeTimer: null,
+  resizeMarkmapContainers() {
+    if (this.resizeTimer) clearTimeout(this.resizeTimer);
+    this.resizeTimer = setTimeout(() => {
+      $('.md-embed-markmap.activated svg').each(function mmresized() {
+        const circles = $(this).find('circle');
+        const firstNode = circles[circles.length - 1];
+        firstNode.dispatchEvent(new Event('click'));
```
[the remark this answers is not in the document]
miteshashar: This is a hacky way to resize the markmap visualisation, since there was no way through the API. We are triggering this on window resize.

jace (approved):

funnel/assets/js/comments.js
```diff
@@ -172,7 +172,7 @@ const Comments = {
         },
         activateForm(action, textareaId, parentApp = app) {
           if (textareaId) {
-            const copyTextAreaContent = function (view) {
+            const copyTextAreaContent = (view) => {
```
jace, line 175: @vidya-ram is this okay? This change caused problems last time.

funnel/assets/js/utils/markmap.js
```diff
@@ -41,6 +66,7 @@ const MarkmapEmbed = {
           loadMarkmapScript();
         } else {
           self.addMarkmap();
+          window.addEventListener('resize', this.resizeMarkmapContainers.bind(this));
```
jace, line 69: Does the event listener need to be removed when the content changes in an SPA?

funnel/assets/js/utils/prism.js
```diff
@@ -0,0 +1,58 @@
...
+        window.Prism.highlightElement(this);
+      }
+    );
+  },
+  hooked: false,
+  loadPrism() {
+    const CDN_CSS = 'https://unpkg.com/prismjs/themes/prism.min.css';
+    const CDN = [
+      'https://unpkg.com/prismjs/components/prism-core.min.js',
+      'https://unpkg.com/prismjs/plugins/autoloader/prism-autoloader.min.js',
```
jace, line 14: Shouldn't these URLs be in webpack or other dependency management system? They'll get scattered around code like this otherwise, and will be unauditable.

jace (approved):

PR #1488, opened 2022-09-22
Add Telegram notifications for repo activity

jace:


jace (commented):
This appears to be good enough to merge now.

jace (commented):
I think review URL is fixed now.

jace (commented):
This fix makes it simpler but also harder to maintain because we've embedded JSON inside YAML, but the JSON isn't validated at commit and can fail at runtime. Requires extra care when editing.

jace (commented):
I think we've fixed the newlines problem.

And also gone back from JSON to YAML because newlines in JSON were the problem.

PR #1499, opened 2022-10-05
Merge `with` statements

arnu515:
These two `with` statements can be merged into one. It will be a bit faster this way


jace:
Thank you for the contribution. We missed this last week as I was travelling and mostly offline. Merging now.

arnu515:
Thanks!

-------- Original Message --------

PR #1510, opened 2022-11-03
Add Codemirror 6 to submission form

vidya-ram:


miteshashar:
Submission form views when there are no labels created for the project.

*New Submission*
image

*Edit Submission*
image

The placement of the 3 possible fields here should be handled such that they are aesthetically placed in the UI.

1 field: Left
2 fields: Left, Right
3 fields: Left, Center, Right

miteshashar:
There is no way to go back from the Add Collaborator popup. The only way is reloading the page, which would mean all unsaved changes are lost.

image

vidya-ram:
> Submission form views when there are no labels created for the project.
> 
> **New Submission** image
> 
> **Edit Submission** image
> 
> The placement of the 3 possible fields here should be handled such that they are aesthetically placed in the UI.
> 
> 1 field: Left 2 fields: Left, Right 3 fields: Left, Center, Right

Fixed

vidya-ram:
> There is no way to go back from the Add Collaborator popup, in case one does not want to add one. image The only way is reloading the page, which would mean all unsaved changes are lost.
> 
> Moreover, the user dropdown does not load any values. image
> 
> Clicking 'Add Collaborator' goes to the API url(losing all changes made to content). On the new page, the user field renders as an autocomplete field, though in the modal, it renders as a dropdown. image
> 
> Submitting this form successfully gets processed, but replaces the full page with JSON. image

Fixed

miteshashar:
Submitting blank form shows a toastr message with text "parsererror".

image

vidya-ram:
> Submitting blank form shows a toastr message with text "parsererror".
> 
> image

Fixed

funnel/static/js/scripts.js
```diff
@@ -72,6 +72,14 @@ function activate_widgets() {
   });
 }
 
+function activate_select2() {
+  /* Upgrade to jquery 3.6 select2 autofocus isn't working. This is to fix that problem.
+    select2/select2#5993  */
+  $(document).on('select2:open', function () {
+    document.querySelector('.select2-search__field').focus();
```
jace, line 79: Will this work if the page has multiple select2 widgets?

funnel/views/proposal.py
```diff
@@ -247,33 +247,44 @@ def edit(self) -> ReturnRenderWith:
...
-                'html': render_template(
-                    'collaborator_list.html.jinja2',
-                    collaborators=[
-                        _m.current_access(datasets=['primary', 'related'])
-                        for _m in self.obj.memberships
-                    ],
-                ),
-            }, 201
+        if request.method == 'POST':
+            if collaborator_form.validate_on_submit():
```
jace, line 251: `form.validate_on_submit()` is shorthand for `request.method == 'POST' and form.validate()` so the first `if` statement is not required.

jace (changes requested):
One redundant `if` statement in the Python code.

PR #1516, opened 2022-11-14
Fix delete operation in labels and submission form page

vidya-ram:
1. Fix delete collaborator in submission form page
2. Fix deletion and archive of labels
3. Prevent double submit of cfp toggle

### DIFF
--- a/funnel/views/label.py
+++ b/funnel/views/label.py
@@ -8,6 +8,7 @@ from flask import flash, request
 from werkzeug.datastructures import MultiDict
 
 from baseframe import _, forms
+from baseframe.forms import render_form
 from coaster.views import ModelView, UrlForView, render_with, requires_roles, route
 
 from .. import app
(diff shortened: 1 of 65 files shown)


funnel/views/label.py
```diff
@@ -205,39 +206,60 @@ def edit(self) -> ReturnRenderWith:
...
             flash(_("The label has been archived"), category='success')
-        else:
-            flash(_("CSRF token is missing"), category='error')
-        return render_redirect(self.obj.project.url_for('labels'))
+        return render_form(
+            form=form,
+            title=_("Confirm archive of label"),
+            message=_(
+                "Archive this label? This operation is permanent and cannot be"
+                " undone"
```
jace, line 222: Archive is not permanent.

jace (changes requested):
A couple small issues.

PR #1520, opened 2022-11-15
Markdown preview fix

djamg:
Reverted back to old markdown utils in the API


miteshashar (approved):
Checked by running the same diff I shared earlier, looks good.
P.S. I did not run & test it.

vidya-ram (approved):

jace (changes requested):

jace (approved):

PR #1530, opened 2022-11-23
Ics fix

djamg:
The upcoming sessions are added to the calendar instead of the first session in a project.

### DIFF
--- a/funnel/views/schedule.py
+++ b/funnel/views/schedule.py
@@ -135,7 +135,7 @@ def schedule_ical(project: Project, rsvp: Optional[Rsvp] = None):
     cal.add('refresh-interval;value=duration', 'PT12H')
     cal.add('x-published-ttl', 'PT12H')
     for session in project.scheduled_sessions:
-        if session.start_at > utcnow():
+        if session.end_at > utcnow():
             cal.add_component(session_ical(session, rsvp))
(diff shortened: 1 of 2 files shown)


funnel/views/schedule.py
```diff
@@ -135,7 +135,8 @@ def schedule_ical(project: Project, rsvp: Optional[Rsvp] = None):
     cal.add('refresh-interval;value=duration', 'PT12H')
     cal.add('x-published-ttl', 'PT12H')
     for session in project.scheduled_sessions:
-        cal.add_component(session_ical(session, rsvp))
+        if session.end_at > utcnow():
+            cal.add_component(session_ical(session, rsvp))
```
jace, line 139: Function `schedule_ical` is also used by the subscribe to calendar feature. Add a parameter `future_only: bool = False` and change the `if` condition here to `if future_only and session.end_at > now`. Define `now = utcnow()` outside the `for` loop so it's not being called repeatedly.

In `views/notifications/rsvp_notification.py` line 45, pass `future_only=True`.

jace (changes requested):

funnel/views/schedule.py
jace: Sorry, the `if` condition should be `if not future_only or session.end_at > now`

jace (approved):

PR #1548, opened 2022-12-15
Send notification read status to server on opening url with `?utm_source=<notification_id>`

### LINKED ISSUE #1206: Opening a notification link from SMS/email should also mark it read
Notifications currently remain unread unless the user opens the Updates page and waits for the timeout period. This shouldn't be necessary.

The links we generate currently include a `utm_source=<notification-id>`. This parameter is not used in the backend, but could be used to mark a matching UserNotification instance as read.

This will require a request processor (before/after) to watch for this tag (or a replacement tag with a lower chance of false positives).

There is also a small attack vector when the notification id in the URL doesn't match the page being rendered. Example scenario:

1. Two users receive a notification with the same id (project updates, comments, multiple other scenarios).
2. The first user fails to read the notification because their delivery channels are ineffective (not checking SMSes or emails).
3. The second user performs an attack by sending a crafted URL to the first user using a channel they consume (say WhatsApp messages). This URL has some other page but with the notification id.
4. The notification is now marked as read even though the first user didn't read the intended content.

This attack can be mitigated by allowing a notification to validate URLs -- requiring a bind between view handlers and notification types. However, it is also low risk enough to be ignored until we see actual abuse.

vidya-ram:
Fixes #1206

### DIFF
--- a/funnel/templates/layout.html.jinja2
+++ b/funnel/templates/layout.html.jinja2
@@ -264,6 +264,7 @@
       window.Hasgeek.Config.unreadCommentUrl = {{ url_for('comments')|tojson }};
     {% endif %}
     window.Hasgeek.Config.shorturlApi = {{ url_for('create_shortlink')|tojson }};
+    window.Hasgeek.Config.markReadUrl = {{ url_for('notification_mark_read', eventid_b58='eventid_b58')|tojson }};
   </script>
   <script src="{{ built_asset('app.js') }}" type="text/javascript"></script>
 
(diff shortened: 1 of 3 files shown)


funnel/templates/layout.html.jinja2
```diff
@@ -264,6 +264,7 @@
       window.Hasgeek.Config.unreadCommentUrl = {{ url_for('comments')|tojson }};
     {% endif %}
     window.Hasgeek.Config.shorturlApi = {{ url_for('create_shortlink')|tojson }};
+    window.Hasgeek.Config.markReadUrl = {{ url_for('notification_mark_read', eventid_b58='eventid_b58')|tojson }};
```
jace, line 267: Maybe we should disambiguate by calling it `markNotificationReadUrl`.

jace (approved):

PR #1551, opened 2022-12-17
Prism plugins

miteshashar:


funnel/assets/sass/_layout.scss
```diff
@@ -957,6 +957,38 @@
   height: 99% !important;
 }
 
+:not(pre) > code,
```
jace, line 960: Doesn't the `:not()` selector have a significant performance impact?

jace: Apparently not a concern anymore. I remember when it was introduced and was controversial for the performance impact it caused. https://stackoverflow.com/questions/44668606/css-not-performance

PR #1570, opened 2023-01-03
Changed the padding on ordered and un-ordered lists to fix clipping

anishTP:


jace:
We currently compile SASS files before git commit, not at deploy. This can be done with `compass compile`, and Compass can be installed with `gem install compass` (using Ruby Gems).

Assigning this to @vidya-ram as she's current maintainer of our SASS files.

PR #813, opened 2020-08-06
Profile website field

iambibhas:

PR #1579, opened 2023-01-04
Markdown table width

anishTP:


package.json
```diff
@@ -57,9 +57,11 @@
     "jsqr": "^1.4.0",
     "leaflet": "^1.7.1",
     "markmap": "^0.6.1",
+    "markmap-common": "^0.14.2",
     "markmap-lib": "^0.14.3",
     "markmap-view": "^0.14.3",
     "mermaid": "^9.3.0",
+    "node": "^19.3.0",
```
jace, line 64: Server has Node 16 and cannot upgrade at this time.

funnel/assets/sass/update.scss
```diff
@@ -39,7 +39,7 @@
   .update__content {
     display: inline-block;
     width: 100%;
-    word-break: break-word;
+    word-break: normal;
```
jace, line 42: `break-word` was introduced in #813 because some URLs were so long that they were overflowing into the sidebar, and causing side scroll on mobile. [MDN docs suggest the modern replacement](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) is:

```css
word-break: normal;
overflow-wrap: break-word;
```

jace (changes requested):

anishTP:
This fix is inadequate. A new one is being worked on.

PR #1582, opened 2023-01-06
Tests for mustache template escaper.

miteshashar:


tests/unit/utils/test_mustache.py
```diff
@@ -0,0 +1,82 @@
...
+
+### Vendors
+> No vendors
+""",
+)
+
+
+@pytest.mark.parametrize(('template'), TEMPLATES.keys())
+def test_mustache_md(template):
+    (template, expected_output) = TEMPLATES[template]
```
jace, line 80: IMO, don't use the same dict for parametrization and inside the test. Parametrize both.

jace (changes requested):

PR #1597, opened 2023-01-23
Fix otp login modal on project page

vidya-ram:


funnel/templates/password_login_form.html.jinja2
```diff
@@ -5,7 +5,7 @@
         method="post"
         class="mui-form mui-form--margins"
         accept-charset="UTF-8"
-        action="{{ url_for('login') }}">
+        action="{{ url_for('login', next=request.path) }}%23register-modal">
```
jace, line 8: You can specify `_anchor='register-modal'` as a parameter to `url_for`.

jace (commented):
One small optimisation.

Does Recaptcha work if the handling of RecaptchaField is removed?

vidya-ram:
> One small optimisation.
> 
> Does Recaptcha work if the handling of RecaptchaField is removed?

This was the old and redundant one. We had newly introduced this as part of OTP
https://github.com/hasgeek/funnel/blob/main/funnel/templates/forms.html.jinja2#L214

funnel/templates/otpform.html.jinja2
```diff
@@ -2,7 +2,7 @@
 {% from "forms.html.jinja2" import renderform, renderform_inner %}
 
 {% block form %}
-  <form data-parsley-validate="true" id="{{ ref_id }}" method="post" {%- if action %} action="{{ action }}" {%- endif %}{%- if multipart %} enctype="multipart/form-data" {%- endif %} accept-charset="UTF-8" class="mui-form mui-form--margins hg-form">
+  <form data-parsley-validate="true" id="{{ ref_id }}" method="post" hx-post="{{ url_for('login', next=request.path) }}%23register-modal" {%- if action %} action="{{ url_for('login', next=request.path) }}%23register-modal" {%- endif %}{%- if multipart %} enctype="multipart/form-data" {%- endif %} accept-charset="UTF-8" class="mui-form mui-form--margins hg-form" hx-target="#loginformwrapper">
```
jace, line 5: OTP form is also used for `sudo`, where it does not redirect to login. This one should not hardcode login URL.

jace (changes requested):

funnel/views/login.py
```diff
@@ -141,6 +143,10 @@ def login() -> ReturnView:
     # Remember where the user came from if it wasn't already saved.
     save_session_next_url()
     next_url = session['next']
+    action_url = url_for('login', next=next_url)
```
jace, line 146: This line should be in the `else:` branch.

PR #1600, opened 2023-01-24
JS helper function to get faicon html

vidya-ram:


funnel/assets/js/utils/helper.js
```diff
@@ -402,6 +402,25 @@ const Utils = {
     }
     return false;
   },
+  getFaviconHTML(icon, iconSize = 'body', baseline = true, cssClassArray=[]) {
```
jace, line 405: Favicon != FA Icon. Function needs to be renamed. Also, we should upgrade from FA5 to FA6.

PR #1605, opened 2023-01-27
Import prismjs from webpack

vidya-ram:


funnel/assets/js/utils/helper.js
```diff
@@ -402,7 +402,7 @@ const Utils = {
     }
     return false;
   },
-  getFaviconHTML(icon, iconSize = 'body', baseline = true, cssClassArray=[]) {
+  getFaviconHTML(icon, iconSize = 'body', baseline = true, cssClassArray = []) {
```
jace, line 405: Thought this was renamed already. Looks like this needs to be updated against `main`.

webpack.config.js
```diff
@@ -125,6 +125,10 @@ module.exports = {
           from: 'node_modules/leaflet/dist/images',
           to: 'css/images',
         },
+        {
+          from: 'node_modules/prismjs/components/*.min.js',
+          to: 'js/prismjs/components/[name].js',
```
vidya-ram, line 130: Prism language packs are copies to this build folder. Depending on the code snippets in markdown, it will be fetched.

miteshashar (approved):

PR #1628, opened 2023-02-13
Reintroduce recaptcha for for new email/phone

vidya-ram:


funnel/templates/forms.html.jinja2
```diff
@@ -211,34 +219,51 @@
...
-    var formId = "{{ formId }}";
-    var forWrapperId = "{{ forWrapperId }}";
-    window.onInvisibleRecaptchaSubmit = function (recaptcha_response) {
-      var postUrl = $('#' + formId).attr('action');
-      var onSuccess = function(responseData) {
-        $('#' + forWrapperId).html(responseData);
+{%- macro recaptcha(formId, ajax=false, forWrapperId='') %}
+  {%- if ajax %}
+    <script type="text/javascript">
+      var formId = "{{ formId }}";
```
jace, line 225: Use `var formId = {{ formId|tojson }};`. Not using any filter will cause the default HTML escaping filter to be applied.

funnel/templates/forms.html.jinja2
```diff
@@ -211,34 +219,51 @@
...
+      var forWrapperId = "{{ forWrapperId }}";
```
jace, line 226: Also `|tojson` here.

funnel/templates/forms.html.jinja2
```diff
@@ -211,34 +219,51 @@
...
 {%- endmacro %}
-{%- macro recaptcha(formId, forWrapperId) %}
-  <script type="text/javascript">
-    var formId = "{{ formId }}";
-    var forWrapperId = "{{ forWrapperId }}";
-    window.onInvisibleRecaptchaSubmit = function (recaptcha_response) {
-      var postUrl = $('#' + formId).attr('action');
-      var onSuccess = function(responseData) {
-        $('#' + forWrapperId).html(responseData);
+{%- macro recaptcha(formId, ajax=false, forWrapperId='') %}
```
jace, line 222: Is `forWrapperid` a typo for `formWrapperId`?

funnel/templates/loginform.html.jinja2
```diff
@@ -5,6 +5,6 @@
 {% block form %}
   {{ passwordlogin(form, formid, ref_id, action) }}
   {%- if form and form.recaptcha is defined %}
-    {{ recaptcha(ref_id, 'loginformwrapper') }}
+    {{ recaptcha(ref_id, true, 'loginformwrapper') }}
```
jace, line 8: IMO, the macro should have the `ajax` parameter last, and invocations should explicitly use `ajax=true`. The standalone `true` here is ambiguous, and since we don't have type checking for Jjinja2 templates, this will cause future accidents.

funnel/templates/password_login_form.html.jinja2
```diff
@@ -63,7 +63,7 @@
       {% trans terms_url=url_for('policy', path='policy/terms'), privacy_path=url_for('policy', path='policy/privacy') %}By signing in, you agree to Hasgeek’s <a href="{{ terms_url }}" target="_blank" rel="noopener">terms of service</a> and <a href="{{ privacy_path }}" target="_blank" rel="noopener">privacy policy</a>{% endtrans %}
     </p>
   </form>
-  {%- if loginform and loginform.recaptcha is defined %}{{ recaptcha(ref_id, 'loginformwrapper') }}{%- endif %}
+  {%- if loginform and loginform.recaptcha is defined %}{{ recaptcha(ref_id, true, 'loginformwrapper') }}{%- endif %}
```
jace, line 66: Same here with `ajax=true`

jace (changes requested):
One possible typo, two string escapes and one parameter order recommendation.

jace (approved):

PR #1638, opened 2023-02-16
Markdown plugin for abbreviations

miteshashar:


funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+
+    simple_re = re.compile('(?:' + labels_re_str + ')')
+
+    match_re_str = (
+        '(^|' + PADDING_CHARACTERS + ')'
+        '(' + labels_re_str + ')'
+        '($|' + PADDING_CHARACTERS + ')'
+    )
+
+    match_re = re.compile(match_re_str)
```
jace, line 104: `re.compile` should be called once per startup to be useful. This should be outside the function.

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+    '['
+    + ''.join(
+        chr(x) for x in range(65536) if unicodedata.category(chr(x))[0] in ('P', 'Z')
+    )
+    .replace('\\', '\\\\')
+    .replace('[', '\\[')
+    .replace(']', '\\]')
+    + '\r\n$+<=>^`|~'
+    + ']'
+)
```
jace, line 33: Since this is in regex format and not a simple string of characters, the name should be `PADDING_CHARS_REGEX` or `PADDING_CHARS_RE` (we use `re` lowercase prefix or suffix for a compiled regex).

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+
+    j, block_tokens_length = 0, len(block_tokens)
+    while j < block_tokens_length:
+        block_token = block_tokens[j]
+        if block_token.type != 'inline':
+            j += 1
+            continue
+        tokens = block_token.children
+
+        i = len(tokens) - 1  # type: ignore[arg-type]
```
jace, line 114: It'll help to have clearer names for the `i` and `j` variables to know what they do.

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+            i -= 1
+
+        j += 1
+
+
+def abbr_plugin(md: MarkdownIt):
+    md.block.ruler.before(
+        'reference', 'abbr_def', abbr_def, {'alt': ['paragraph', 'reference']}
+    )
+    md.core.ruler.after('linkify', 'abbr_replace', abbr_replace)
```
jace, line 173: What happens if the `reference` and `linkify` rules are removed? Do these get added in the correct place?

tests/unit/utils/markdown/data/abbr.toml
```diff
@@ -0,0 +1,2687 @@
...
+The language [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) is a markup language for webpages.
+
+*[CSS]: Cascading Style Sheets
+
+CSS is a language used to define the presentation of HTML webpages.
+
+> You can find more about CSS on https://developer.mozilla.org/en-US/docs/Web/CSS
+
+### Tests for padding characters
+*[ABBR]: Abbrieviation
```
jace, line 16: Typo. This should be "Abbreviation"

jace (changes requested):
`re.compile` should be at module level for performance. A couple other cosmetic issues.

funnel/utils/markdown/mdit_plugins/abbr.py
miteshashar: `reference` & `linkify` are rules specified by the library. They aren't expected to be removed unless the library authors make that change.
In that case, our tests would fail on upgrade.

However, if you meant what would happen if they are disabled, then in that case they would be in the rule chain sequence, but not be run. It should not effect the plugin in anyway, if so.

funnel/utils/markdown/mdit_plugins/abbr.py
miteshashar: This cannot be run on startup, since it uses the user's inputs.

funnel/utils/markdown/mdit_plugins/abbr.py
jace: Okay, this is not a problem then.

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+from markdown_it.rules_inline import StateInline
+from markdown_it.token import Token
+
+__all__ = ['abbr_plugin']
+
+ASTERISK_CHAR = 0x2A  # ASCII value for `*`
+OPEN_SQBR_CHAR = 0x5B  # ASCII value for `[`
+CLOSE_SQBR_CHAR = 0x5D  # ASCII value for `]`
+BACKSLASH_CHAR = 0x5C  # ASCII value for `\`
+COLON_CHAR = 0x3A  # ASCII value for `:`
```
jace, line 21: I'm wondering if we should define these constants as `ord('*')`. It's a function call, but it's self-validating unlike the comment.

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+
+PADDING_CHARACTERS_RE = (
+    '['
+    + ''.join(
+        chr(x) for x in range(65536) if unicodedata.category(chr(x))[0] in ('P', 'Z')
+    )
+    .replace('\\', '\\\\')
+    .replace('[', '\\[')
+    .replace(']', '\\]')
+    + '\r\n$+<=>^`|~'
```
jace, line 31: How come no `\t` in this list?

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+COLON_CHAR = 0x3A  # ASCII value for `:`
+
+PADDING_CHARACTERS_RE = (
+    '['
+    + ''.join(
+        chr(x) for x in range(65536) if unicodedata.category(chr(x))[0] in ('P', 'Z')
+    )
+    .replace('\\', '\\\\')
+    .replace('[', '\\[')
+    .replace(']', '\\]')
```
jace, line 30: This may be clearer if it uses raw strings:

```python
    .replace(r'\', r'\\')
    .replace(r'[', r'\[')
    .replace(r']', r'\]')
```

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+    + '\r\n$+<=>^`|~'
+    + ']'
+)
+
+
+def abbr_def(state: StateBlock, start_line: int, end_line: int, silent: bool):
```
jace, line 36: return type `-> bool:`

jace: Docstring needed since it's not apparent to reader where this function is called.

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+    pos = state.bMarks[start_line] + state.tShift[start_line]
+    maximum = state.eMarks[start_line]
+
+    if pos + 2 >= maximum:
+        return False
+
+    # Check if line is of the format `*[Abbreviation]:Full Form`
```
jace, line 43: Isn't that `*[Abbreviation]: Full form` with a space?

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+        return False
+
+    if silent:
+        return True
+
+    # Extract label and title and store it in state.env
+    label = state.src[label_start:label_end].replace('\\', '')
+    title = state.src[label_end + 2 : maximum].strip()
+    if len(label) == 0 or len(title) == 0:
+        return False
```
jace, line 74: Shouldn't this be a regex? It'll be far faster than char-by-char testing.

funnel/utils/markdown/mdit_plugins/abbr.py
```diff
@@ -0,0 +1,173 @@
...
+        [state.md.utils.escapeRE(k) for k in sorted(state.env['abbr'].keys(), key=len)]
+    )
+
+    simple_re = re.compile('(?:' + labels_re_str + ')')
+
+    match_re_str = (
+        '(^|' + PADDING_CHARACTERS_RE + ')'
+        '(' + labels_re_str + ')'
+        '($|' + PADDING_CHARACTERS_RE + ')'
+    )
```
jace, line 102: I just realised this padding char regex is unnecessary. This could simply be `(^|\W)` and `($|\W)`, with the regex compiled using `re.UNICODE` flag.

[Documentation on `\W`](https://docs.python.org/3/library/re.html#index-35).

jace (changes requested):

funnel/utils/markdown/mdit_plugins/abbr.py
jace: Replacement regex:

`'^\s*\*\[(\w+)\]:\s*(.*)$'`

jace: I see `elif ch == BACKSLASH_CHAR:`, which means the following abbreviation is allowed:

`*[ab[b\]r]: Make ab[b]r into an abbreviation`

jace: IMO, this `ab[b]r` type example is not likely, and in the absence of an official spec for Markdown abbreviations, we can ignore these edge cases.

The `\w+` version allows an abbreviation to be a single word without spaces or hyphens or punctuation. This will not work for an abbreviation of the form `CAN-SPAM` because of the hyphen, so maybe we need the `[^\]]+` version allowing anything other than a `]`.

Or we could use `(?:[^\]|\])+` to allow the `\]` escape as well, giving an overall regex of:

`r'^\s*\*\[((?:[^\]|\])+)\]:\s*(.*)$'`

jace: Original implementation in PHP Markdown Extra: https://github.com/michelf/php-markdown/blob/932cbc7760983ce44f693fc3138f7ccebe685ccd/Michelf/MarkdownExtra.php#L1800-L1871

funnel/utils/markdown/mdit_plugins/abbr.py
miteshashar: https://docs.python.org/3/library/re.html#re.ASCII
> Note that for backward compatibility, the re.U flag still exists (as well as its synonym re.UNICODE and its embedded counterpart (?u)), but these are redundant in Python 3 since matches are Unicode by default for strings (and Unicode matching isn’t allowed for bytes).

miteshashar (commented):
All changes done

jace (dismissed):

jace (approved):

PR #1644, opened 2023-02-20
Account follower and project registration models

jace:


jace:
`ActorMembershipMixin` makes user optional, but this is only required for project registration. Account follower requires a mandatory user account.

The mixin should make this configurable using a `__user_optional__` flag.

PR #1649, opened 2023-02-26
Move to pip-compile-multi for dependency management

miteshashar:


.github/workflows/pytest.yml
```diff
@@ -65,10 +65,11 @@ jobs:
       - name: Install Geckodriver (for browser testing)
         uses: browser-actions/setup-geckodriver@latest
       - name: Install Python
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
+          cache-dependency-path: '**/requirements/test.txt'
```
jace, line 72: Doesn't this need both `base.txt` and `test.txt`?

jace (approved):

PR #1658, opened 2023-03-08
Don't hide links in markdown heading

vidya-ram:


funnel/assets/sass/components/_markdown.scss
```diff
@@ -17,6 +15,8 @@
     font-weight: 700;
     a {
       color: $mui-text-dark;
+    }
+    a.header-anchor {
       display: none;
     }
     &:hover a {
```
jace, line 22: This should also be `a.header-anchor`

jace (changes requested):

jace (approved):

PR #1672, opened 2023-03-20
New style for project CTA buttons

vidya-ram:


funnel/views/project.py
```diff
@@ -71,70 +71,70 @@ class CountWords:
...
-        __("You have registered"),
+        __("You are the first"),
         __("One follower so far"),
-        __("You are following this"),
+        __("You are the first"),
     ),
     CountWords(
         __("Two registrations so far"),
-        __("You and one other have registered"),
+        __("You & one other"),
```
jace, line 80: Lazy strings must be HTML-escaped, so `&` should be `&amp;`.

funnel/views/project.py
```diff
@@ -71,70 +71,70 @@ class CountWords:
...
+        __("You & nine others"),
     ),
 ]
 greater_than_10_count = CountWords(
     __("{num} registrations so far"),
-    __("You and {num} others have registered"),
+    __("You & {num} others"),
     __("{num} followers so far"),
-    __("You and {num} others are following"),
+    __("You & {num} others"),
```
jace, line 137: There are corresponding tests to be updated.

jace (changes requested):

jace (approved):

PR #1677, opened 2023-03-27
Use different text for register button

vidya-ram:


funnel/views/project.py
```diff
@@ -77,64 +77,64 @@ class CountWords:
     ),
     CountWords(
         __("Two registrations so far"),
-        __("You &amp; one other"),
+        __("You & one other"),
```
jace, line 80: Lazy strings should be HTML escaped, so this should be `&amp;`. The corresponding test in `tests/unit/views/project_test.py` is broken because it's still expecting `&amp;` there.

PR #1687, opened 2023-04-12
On touch screen devices, show alternate text for registered users

vidya-ram:


funnel/views/project.py
```diff
@@ -72,70 +72,70 @@ class CountWords:
     CountWords(__("Be the first to register!"), '', __("Be the first follower!"), ''),
     CountWords(
         __("One registration so far"),
-        __("You are the first"),
+        __("You have registered"),
         __("One follower so far"),
-        __("You are the first"),
+        __("You are following this"),
```
jace, line 77: These changes have to be reflected in the test as well — the test is now broken.

jace (changes requested):

jace (approved):

PR #1688, opened 2023-04-17
Upgrade pip-audit in pre-commit and python packages

miteshashar:
Upgrade python packages. pip-tools 6.12.3 was not compatible with latest version of pip(23.1).Update pip-audit version to 2.5.4 to overcome https://github.com/pypa/pip-audit/issues/552

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -14,7 +14,7 @@
     #   baseframe
 async-generator==1.10
     # via trio
-beautifulsoup4==4.12.0
+beautifulsoup4==4.12.2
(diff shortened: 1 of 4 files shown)


requirements/test.txt
```diff
@@ -96,10 +93,6 @@ sttable==0.0.1
     # via -r requirements/test.in
 tenacity==8.2.2
     # via pytest-selenium
-tomli==2.0.1
```
jace, line 99: Which Python was this built under? The removal of tomli is surprising.

jace (dismissed):

PR #1700, opened 2023-05-08
Move select2 import from baseframe to webpack

vidya-ram:
1. Move select2 import from baseframe to webpack
2. Change select2 to a dynamic import to reduce bundle size for pages that don't use select2 and improve page performance.


funnel/__init__.py
```diff
@@ -135,7 +135,6 @@
     requires=['funnel'],
     ext_requires=[
         'pygments',
```
jace, line 137: We're not using Pygments anymore. That also can be removed.

funnel/assets/js/utils/form_widgets.js
```diff
@@ -60,7 +60,7 @@ export const Widgets = {
     };
     this.activateToggleSwitch(checkboxId, onSuccess);
   },
-  activate_select2() {
+  activateSelect2Focus() {
     /* Upgrade to jquery 3.6 select2 autofocus isn't working. This is to fix that problem.
       select2/select2#5993  */
```
jace, line 65: This ticket is marked as closed as jQuery 3.7 fixes it. Do we still need this patch?

jace (approved):
Two side issues, not directly affecting this PR.

funnel/__init__.py
vidya-ram: Sure. Will take it up as a new PR

funnel/assets/js/utils/form_widgets.js
vidya-ram: Will open a new PR to upgrade to jQuery 3.7 and remove this fix

PR #1715, opened 2023-05-19
Changed utm tracking tags

djamg:
This change will affect GA and Matoma analytics data. 
We are also migrating to GA4 from Universal Analytics and historic data will only be available on Universal Analytics property in Google Analytics Dashboard.


funnel/views/notification.py
```diff
@@ -198,9 +198,12 @@ def tracking_tags(
...
             unsubscribe links and other specialized links will want to specify another)
         """
-        tags = {'utm_campaign': campaign, 'utm_medium': transport}
+        tags = {
+            'utm_campaign': f'{self.notification.type}-{self.notification.created_at.strftime("%Y%m%d-%H%M")}',
+            'utm_medium': transport,
+        }
         if not self.notification.for_private_recipient:
-            tags['utm_source'] = self.notification.eventid_b58
+            tags['utm_source'] = 'notification'
```
jace, line 206: The `if` clause is no longer relevant as there is no event identifier, so nothing that can be traced to the individual.

funnel/views/notification.py
```diff
@@ -198,9 +198,12 @@ def tracking_tags(
         :param campaign: Reason for this link being sent (default 'notification' but
             unsubscribe links and other specialized links will want to specify another)
```
jace, line 199: This parameter is now being ignored.

jace (changes requested):

jace (approved):

jace (approved):

PR #1727, opened 2023-05-29
Sample files for settings - basic, development & production

miteshashar:


jace:
IMO, we should be dumping settings files, so even the existing sample should go away. We can offer a sample `.env` instead.

miteshashar:
That would make sense if eventually coaster will also stop loading instance settings.

jace:
Coaster will not load from any files if called with `coaster.app.init_app(app, ['env'])`. The call in Funnel is currently `coaster.app.init_app(app, ['py', 'toml'])`

PR #1732, opened 2023-05-29
Load config from env, deprecating config from `.py` files (resolves #1729)

### LINKED ISSUE #1729: Per-app config from ENV required for devserver
Funnel provides multiple apps from a single module. These apps end up sharing configuration that is explicitly meant for a single app. Currently, a special config file `hasgeekapp.py` is used to isolate config for the main app, but this is a kludge.

We're migrating to config from environment variables. Since a worker (uwsgi) will be bound to only a specific app (currently `app` or `shortlinkapp`), it can supply app-specific config in the env and simply ignore the misconfiguration of other apps. However, this doesn't work for development where the apps are multiplexed via a single host app in `funnel.devtest:AppByHostWsgi`. They will again share a single configuration. Also, the `app_url_for` helper consults the config of unused apps, so misconfiguration is a problem there.

We can overcome this by using app-specific env prefixes. Flask defaults to `FLASK_` prefix, which we can keep for common config, and then apply a second round of config using `FLASK_{app_name}_` prefix. This feature needs to be exposed via Coaster's `coaster.app.init_app`.

jace:


jace:
`testing.py` should be removed in this PR. Hardcoded values can go directly into the `app` fixture in `tests/conftest.py`. Values loaded from env don't need to be in this file at all.

miteshashar (approved):

PR #1733, opened 2023-06-02
Drop requirement for REDIS_URL in settings

jace:


jace:
RQ is not tested in pytest as `RQ_ASYNC` is set to False. The database number we put in `.testenv` is immaterial. Maybe it too needs a non-async test with a spawned background worker.

miteshashar (approved):

PR #1734, opened 2023-06-02
Remove unmaintained types-* and upgrade others

jace:


miteshashar (approved):
Approved since all tests pass on GHA

PR #1736, opened 2023-06-04
Enable caching for webpack, babel-loader and ESLintPlugin

miteshashar:


vidya-ram (approved):

jace (approved):

.github/workflows/pytest.yml
```diff
@@ -65,17 +67,45 @@ jobs:
...
       - name: Install Node
         uses: actions/setup-node@v3
         with:
           node-version: latest
           cache: npm
+      - name: Cache node modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: ${{ matrix.os }}-${{ matrix.python-version }}-node_modules-${{ hashFiles('package-lock.json') }}
```
jace, line 86: Why is the Python version a factor here? For parallelized access to cache?

.github/workflows/pytest.yml
```diff
@@ -65,17 +67,45 @@ jobs:
...
+      #   uses: actions/cache@v3
+      #   with:
+      #     path: .eslintcache
+      #     key: ${{ matrix.os }}-${{ matrix.python-version }}-eslintcache-${{ hashFiles('webpack.config.js') }}-${{ hashFiles('.eslintrc.js') }}-${{ hashFiles('funnel/assets/**/*.js') }}-${{ hashFiles('funnel/assets/**/*.scss') }}
+      # - name: Cache .webpack_cache
+      #   id: webpack-cache
+      #   uses: actions/cache@v3
+      #   with:
+      #     path: .webpack_cache
+      #     key: ${{ matrix.os }}-${{ matrix.python-version }}-webpack-${{ hashFiles('webpack.config.js') }}-${{ hashFiles('funnel/assets/**/*.js') }}-${{ hashFiles('funnel/assets/**/*.scss') }}
```
jace, line 105: Remove these lines. We have a copy anyway in the PR.

webpack.config.js
```diff
@@ -17,6 +17,10 @@ module.exports = {
       path: require.resolve('path-browserify'),
     },
   },
+  cache: {
+    type: 'filesystem',
+    cacheDirectory: path.resolve(__dirname, '.webpack_cache'),
+  },
```
jace, line 23: Does this cache work even if GHA cache is not being used?

jace (changes requested):

webpack.config.js
miteshashar: Yes. This is independent.

.github/workflows/pytest.yml
miteshashar: The 2 tests are running in parallel and their caches should be maintained separately, else the cache for one would be overwritten by another. Even if the content of the caches generated by both are the same, this is an area of uncertainty IMO, depending on how caches are being maintained behind the scenes.

jace: Will it work to use just `{{ matrix }}`? Let all the sub-vars be used then.

miteshashar: I am not sure if `{{ matrix }}` is accepted by GHA? Should I test it out?

jace: Yes. There may be a way to get a flattened representation too.

jace (approved):

.github/workflows/pytest.yml
jace: For flattened representation, use `${{ toJSON(matrix) }}` https://docs.github.com/en/actions/learn-github-actions/expressions#tojson

jace (approved):

jace (approved):

jace (approved):

PR #1740, opened 2023-06-08
Use async for dailystats and include data from Matomo analytics

djamg:
This PR introduces the first instance of async functions in Funnel. Async is used only to make concurrent HTTP requests at this time, but the learnings from this implementation allow for expanded usage elsewhere.


funnel/cli/periodic.py
```diff
@@ -29,6 +30,39 @@ class DataSource:
...
+    pages = requests.get(pages, timeout=5).json()
+
+    referrers_list = [(k['label'], k['nb_visits']) for k in referrers]
+    socials_list = [(k['label'], k['nb_visits']) for k in socials]
+    pages_list = [(k['label'], k['nb_visits']) for k in pages]
+
+    message = ""
+
+    message += "\n*Top Pages:*\n"
+    for index, i in enumerate(itertools.islice(pages_list, 7), 1):
```
jace, line 52: Why is `itertools.islice` needed here? It does the same thing as `pages_list[:7]`

funnel/cli/periodic.py
```diff
@@ -29,6 +30,39 @@ class DataSource:
...
+    tz = pytz.timezone('Asia/Kolkata')
+    now = utcnow().astimezone(tz)
+    yesterday = (now - relativedelta(days=1)).strftime('%Y-%m-%d')
+    referrers = f'https://matomo.hasgeek.tech?module=API&method=Referrers.getWebsites&idSite=1&period=day&date={yesterday}&filter_limit=-1&format=json&token_auth={app.config["MATOMO_TOKEN"]}'
+    socials = f'https://matomo.hasgeek.tech?module=API&method=Referrers.getSocials&idSite=1&period=day&date={yesterday}&filter_limit=-1&format=json&token_auth={app.config["MATOMO_TOKEN"]}'
+    pages = f'https://matomo.hasgeek.tech?module=API&method=Actions.getPageTitles&idSite=1&period=day&date={yesterday}&filter_limit=-1&format=json&token_auth={app.config["MATOMO_TOKEN"]}'
+
+    referrers = requests.get(referrers, timeout=5).json()
+    socials = requests.get(socials, timeout=5).json()
+    pages = requests.get(pages, timeout=5).json()
```
jace, line 43: You need exception handlers for when the timeout happens.

funnel/cli/periodic.py
```diff
@@ -29,6 +30,39 @@ class DataSource:
...
+    socials_list = [(k['label'], k['nb_visits']) for k in socials]
+    pages_list = [(k['label'], k['nb_visits']) for k in pages]
+
+    message = ""
+
+    message += "\n*Top Pages:*\n"
+    for index, i in enumerate(itertools.islice(pages_list, 7), 1):
+        message += f"{index}. {i[0][1:20]}: {i[1]}\n"
+
+    message += "\n*Referrers:*\n"
```
jace, line 55: IIRC, in Telegram's Markdown mode, bold format requires two asterisks. This should be `**Referrers:**`

jace (changes requested):

funnel/cli/periodic.py
djamg: Tested it and confirmed. Its only one asterisk: `*Referrers:*`

jace (approved):

miteshashar (approved):

jace (approved):

PR #1745, opened 2023-06-12
Added Matomo visit metric

djamg:


funnel/cli/periodic/stats.py
```diff
@@ -109,11 +109,21 @@ class MatomoData:
     referrers: Sequence[MatomoResponse]
     socials: Sequence[MatomoResponse]
     pages: Sequence[MatomoResponse]
+    visits: Dict
```
jace, line 112: This needs to be a fully specced type.

funnel/cli/periodic/stats.py
```diff
@@ -331,6 +343,12 @@ async def dailystats() -> None:
                 f"\n"
             )
 
+    if matomo_data.visits:
+        message += f"\n*Unique visits:* {matomo_data.visits['nb_uniq_visitors']}\n"
```
jace, line 347: This data should be embedded above in the user count.

jace (approved):

PR #1746, opened 2023-06-12
Update RSVP SMS templates

djamg:


funnel/views/notifications/rsvp_notification.py
```diff
@@ -16,12 +16,43 @@
...
 
 
+class RegistrationConfirmationTemplate(SmsTemplate):
+    """DLT registered template for rsvp notificaion."""
+
+    registered_template = (
+        'You have registered for {#var#}, scheduled for {#var#}. '
+        'For more information, visit {#var#}.'
+        '\n\nhttps://bye.li to stop - Hasgeek'
+    )
```
jace, line 32: Make one for the other template too, with no scheduled date.

funnel/views/notifications/rsvp_notification.py
```diff
@@ -86,19 +117,17 @@ def email_content(self):
             ),
         )
 
-    def sms(self) -> OneLineTemplate:
+    def sms(self) -> RegistrationConfirmationTemplate:
         project = self.rsvp.project
         next_at = project.next_starting_at()
         if next_at:
```
jace, line 123: Return either of two templates based on `next_at` being not None

jace (changes requested):

jace (approved):

PR #1748, opened 2023-06-13
Gracefully handle email recipient rejection by the SMTP server

jace:
A string in the form `anything@123` will pass as an email address in the login form, but will be rejected by the SMTP server. This patch recasts the SMTP error as a form validation error.

A test is included. Since Flask-Mailman will silently drop email in a test environment, the test has to mock Flask-Mailman instead of the underlying `smtplib.SMTP.sendmail` where the error is raised in production.

### DIFF
--- a/devserver.py
+++ b/devserver.py
@@ -7,6 +7,8 @@ import sys
 from flask.cli import load_dotenv
 from werkzeug import run_simple
 
+from coaster.utils import getbool
+
 if __name__ == '__main__':
     load_dotenv()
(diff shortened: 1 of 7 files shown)


jace:
This implementation recasts all errors as an `OtpDeliveryError`. We need to retain the distinct variants of `TransportError` and show `TransportRecipientError` as a form validation error, but the other types as flash messages.

miteshashar (dismissed):

PR #1755, opened 2023-06-16
Fix decorator typing and sync isort and ruff config

jace:
Decorators are now typed using ParamSpec, removing the need for a type cast.

In the process, ruff and isort were found to be in conflict with each other as ruff replicates isort's functionality but didn't have the same config. Isort had a non-default value for `from_first`. Reverting to the default has caused imports to be reordered in a large number of files.


jace:
This PR is pending on #1759 and hasgeek/coaster#389, which should fix Webpack asset handling in CI.

PR #1756, opened 2023-06-17
Working webpack cache and temporary solution of building assets twice to make tests pass on GHA

miteshashar:
Add caching for webpack and built assets on pytest gha workflow; experiment with building assets twice in pytest gha workflow as a temporary solution


jace:
Second build is not necessary with the new webpack extension in #1759. The cache may be useful though.

jace (approved):

PR #1770, opened 2023-06-28
Update logging config for the new names used in Coaster

jace:


funnel/__init__.py
```diff
@@ -27,10 +27,13 @@
 
 #: Main app for hasgeek.com
 app = Flask(__name__, instance_relative_config=True)
+app.name = 'funnel'
 #: Shortlink app at has.gy
 shortlinkapp = Flask(__name__, static_folder=None, instance_relative_config=True)
+app.name = 'shortlink'
```
miteshashar, line 33: This should be `shortlinkapp.name`

funnel/__init__.py
```diff
@@ -27,10 +27,13 @@
...
 #: Unsubscribe app at bye.li
 unsubscribeapp = Flask(__name__, static_folder=None, instance_relative_config=True)
+app.name = 'unsubscribe'
```
miteshashar, line 36: This should be `unsubscribeapp.name`

sample.env
```diff
@@ -110,16 +110,26 @@ FLASK_SES_NOTIFICATION_TOPICS='[]'
 
 # --- Logging
 # Logfile (will be logrotated by date)
-FLASK_LOGFILE=error.log
+FLASK_LOG_FILE=error.log
```
miteshashar, line 113: Can we move this to `logs/error.log`?

There's a caveat here though.
I have already done this on the docker branch. The logs` folder has to be manually created as python logging doesn't create a folder on its own.

One solution could be to have a blank file checked in inside the `logs` folder.

miteshashar (changes requested):

sample.env
jace: For the new server, I'm considering `/home/apps/logs/{app.name}.log` so they're all in a single folder. Also, the new default is to _not_ log to a file. I'll change it reflect here.

miteshashar (approved):

PR #1776, opened 2023-07-03
Use a range query for Matomo's week/month-to-date stats

djamg:


jace (approved):

funnel/cli/periodic/stats.py
```diff
@@ -149,6 +149,15 @@ async def matomo_response_json(
 
 
 async def matomo_stats(date: str = 'yesterday') -> MatomoData:
+    # Dates in report timezone (for display)
+    tz = pytz.timezone(app.config['TIMEZONE'])
+    now = utcnow().astimezone(tz)
+    # Dates cast into UTC (for db queries)
```
jace, line 155: This comment is not valid here.

funnel/cli/periodic/stats.py
```diff
@@ -149,6 +149,15 @@ async def matomo_response_json(
...
+    today = midnight_to_utc(now)
+    last_week = today - relativedelta(weeks=1)
+    last_month = today - relativedelta(months=1)
+    week_range = f'{last_week.strftime("%Y-%m-%d")},{today.strftime("%Y-%m-%d")}'
+    month_range = f'{last_month.strftime("%Y-%m-%d")},{today.strftime("%Y-%m-%d")}'
```
jace, line 160: These two queries will be one day short, because "today" is exactly at midnight and there's no data for the day. This needs "yesterday".

The db queries are using datetime, but Matomo only needs date.

jace (changes requested):

funnel/cli/periodic/stats.py
```diff
@@ -149,6 +149,15 @@ async def matomo_response_json(
 
 
 async def matomo_stats(date: str = 'yesterday') -> MatomoData:
```
jace, line 151: `date` as a parameter is useless now given everything else is hardcoded in the function. Might as well remove from here and put it in the body as `date = 'yesterday'`

jace (changes requested):

jace (approved):

PR #1778, opened 2023-07-03
Change explored communities section on homepage

vidya-ram:
As part of improving home page design, change explored communities section on homepage to show featured profile from app config

### DIFF
--- a/funnel/views/index.py
+++ b/funnel/views/index.py
@@ -94,7 +94,9 @@ class IndexView(ClassView):
             .order_by(Project.next_session_at.asc())
             .all()
         )
-        featured_profiles = Profile.query.filter(Profile.name_in(app.config.get('FEATURED_ACCOUNTS', []))).all()
+        featured_profiles = Profile.query.filter(
+            Profile.name_in(app.config.get('FEATURED_ACCOUNTS', []))
+        ).all()
 
         return {
             'all_projects': [
--- a/funnel/views/index.py
+++ b/funnel/views/index.py
@@ -94,9 +94,15 @@ class IndexView(ClassView):
             .order_by(Project.next_session_at.asc())
             .all()
         )
-        featured_profiles = Profile.query.filter(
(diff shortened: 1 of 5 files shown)


funnel/views/index.py
```diff
@@ -94,6 +94,9 @@ def home(self) -> ReturnRenderWith:
             .order_by(Project.next_session_at.asc())
             .all()
         )
+        featured_profiles = (
+            Profile.query.filter_by(name=p).first() for p in current_app.config['FEATURED_PROFILES']
+        )
```
jace: Faster query:

`featured_profiles = Profile.query.filter(Profile.name_in(app.config.get('FEATURED_ACCOUNTS', []))).all()`

Note the config name change — we're deprecating the term Profile, so this is a bit in advance of the main PR. Also, read from `app` not `current_app`, as this view is limited to one app.

funnel/views/index.py
```diff
@@ -116,13 +119,7 @@ def home(self) -> ReturnRenderWith:
...
             ),
             'featured_profiles': [
-                p.current_access(datasets=('primary', 'related'))
-                for p in Profile.query.filter(
-                    Profile.is_verified.is_(True),
-                    Profile.organization_id.is_not(None),
-                )
-                .order_by(sa.func.random())
-                .limit(6)
+                p.access_for(roles={'all'}, datasets=('primary', 'related')) for p in featured_profiles if p
```
jace: The `if p` part can be removed now as the list is guaranteed to not have `None` in it

funnel/views/index.py
```diff
@@ -5,7 +5,7 @@
 import os.path
 from dataclasses import dataclass
 
-from flask import Response, g, render_template
+from flask import Response, g, render_template, current_app
```
jace: `current_app` is not needed here.

jace (changes requested):

jace (approved):

PR #1782, opened 2023-07-04
SMS templates for Comment notifications

djamg:


funnel/views/notifications/comment_notification.py
```diff
@@ -19,9 +19,61 @@
...
 from ..helpers import shortlink
 from ..notification import RenderNotification
+from .mixins import TemplateVarMixin
+
+
+class CommentReplyTemplate(TemplateVarMixin, SmsTemplate):
+    """DLT registered template for RSVP without a next session."""
+
+    registered_template = (
+        "{#var#} has replied to your comment: {#var#}"
```
jace, line 32: `registered_template` should use single quotes as it's a fixed reference string. Exception when it has a single quote in the body, such as `there's` (in proposal template), where double quotes are okay.

jace (changes requested):

funnel/views/notifications/comment_notification.py
```diff
@@ -19,9 +19,61 @@
     NewCommentNotification,
     User,
 )
-from ...transports.sms import OneLineTemplate
+from ...transports.sms import OneLineTemplate, SmsTemplate
```
jace, line 22: Where is OneLineTemplate being used?

jace (approved):

funnel/views/notifications/comment_notification.py
djamg: It is being used in the CommentReportNotification

PR #1783, opened 2023-07-04
Email templates

anishTP:
Added the notification view for project published that gets triggered on `/transition` route. Currently gets triggered when a project gets published and withdrawn. The pinned update macro also needs to be incorporated. And a bug fix in Windows mail.


jace:
This has moved to #1792.

PR #1784, opened 2023-07-05
markdown-it-py plugin anchors_ext

miteshashar:
Markdown plugin to modify the token stream produced by the mdit_py plugin heading_anchors to make the heading a permalink, instead of an appended permalink


funnel/utils/markdown/base.py
```diff
@@ -280,6 +287,7 @@ def render(self, text: Optional[str]) -> Optional[Markup]:
         'footnote',
         'footnote_ext',  # Must be after 'footnote' to take effect
         'heading_anchors',
+        'anchors_ext',  # Must be after 'heading_anchors' to take effect
```
jace, line 290: Should we call it `heading_anchors_fix`?

funnel/utils/markdown/mdit_plugins/anchors_ext.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def anchors_ext(state: StateInline) -> None:
+    prev_token = None
+
+    for token in state.tokens:
+        if prev_token is None:
+            prev_token = token
+            continue
+        if token.type == 'inline' and prev_token.type == 'heading_open':  # type: ignore
```
jace, line 18: `type: ignore[specific_type]` needed

funnel/utils/markdown/base.py
miteshashar: Yes, sure!
Was just following the same convention as `footnotes_ext`.

funnel/utils/markdown/mdit_plugins/anchors_ext.py
miteshashar: Ok

miteshashar: Though, the failure was statement not reachable.

tests/unit/utils/markdown/data/headings.toml
```diff
@@ -29,7 +29,12 @@ profiles = [ "basic", "document",]
 [config.custom_profiles.heading_anchors]
 preset = "default"
 options_update = {"breaks" = true}
-plugins = [ "heading_anchors",]
+plugins = [ "heading_anchors"]
```
jace, line 32: Space mismatch on either side.

jace (approved):

PR #1797, opened 2023-07-12
Mailer models compatible with Hasmail, with no mailer functionality yet

jace:


funnel/models/mailer.py
```diff
@@ -0,0 +1,455 @@
...
+        self._nickname = value
+
+    @property
+    def email(self) -> str:
+        return self._email
+
+    @email.setter
+    def email(self, value: str) -> None:
+        self._email = value.lower()
+        self.md5sum = md5sum(value)
```
jace, line 368: To be removed for EmailAddressMixin.

funnel/models/mailer.py
```diff
@@ -0,0 +1,455 @@
...
+
+    @property
+    def revision_id(self) -> Optional[int]:
+        return self.draft.revision_id if self.draft else None
+
+    def is_latest_draft(self) -> bool:
```
jace, line 374: Maybe `is_using_latest_draft` and as hybrid property to find all that need updating.

funnel/models/mailer.py
```diff
@@ -0,0 +1,455 @@
...
+    def get_rendered(self) -> MarkdownString:
+        """Get Mustache-rendered Markdown text."""
+        if self.draft:
+            return mustache_md(self.template or '', self.template_data())
+        draft = self.mailer.draft()
+        if draft is not None:
+            return mustache_md(draft.template or '', self.template_data())
+        return MarkdownString('')
+
+    def get_preview(self) -> str:
```
jace, line 403: This should return `Markup`

funnel/models/mailer.py
```diff
@@ -0,0 +1,455 @@
...
+        if actor is not None and actor == self.user:
+            perms.update(['edit', 'delete', 'send', 'new-recipient', 'report'])
+        return perms
+
+    def draft(self) -> Optional[MailerDraft]:
+        if self.drafts:
+            return self.drafts[-1]
+        return None
+
+    def render_preview(self, text: str) -> str:
```
jace, line 181: Return as `Markup`

PR #1801, opened 2023-07-14
Remove CleanWebpackPlugin and use output.clean option

### LINKED ISSUE #1735: Remove clean-webpack-plugin and enable usage of output.clean option in webpack 5
[clean-webpack-plugin is deprecated in webpack 5](https://stackoverflow.com/questions/64617228/cleanwebpackplugin-does-not-clean-in-webpack-5).

We should move to use the native `output.clean` option available in webpack 5.

vidya-ram:
Fixes #1735


miteshashar (dismissed):
Build works fine for me on node v18.16. Also fine on GHA.

Browsed through basic pages and did not notice anything unusual.

jace:
This has a merge conflict with `main` courtesy the upgrades in #1814.

jace (approved):
Is this downgrading Cypress and select2 again?

PR #1815, opened 2023-07-21
Replace Selenium with Playwright for browser tests

djamg:


tests/conftest.py
```diff
@@ -59,6 +59,11 @@ def firefox_options(firefox_options):
     return firefox_options
 
 
+@pytest.fixture(scope="session")
+def browser_context_args(browser_context_args):
+    return {**browser_context_args, "ignore_https_errors": True}
```
jace, line 64: Why is this needed?

djamg: ignores SSL errors and certificate warnings

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+
+def check_recaptcha_loaded(page: Page) -> None:
+    page.wait_for_selector(
+        "#form-passwordlogin > div.g-recaptcha > div > div.grecaptcha-logo > iframe",
+        timeout=10000,
+    )
+
+
+@given("Anonymous visitor is on the home page")
+def given_anonuser_home_page(live_server, page) -> None:
```
jace, line 25: Specify type for Page here too, and in all other tests.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+    page.goto(live_server.url)
+    expect(page).to_have_title(re.compile("Test Hasgeek"))
```
jace, line 27: There's no point calling re.compile at runtime. If this function accepts a plain string, you can pass one, as this is a plain text pattern.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+
+@when(
+    parsers.re(
+        "they navigate to the login page and submit"
+        " (?P<phone_or_email>a phone number|an email address)"
+    ),
+    target_fixture='anon_username',
+)
+def when_anonuser_navigates_login_and_submits(
+    app, live_server, phone_or_email, page
```
jace, line 38: Type `phone_or_email: str`

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+) -> dict[str, str]:
+    if phone_or_email == "a phone number":
+        username = '8123456789'
+    elif phone_or_email == "an email address":
+        username = 'anon@example.com'
+    else:
+        pytest.fail("Unknown username type")
+    page.click(".header__button")
```
jace, line 46: These should use single quotes as they're symbols.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+    check_recaptcha_loaded(page)
+    page.wait_for_selector("input[name=username]").fill(username)
+    page.click("#form-passwordlogin button")
```
jace, line 49: Likewise in both these lines – single quotes.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+def given_twoflower_visits_project(
+    live_server, page, db_session, user_twoflower, new_project
+) -> None:
+    user_twoflower.password = TWOFLOWER_PASSWORD
+    user_twoflower.add_phone(TWOFLOWER_PHONE)
+    user_twoflower.add_email(TWOFLOWER_EMAIL)
+    new_project.publish()
+    db_session.add(new_project)
+    db_session.commit()
+    page.goto(live_server.url + new_project.account.urlname + '/' + new_project.name)
```
jace, line 165: Use `new_project.absolute_url` instead of constructing it like this.

jace (changes requested):
Many strings with double quotes (esp CSS class selectors) should be in single quotes to indicate they're symbols – any change to the value here should be synced with the same string elsewhere. Other issues are as indicated in comments.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+
+scenarios('account/register.feature')
+pytestmark = pytest.mark.usefixtures('live_server')
+
+TWOFLOWER_EMAIL = 'twoflower@example.org'
+TWOFLOWER_PHONE = '+12015550123'
+TWOFLOWER_PASSWORD = 'te@pwd3289'  # nosec
+
+
+def check_recaptcha_loaded(page: Page) -> None:
```
jace, line 15: This should be `wait_until_recaptcha_loaded` and it should return a bool indicating success or failure.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+    return {'phone_or_email': phone_or_email, 'username': username}
+
+
+@then("they are prompted for their name and the OTP, which they provide")
+def then_anonuser_prompted_name_and_otp(live_server, anon_username, page: Page) -> None:
+    page.wait_for_selector('input[name=fullname]').fill('Twoflower')
+    if anon_username['phone_or_email'] == "a phone number":
+        otp = live_server.transport_calls.sms[-1].vars['otp']
+    elif anon_username['phone_or_email'] == "an email address":
+        otp = live_server.transport_calls.email[-1].subject.split(' ')[-1]
```
jace, line 57: For safety, this should use a regex pattern to extract the OTP from the subject. If the subject format changes, this test will fail for regex not matching instead of incorrect OTP – making the problem easier to spot.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+        pytest.fail("Unknown username type")
+    page.wait_for_selector('input[name=otp]').fill(otp)
+    page.click('#form-otp button')
+
+
+@then("they get an account and are logged in")
+def then_they_are_logged_in(live_server, page: Page) -> None:
+    assert (
+        page.get_by_text("You are now one of us. Welcome aboard!").inner_text()
+        == "You are now one of us. Welcome aboard!"
```
jace, line 68: Don't repeat the string. You can instead test that the selector found a match.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+
+@given("Twoflower visitor is on the home page")
+def when_twoflower_visits_homepage(
+    live_server, page: Page, db_session, user_twoflower
+) -> None:
+    page.goto(live_server.url)
+    user_twoflower.password = TWOFLOWER_PASSWORD
+    user_twoflower.add_phone(TWOFLOWER_PHONE)
+    user_twoflower.add_email(TWOFLOWER_EMAIL)
+    db_session.commit()
```
jace, line 80: IMO, move these lines into a fixture, something like `user_twoflower_with_password_and_contacts`. The fixture depends on the underlying fixtures and returns the user object.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+    return {'phone_or_email': phone_or_email, 'username': username}
+
+
+@given("Twoflower is on the project page")
+def given_twoflower_visits_project(
+    live_server, page: Page, db_session, user_twoflower, new_project
+) -> None:
+    user_twoflower.password = TWOFLOWER_PASSWORD
+    user_twoflower.add_phone(TWOFLOWER_PHONE)
+    user_twoflower.add_email(TWOFLOWER_EMAIL)
```
jace, line 161: These lines should use the same fixture as above.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+    new_project.publish()
```
jace, line 162: You've got this twice. Move this into a `published_project` fixture.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,195 @@
...
+@given("the server uses Recaptcha")
+def given_server_uses_recaptcha(
+    live_server, db_session, user_twoflower, new_project, funnel
+) -> None:
+    user_twoflower.password = TWOFLOWER_PASSWORD
+    user_twoflower.add_phone(TWOFLOWER_PHONE)
+    user_twoflower.add_email(TWOFLOWER_EMAIL)
+    new_project.publish()
+    db_session.add(new_project)
+    db_session.commit()
```
jace, line 177: One again, two reusable fixtures.

jace (changes requested):
There's a lot of repetition that should be moved into fixtures. One OTP extraction needs regex instead of the string.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+pytestmark = pytest.mark.filterwarnings(
+    "ignore:Object of type <AccountPhone> not in session",
+    "ignore:Object of type <AccountEmail> not in session",
+)
+
+
+@pytest.fixture()
+def published_project(db_session, new_project):
+    new_project.publish()
+    db_session.add(new_project)
```
jace, line 26: This is not needed. It's already in the session.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+scenarios('account/register.feature')
+pytestmark = pytest.mark.usefixtures('live_server')
+
+TWOFLOWER_EMAIL = 'twoflower@example.org'
+TWOFLOWER_PHONE = '+12015550123'
+TWOFLOWER_PASSWORD = 'te@pwd3289'  # nosec
+
+pytestmark = pytest.mark.filterwarnings(
+    "ignore:Object of type <AccountPhone> not in session",
+    "ignore:Object of type <AccountEmail> not in session",
```
jace, line 19: Don't filter these. The problem is in the `add` classmethod of EmailAddress and PhoneNumber. It should be resolved there.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+)
+
+
+@pytest.fixture()
+def published_project(db_session, new_project):
```
jace, line 24: Specify types: `new_project: models.Project) -> models.Project:`

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+    new_project.publish()
+    db_session.add(new_project)
+    db_session.commit()
+    return new_project
+
+
+@pytest.fixture()
+def user_twoflower_with_password_and_contact(db_session, user_twoflower):
```
jace, line 32: Also needs type definitions here.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+    user_twoflower.password = TWOFLOWER_PASSWORD
+    user_twoflower.add_phone(TWOFLOWER_PHONE)
+    user_twoflower.add_email(TWOFLOWER_EMAIL)
+    db_session.add(user_twoflower)
```
jace, line 36: This line is not required.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+    ),
+    target_fixture='anon_username',
+)
+def when_anonuser_navigates_login_and_submits(
+    app, live_server, phone_or_email: str, page: Page
+) -> dict[str, str]:
+    if phone_or_email == "a phone number":
+        username = '8123456789'
+    elif phone_or_email == "an email address":
+        username = 'anon@example.com'
```
jace, line 67: I think you should put both these constants at the top of the file. We'll have to audit for where we're using these values across teats and put them into shared fixtures – apart from ensuring we're not using real data.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+
+@then("they are prompted for their name and the OTP, which they provide")
+def then_anonuser_prompted_name_and_otp(live_server, anon_username, page: Page) -> None:
+    page.wait_for_selector('input[name=fullname]').fill('Twoflower')
+    if anon_username['phone_or_email'] == "a phone number":
+        otp = live_server.transport_calls.sms[-1].vars['otp']
+    elif anon_username['phone_or_email'] == "an email address":
+        subject = live_server.transport_calls.email[-1].subject
+        otp_pattern = r"\b\d{4}\b"
+        otp = re.search(otp_pattern, subject).group(0)
```
jace, line 85: The var is not needed. Specify the pattern directly here – but with single quotes as it's not a UI string.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+    page.click('#form-otp button')
+
+
+@then("they get an account and are logged in")
+def then_they_are_logged_in(
+    live_server, user_twoflower_with_password_and_contact, page: Page
+) -> None:
+    assert (
+        page.wait_for_selector('.alert__text').inner_text()
+        == "You are now one of us. Welcome aboard!"
```
jace, line 98: This string is going to be a problem as it's UI and could change unexpectedly. It can stay for now, but we'll need a better marker (future PR).

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+    )
+
+
+@given("Twoflower visitor is on the home page")
+def when_twoflower_visits_homepage(
+    live_server, page: Page, db_session, user_twoflower_with_password_and_contact
+) -> None:
+    page.goto(live_server.url)
+    db_session.commit()
```
jace, line 107: Always do db operations first, browser navigation second.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+@when("they click on follow")
+def when_they_click_follow(page: Page) -> None:
+    page.wait_for_selector("#register-nav").click()
+
+
+@then("a register modal appears")
+def then_register_modal_appear(page: Page) -> None:
+    assert (
+        page.wait_for_selector('xpath=//*[@id="passwordform"]/p[2]').inner_text()
+        == "Tell us where you’d like to get updates. We’ll send an OTP to confirm."
```
jace, line 158: Another UI string to change for the next PR.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+@when(
+    parsers.re("they enter (?P<phone_or_email>a phone number|an email address)"),
+    target_fixture='anon_username',
+)
+def when_they_enter_email(page: Page, phone_or_email: str) -> dict[str, str]:
+    wait_until_recaptcha_loaded(page)
+    if phone_or_email == "a phone number":
+        username = '8123456789'
+    elif phone_or_email == "an email address":
+        username = 'anon@example.com'
```
jace, line 171: Two more constants.

tests/e2e/account/pw_register_test.py
```diff
@@ -0,0 +1,215 @@
...
+    page.wait_for_selector('input[name=username]').fill(username)
+    page.click('#form-passwordlogin button')
+    return {'phone_or_email': phone_or_email, 'username': username}
+
+
+@given("Twoflower is on the project page")
+def given_twoflower_visits_project(
+    live_server,
+    page: Page,
+    db_session,
```
jace, line 183: This doesn't strictly matter, but db_session should be first, then database fixtures, then the server, then the browser.

I remember there was an issue with fixture order in one pytest-bdd test. Reordering fixed it.

jace (changes requested):
Some cosmetic changes left. You should also remove the pytest-selenium tests, remove the dependency, remove the pin on selenium as required by pytest-selenium, and also remove the ignore rule for it in pip-audit. Also remove the `playright` prefix on the file names – this PR should completely switch over from pytest-selenium to playwright.

jace (approved):

PR #1816, opened 2023-07-21
Changed messaging for the 403 template

miteshashar:


funnel/templates/errors/403.html.jinja2
```diff
@@ -4,10 +4,10 @@
 {% block top_title %}
   <div class="mui--text-center">
     <img src="{{ url_for('static', filename='img/error-403.svg') }}" class="emoticon" alt="{% trans %}Unavailable{% endtrans %}"/>
-    <h1 class="mui--text-light mui--text-display1">{% trans %}Unavailable{% endtrans %}</h1>
+    <h1 class="mui--text-light mui--text-display1">{% trans %}Forbidden{% endtrans %}</h1>
```
jace, line 7: We used `Unavailable` because the `Forbidden` word suggests the user attempted something bad. They just opened a link. Revert this.

jace (changes requested):

funnel/templates/errors/403.html.jinja2
```diff
@@ -4,10 +4,10 @@
...
   </div>
 {% endblock top_title %}
 
 {% block content %}
-  <p class="mui--text-subhead mui--text-center">{% trans %}This page is not currently available{% endtrans %}</p>
+  <p class="mui--text-subhead mui--text-center">{% trans %}You do not have access to this page{% endtrans %}</p>
```
jace, line 12: This can be friendlier too:

* This page is private and you do not have access
* This page is private. If you were sent a link, ask them to give you access

Second version will need to explain how. We allow user search by full name (ambiguous), username (precise but optional), or email address (precise but useless if user account has phone number only).

jace (approved):

PR #1823, opened 2023-07-27
Send 24-hours advance notification for in-person sessions

djamg:


funnel/cli/periodic/notification.py
```diff
@@ -44,3 +44,33 @@ def project_starting_alert() -> None:
                 fragment=project.next_session_from(use_now + timedelta(minutes=10)),
             )
         )
+
+
+@periodic.command('project_starting_tomorrow_alert')
+def project_starting_tomorrow_alert() -> None:
+    """Send notifications for projects starting the next day."""
```
jace, line 51: This function isn't timezone aware. It's calculating tomorrow at UTC, not the project or user timezone. It doesn't also specify periodicity. Is this safe to run multiple times per day? What time in the day should the alert go out?

jace (changes requested):

jace (dismissed):

jace (dismissed):

jace (approved):

PR #1824, opened 2023-07-27
Removed footer from project starting email notification

anishTP:


jace:
Already resolved in #1828.

PR #1839, opened 2023-08-03
Show past featured sessions with videos on profile page

vidya-ram:


jace (approved):
This works, but only because we don't have too many past featured sessions. The infinite scroll will block the rest of the page otherwise. We need videos to be a separate tab to make this a permanent feature.

PR #1867, opened 2023-09-05
Add separate page for generating short links with custom tags.

### LINKED ISSUE #1738: Shortlinks generated in UI need a way to incorporate additional tags
When a shortlink is generated through the "Share" interface in UI (either webshare, or share from the custom menu which has copy, email, social etc), the shortlink includes the URL and specific UTM tags: `utm_source` (`webshare` or `share-*`) and `utm_medium` (`email` or `social` if not using webshare). There is no facility to use other UTM tags such as `utm_campaign`, nor for any other form of tagging.

This requirement of additional tags is not generic, and therefore not necessary for all users. However, it is a definite requirement for internal use. This issue is for discussing possible solutions.

vidya-ram:
Fixes #1738


funnel/templates/siteadmin_generate_shortlinks.html.jinja2
```diff
@@ -0,0 +1,28 @@
+{% extends "layout.html.jinja2" %}
+
+{% block title %}{% trans %}Get shortlink with additional tags{% endtrans %}{% endblock title %}
+
+{% block content %}
+  <div class="display-inlineblock search search--small search--icon mui--z1">
+    <div class="textfield">
+      {% if current_auth.user and current_auth.user.is_site_editor %}
```
jace, line 8: There's an unnecessary gate here.

jace (changes requested):

PR #1900, opened 2023-10-10
Vimeo fix

djamg:


funnel/assets/js/utils/embedvideo.js
```diff
@@ -29,11 +34,11 @@ const Video = {
   },
   embedIframe(videoWrapper, videoUrl) {
     let videoEmbedUrl = '';
-    const { type, videoId } = this.getVideoTypeAndId(videoUrl);
+    const { type, videoId, paramId } = this.getVideoTypeAndId(videoUrl);
     if (type === 'youtube') {
       videoEmbedUrl = `<iframe src='//www.youtube.com/embed/${videoId}' frameborder='0' allowfullscreen></iframe>`;
     } else if (type === 'vimeo') {
-      videoEmbedUrl = `<iframe src='https://player.vimeo.com/video/${videoId}' frameborder='0' allowfullscreen></iframe>`;
+      videoEmbedUrl = `<iframe src='https://player.vimeo.com/video/${videoId}?h=${paramId}' frameborder='0' allowfullscreen></iframe>`;
```
jace, line 41: This is not accounting for when paramId is not present (all existing videos).

funnel/assets/js/utils/embedvideo.js
```diff
@@ -8,14 +8,19 @@ const Video = {
...
     let type = '';
     if (regexMatch && regexMatch.length > 5) {
       if (regexMatch[3].indexOf('youtu') > -1 || regexMatch[3].indexOf('y2u') > -1) {
         type = 'youtube';
       } else if (regexMatch[3].indexOf('vimeo') > -1) {
         type = 'vimeo';
+        return {
+          type,
+          videoId: regexMatch[6],
+          paramId: regexMatch[9],
```
jace, line 22: Use named patterns. Numbers are fragile.

jace (changes requested):
Why do we even have this client-side regex? It's fragile and needs multiple test cases.

funnel/assets/js/utils/embedvideo.js
```diff
@@ -8,14 +8,22 @@ const Video = {
...
     let type = '';
     if (regexMatch && regexMatch.length > 5) {
-      if (regexMatch[3].indexOf('youtu') > -1 || regexMatch[3].indexOf('y2u') > -1) {
+      if (
+        regexMatch.groups.service.indexOf('youtu') > -1 ||
+        regexMatch.groups.service.indexOf('y2u') > -1
+      ) {
         type = 'youtube';
-      } else if (regexMatch[3].indexOf('vimeo') > -1) {
+      } else if (regexMatch.groups.service.indexOf('vimeo') > -1) {
```
vidya-ram, line 20: @djamg Have you tested it for different youtube embed urls?

jace: Good to merge after this is confirmed.

funnel/assets/js/utils/embedvideo.js
```diff
@@ -8,14 +8,22 @@ const Video = {
...
         type = 'vimeo';
+        return {
+          type,
+          videoId: regexMatch.groups.videoId,
+          paramId: regexMatch.groups.paramId,
```
vidya-ram, line 25: This one return moved to the end of the condition would suffice. `paramId` will be null incase there is no match found.

vidya-ram (dismissed):

funnel/assets/js/utils/embedvideo.js
```diff
@@ -8,13 +8,16 @@ const Video = {
   */
   getVideoTypeAndId(url) {
     const regexMatch = url.match(
-      /(http:|https:|)\/\/(player.|www.)?(y2u\.be|vimeo\.com|youtu(be\.com|\.be|be\.googleapis\.com))\/(video\/|embed\/|live\/|watch\?v=|v\/)?([A-Za-z0-9._%-]*)(&\S+)?/
+      /(http:|https:|)\/\/(player.|www.)?(?<service>y2u\.be|vimeo\.com|youtu(be\.com|\.be|be\.googleapis\.com))\/(video\/|embed\/|live\/|watch\?v=|v\/)?(?<videoId>[A-Za-z0-9._%-]*)(&\S+)?((\?h=)(?<paramId>[^&]+))?/
```
jace, line 11: IMO, rename `service` to `hostname`.

`\?h=` will miss `url?something=other&h=paramid`, so it should be `(\?|&)h=`, but even better is if you replace this regex with a proper URL parser.

funnel/assets/js/utils/embedvideo.js
```diff
@@ -8,13 +8,16 @@ const Video = {
...
     let type = '';
     if (regexMatch && regexMatch.length > 5) {
-      if (regexMatch[3].indexOf('youtu') > -1 || regexMatch[3].indexOf('y2u') > -1) {
+      if (
+        regexMatch.groups.service.indexOf('youtu') > -1 ||
+        regexMatch.groups.service.indexOf('y2u') > -1
+      ) {
         type = 'youtube';
-      } else if (regexMatch[3].indexOf('vimeo') > -1) {
+      } else if (regexMatch.groups.service.indexOf('vimeo') > -1) {
```
jace, line 20: Change to `regexMatch.groups.hostname === 'vimeo.com'` — look for the full match instead of the word appearing anywhere in between.

funnel/forms/project.py
```diff
@@ -167,6 +167,7 @@ class ProjectLivestreamForm(forms.Form):
                             'y2u.be',
                             'www.vimeo.com',
                             'vimeo.com',
+                            'player.vimeo.com',
```
jace, line 170: This doesn't match the regex, which will also `player.youtube.com` even though that is invalid.

jace (changes requested):
This regex is getting too complex and should be replaced with a URL parser. At the very least, it needs test cases.

PR #1901, opened 2023-10-13
Added signal and handler for project data changes.

anishTP:
Added column for thumbnail images in project table and ran db migration.


funnel/signals.py
```diff
@@ -56,6 +56,7 @@
 org_data_changed = app_signals.signal('org-data-changed')
 team_data_changed = app_signals.signal('team-data-changed')
 session_revoked = app_signals.signal('session-revoked')
+project_data_change = app_signals.signal('project_data_change')
```
jace, line 59: As per convention, the string name should use hyphens.

funnel/views/project.py
```diff
@@ -283,6 +283,8 @@ def new_project(self) -> ReturnView:
 
             flash(_("Your new project has been created"), 'info')
 
+            # trigger signal to generate thumbnail image
+            project_data_change.send(self.obj)
```
jace, line 287: This should be before the db commit, or commit again after.

funnel/views/project.py
```diff
@@ -455,6 +457,8 @@ def edit(self) -> ReturnView:
             db.session.commit()
             flash(_("Your changes have been saved"), 'info')
             tag_locations.queue(self.obj.id)
+            # trigger signal to generate thumbnail image
+            project_data_change.send(self.obj)
```
jace, line 461: Send before db commit.

funnel/views/thumbnails.py
```diff
@@ -0,0 +1,10 @@
+"""View for autogenerating thumbnail previews."""
+
+from __future__ import annotations
+
+from ..signals import project_data_change
+
+
+@project_data_change.connect
```
jace, line 8: This file shouldn't be in this PR.

migrations/versions/2b9456f4e8fd_added_thumbnail_image_column_to_project_.py
```diff
@@ -0,0 +1,45 @@
+"""Added thumbnail_image column to Project model
```
jace, line 1: This file shouldn't be in this PR.

funnel/views/project.py
```diff
@@ -278,11 +278,12 @@ def new_project(self) -> ReturnView:
             project = Project(created_by=current_auth.user, account=self.obj)
             form.populate_obj(project)
             project.make_name()
+            # trigger signal to generate thumbnail image
+            project_data_change.send(self.obj)
```
jace, line 282: `self.obj` here is account, not project. Send `project`, and after it is added to the db session, but before commit.

funnel/views/project.py
```diff
@@ -452,6 +453,8 @@ def edit(self) -> ReturnView:
         form = ProjectForm(obj=self.obj, account=self.obj.account, model=Project)
         if form.validate_on_submit():
             form.populate_obj(self.obj)
+            # trigger signal to generate thumbnail image
```
jace, line 456: That's not the reason for this signal, so the comment is incorrect. Best to remove. It's a _generic_ data change signal that's being used elsewhere to generate a thumbnail, so the comment should be on the receiving side.

funnel/views/thumnbnails.py
```diff
@@ -0,0 +1,15 @@
+"""View for autogenerating thumbnail previews."""
```
jace, line 1: This file should also be removed from this PR.

jace (changes requested):

jace (approved):

PR #1902, opened 2023-10-13
Add a preview_image column to Project model

anishTP:
This PR only adds a new `bytea` column to the project model, meant for hosting an auto-generated image for Open Graph previews.


funnel/models/project.py
```diff
@@ -191,6 +192,13 @@ class Project(UuidMixin, BaseScopedNameMixin, Model):
         read={'all'},
         datasets={'primary', 'without_parent', 'related'},
     )
+
+    thumbnail_image = with_roles(
+        sa.orm.mapped_column(LargeBinary, nullable=True),
+        read={'all'},
+        datasets={'primary', 'without_parent', 'related'},
```
jace, line 199: This can't be in JSON datasets as it's not text. Remove this line.

funnel/models/project.py
```diff
@@ -191,6 +192,13 @@ class Project(UuidMixin, BaseScopedNameMixin, Model):
         read={'all'},
         datasets={'primary', 'without_parent', 'related'},
     )
+
+    thumbnail_image = with_roles(
+        sa.orm.mapped_column(LargeBinary, nullable=True),
```
jace, line 197: Use `sa.LargeBinary` and drop the import. It'll keep the module namespace clean of clutter.

migrations/versions/ce340b5b9d1e_added_thumbnail_image_column_to_project_.py
```diff
@@ -0,0 +1,44 @@
...
+
+def downgrade(engine_name: str = '') -> None:
+    """Downgrade all databases."""
+    # Do not modify. Edit `downgrade_` instead
+    globals().get(f'downgrade_{engine_name}', lambda: None)()
+
+
+def upgrade_() -> None:
+    """Upgrade default database."""
+    # ### commands auto generated by Alembic - please adjust! ###
```
jace, line 33: Remove these lines. They're a reminder that you should review before committing.

funnel/models/project.py
```diff
@@ -191,6 +192,13 @@ class Project(UuidMixin, BaseScopedNameMixin, Model):
         read={'all'},
         datasets={'primary', 'without_parent', 'related'},
     )
+
+    thumbnail_image = with_roles(
```
jace, line 196: This is technically a preview or a cover image, not a thumbnail. We need a more appropriate name.

jace (changes requested):

funnel/models/project.py
jace, line 196: The column needs a type definition, so this should be:

`thumbnail_image: Mapped[bytes | None] = ...`

(Pending column name change.)

jace, line 196: I think we can drop the `with_roles` wrapper entirely. There's no need for any template to access the contents of this field.

We also don't want this to be loaded most of the time, so the column definition needs to be `sa.orm.mapped_column(sa.LargeBinary, nullable=True, deferred=True)`

jace (changes requested):

jace (approved):

PR #1907, opened 2023-10-18
Remove trailing slashes for the incoming urls in shortlink app

djamg:


jace (changes requested):
The problem is not here. Where is the trailing slash being added?

jace:
Closing this as it will break for URLs that do have a trailing slash, such as project URLs.

PR #1911, opened 2023-10-26
Don't send comment reply notifications to all subscribers

djamg:


funnel/models/notification_types.py
```diff
@@ -173,7 +173,7 @@ class NewCommentNotification(Notification[Commentset, Comment], type='comment_ne
     title = __("When there is a new comment on something I’m involved in")
     exclude_actor = True
 
-    roles = ['replied_to_commenter', 'document_subscriber']
+    roles = ['replied_to_commenter', 'project_crew']
```
jace, line 176: We explicitly set comment alerts to go to document subscribers. This is a product change and needs discussion.

jace (changes requested):

jace (approved):

PR #1912, opened 2023-10-27
Show placeholder page for private accounts

vidya-ram:


funnel/views/profile.py
```diff
@@ -62,6 +62,15 @@ def feature_profile_make_private(obj: Account):
     return obj.current_roles.admin and obj.make_profile_private.is_available
 
 
+@Account.features('is_private')
+def feature_profile_is_private(obj: Account):
+    return (
+        obj.is_user_profile
```
jace, line 68: This test shouldn't be there.

funnel/views/profile.py
```diff
@@ -62,6 +62,15 @@ def feature_profile_make_private(obj: Account):
...
+        and not obj.current_roles.admin
+        and not bool(obj.profile_state.ACTIVE_AND_PUBLIC)
```
jace, line 70: These two tests are fine.

jace (changes requested):

jace (approved):

PR #1922, opened 2023-11-06
Switch from typing_extensions to typing for Python 3.11 imports

jace:
Also change dependencies to refer to unreleased code on git where required.


jace:
Python 3.12 builds are still failing courtesy `aiohttp`, which includes support in version `aiohttp>=3.9`, which is still in beta and has significant changes.

djamg (approved):

PR #1931, opened 2023-11-22
Fit two button in mobile

vidya-ram:


jace (approved):
I'm unable to validate this from visual review, so I'm assuming you've tested this. Playwright tests for the various registration criteria will be nice.

PR #1968, opened 2024-02-06
Make restricted session video available to account members

### LINKED ISSUE #1966: Restricted access session videos needs to visible to account members
### Context
Session videos that are restricted to participants are currently visible to only project participants and not account level members. This requires adding account members to project (syncing membership ticket item collection to project)

### Solution
As account membership grants viewership to all content in an account, restricted videos can be made available to account members. This will ease the workflow as members needs not be added to projects.

vidya-ram:
Fixes #1966


jace (approved):
We could use a `feature` tag instead of repeating `or` conditions each time, but this works.

PR #1971, opened 2024-02-12
Change the order of `Membership` section on profile page

vidya-ram:
Move it above `Upcoming` section (First section on the page when there is no featured project)

### DIFF
--- a/funnel/templates/profile.html.jinja2
+++ b/funnel/templates/profile.html.jinja2
@@ -144,72 +144,14 @@
 
     {{ featured_section(featured_project) }}
 
-    {% if membership_project %}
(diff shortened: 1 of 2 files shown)


funnel/templates/profile.html.jinja2
```diff
@@ -210,6 +209,7 @@
       </div>
     {% endif %}
 
+    {{ upcoming_section(upcoming_projects) }}
```
jace, line 212: One twist please: move it up only if the current user is not already a member. Existing members should see upcoming projects first.

jace (changes requested):

jace (approved):

PR #1981, opened 2024-02-28
Support for content layouts in Markdown

jace:


jace:
CSS as UGC may be desirable as it allows users to define custom layouts independently. Since Funnel is designed to segment permissions at the account and project level, custom stylesheets can be located at either level (TBD as a separate ticket). Within the Markdown parser, this can be supported by enabling the attrs plugin with a custom validator:

* `#id` is rewritten to `#md:id`, inserting an `md:` namespace prefix
* `.class` is rewritten to `.md--class` or (if valid) `.md:class` (may need escaping to `.md\:class` in CSS syntax)
* Other attributes (`attr=value` syntax) are restricted or disabled. Image tags _may_ need to allow `size` and `width`

jace:
Image or image gallery embeds could use the container plugin for image(s) + caption. Typical layouts:

1. Splash image (full width without margins, typically as a header, but sometimes mid-page too)
2. Featured image (large, but with margins)
3. Medium (centered on page, not excessively large)
4. Sidebar (offset to a side, typically the first element in a paragraph of text)

```markdown
::: splash-image
![Alt text](/path/to/image)
![Second image creates a carousel](/path/to/image)

Caption _with_ Markdown markup
:::
```

```markdown
::: gallery
![Alt text](/path/to/image)
![Both images appear as clickable thumbnails](/path/to/image)

Caption _with_ Markdown markup
:::
```

These blocks are parsed as Markdown and the container markup (`:::`) merely adds a `<div>` with an appropriate system CSS class (not user customisable). The appearance as carousel or thumbnail gallery is controlled by CSS+JS and depends on the block not containing anything that disrupts the layout, such as text between images.

jace:
The [Attribution plugin](https://github.com/dweidner/markdown-it-attribution) upgrades a regular blockquote to add a source citation if the last line is prefixed with `--`. Needs to be ported to Python.

jace:
Candidate syntax for `details/summary` tags, [originally proposed here](https://talk.commonmark.org/t/html-details-tag/759/10):

```markdown
>-> Summary here
> Details here
```

Output:

<details><summary>Summary here</summary>Details here</details>

Default output without the extension:

>-> Summary here
> Details here

PR #1991, opened 2024-03-21
Auto fill user details in boxoffice payment form for logged in users

vidya-ram:


funnel/templates/project_layout.html.jinja2
```diff
@@ -532,7 +532,10 @@
           widgetElem: "#boxoffice-widget",
           org: {{ project.boxoffice_data.org|tojson }},
           itemCollectionId: {{ project.boxoffice_data.item_collection_id|tojson }},
-          itemCollectionTitle: {{ project.title|tojson }}
+          itemCollectionTitle: {{ project.title|tojson }},
```
jace, line 535: Shouldn't this also be "menuTitle"?

funnel/templates/project_layout.html.jinja2
```diff
@@ -532,7 +532,10 @@
...
+          userFullname: '{{current_auth.user.fullname }}',
```
jace, line 536: Use `|tojson` for all three. Names can contain quote characters.

jace (changes requested):

jace (approved):

PR #2001, opened 2024-03-27
Added timeout for playwright tests

djamg:


jace (approved):
I expected a nicer API for this, but apparently this is how it should be done. Wow!

PR #2004, opened 2024-03-27
Add calendar to accounts

vidya-ram:


funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
 
         return ctx
 
+    def jsonify_calendar(jsonData):
+        return jsonify(jsonData['projects'])
```
jace, line 247: Several problems:

1. As a class method, the first parameter should be `self`, or the method should be decorated with `@staticmethod`
2. The parameter name should be in Python convention, snake-cased: `json_data`
3. This method can be entirely replaced with a lambda function since it has a single line

funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
...
+
+    @route('calendar', methods=['GET'], endpoint='calendar')
+    @requestargs('start', 'end')
+    @render_with(
+        {
+            'text/html': 'profile_calendar.html.jinja2',
+            'application/json': jsonify_calendar,
```
jace, line 254: `'application/json': lambda result: jsonify(result['projects']),`

funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
...
+        }
+    )
+    def org_calendar(
+        self, start: datetime = None, end: datetime = None
+    ) -> ReturnRenderWith:
+        projects = []
+        if start is not None and end is not None:
+            start = datetime.strptime(start, "%Y-%m-%dT%H:%M:%S%z")
+            end = datetime.strptime(end, "%Y-%m-%dT%H:%M:%S%z")
```
jace, line 263: The values were declared as datetime, but here they are assumed to be strings. Also, these calls can fail with bad data. Use `parse_isoformat` instead of `strptime`.

funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
 
         return ctx
 
+    def jsonify_calendar(jsonData):
+        return jsonify(jsonData['projects'])
+
+    @route('calendar', methods=['GET'], endpoint='calendar')
+    @requestargs('start', 'end')
```
jace, line 250: To process these as datetime values, use:

```python
from coaster.utils import parse_isoformat
...
    @requestargs(('start', parse_isoformat), ('end', parse_isoformat))
```

funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
 
         return ctx
 
+    def jsonify_calendar(jsonData):
+        return jsonify(jsonData['projects'])
+
+    @route('calendar', methods=['GET'], endpoint='calendar')
```
jace, line 249: `methods=['GET']` is the default and can be removed here.

funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
...
+    @requestargs('start', 'end')
+    @render_with(
+        {
+            'text/html': 'profile_calendar.html.jinja2',
+            'application/json': jsonify_calendar,
+        }
+    )
+    def org_calendar(
```
jace, line 257: Rename to just `calendar`

funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
...
+        self, start: datetime = None, end: datetime = None
```
jace, line 258: If `None` is a valid value, the types should be `self, start: datetime | None = None, end: datetime | None = None`

funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
...
+                        'start': p.start_at,
+                        'end': p.end_at,
+                        'date_str': datetime_filter(
+                            p.start_at_localized, format='dd MMM yyyy'
+                        ),
+                        'time': datetime_filter(p.start_at_localized, format='hh:mm a'),
+                        'venue': (
+                            p.primary_venue.city if p.primary_venue else p.location
+                        ),
+                        'cfp_open': True if p.cfp_state.OPEN else False,
```
jace, line 286: `'cfp_open': bool(p.cfp_state.OPEN),`

funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
...
+                        'member_access': (
+                            True
+                            if (p.features.rsvp_for_members or p.features.subscription)
+                            else False
```
jace, line 290: `bool(p.features.rsvp_for_members or p.features.subscription)`

funnel/views/profile.py
```diff
@@ -242,6 +243,61 @@ def view(self) -> ReturnRenderWith:
...
+                        ),
+                        'url': p.url_for(),
+                    }
+                    for p in filtered_projects
+                ]
+        return {
+            'profile': self.obj.current_access(datasets=('primary', 'related')),
```
jace, line 297: If the model declares `__json_datasets__`, that is used automatically, so the `datasets` parameter can be removed here.

jace (changes requested):

jace (approved):

PR #2005, opened 2024-03-27
Support account follow

jace:


funnel/templates/macros.html.jinja2
```diff
@@ -271,6 +264,19 @@
   </table>
 {%- endmacro %}
 
+{%- macro addFollowButton(profile, css_class='') %}
```
jace, line 267: Jinja style is lowercase names, so this should be `add_follow_button`

PR #2010, opened 2024-04-05
Grant `project_` prefixed roles

jace:
This is needed for project notifications where the session is an optional fragment.


jace:
GitHub just had an outage that's causing CI to fail here. Tests are passing locally. https://www.githubstatus.com/incidents/5ly0psff2s5d

PR #2014, opened 2024-04-08
Update CSS selector for schedule container

anishTP:


funnel/assets/sass/pages/schedule.scss
```diff
@@ -270,6 +270,12 @@
     max-height: 50vh;
     overflow: scroll;
   }
+  .schedule-grid.schedule-grid--compressed .schedule .schedule__date {
+    display: none;
+  }
+  .schedule-grid.schedule-grid--compressed .schedule .schedule__date--upcoming {
+    display: block;
+  }
```
jace, line 278: This is inefficient because the backend is still expending compute power to calculate and render the schedule. If the schedule is not being shown, the code should be skipped.

What we need: a "show previous" action that loads the previous schedule.

jace (changes requested):

PR #2019, opened 2024-04-16
Styling changes to the profile calendar widget

anishTP:


jace:
Issues in mobile view:
1. The filter icon is not aligned with the right chevron.
2. Some extraneous element is on the right side of the calendar.
3. The Projects list has a white patch below.

imageimage

jace:
In desktop view, the calendar has a white frame visible at bottom.

image

jace:
Widget issues:

* Left side: Left chevron and label aren't visually aligned
* Right side: Chevrons have hover+active state that stays after a click, but filter icon doesn't have it. Hover+active state is not required here.
* Chevron tap targets are reduced from before, which may be an issue on mobile
* Calendar days have an underline on hover but aren't clickable (see day 1)
* Label text continues to overlap day numbers on the next line
image

PR #2030, opened 2024-04-22
Project page header

anishTP:
- Moved the register block to the project-banner_right
- Added a venue card to the right_col below navbar on the project page.
- Changed order of the navbar render on mobile


jace (changes requested):

jace:
The layout needs cleanup for some cases:

Member box alignment and missing icon for venue:
image

Everything is left aligned when there's no image, but the button's visual alignment is off:
image

Here the colour contrast works in the button's favour, but it needs margin from the text above it. The second button declaring "You are a member" is confusing as it has no relevance to this header:
image

anishTP:
> The layout needs cleanup for some cases:
> 
> Member box alignment and missing icon for venue: image

Changed the size of the button to full-width when its the only child.
Replaced map-pin with map-marker-alt icon for the venue.
Screenshot 2024-04-29 at 5 11 52 PM


> 
> Everything is left aligned when there's no image, but the button's visual alignment is off: image

Added padding to above the button when there is no image and start-date. Also made it full-width when its the only child.

Screenshot 2024-04-29 at 5 10 52 PM

> 
> Here the colour contrast works in the button's favour, but it needs margin from the text above it. The second button declaring "You are a member" is confusing as it has no relevance to this header: image

The membership button could be removed since the header already has a label showing membership status. But it is necessary when the user is not a member.
Screenshot 2024-04-29 at 5 10 37 PM

funnel/templates/project_layout.html.jinja2
```diff
@@ -358,7 +338,7 @@
             </div>
           {%- elif sponsorship.member.description %}
             <div class="mui--text-body2 mui--text-light margin-bottom">
-              {{ sponsorship.member.description|preview(min=200, max=300) }}
+              {{ sponsorship.member.description.html|preview(min=200, max=300) }}
```
jace, line 341: This has been fixed in Baseframe. Using `.html` for the `|preview` filter is no longer necessary.

jace (changes requested):

jace (dismissed):

PR #2052, opened 2024-05-06
Fix follow btn

anishTP:
Improved styling on the follow buttons with icon. Fixed button width on icon-only buttons and fixed the sizing and spacing of icons in text button with icon in profile headers.


jace:
image

The calendar widget is missing in project cards on the home and profile pages. Unclear why.

PR #2060, opened 2024-05-16
Remove schedule from the featured project card

anishTP:


jace (commented):
IMO, change this to only show the schedule when the event is ongoing or within a day before. This will require a helper method.

anishTP:
> IMO, change this to only show the schedule when the event is ongoing or within a day before. This will require a helper method.

There is a section in the card that links to live schedule when the session is on. Shall I append to this?

jace (approved):

PR #2081, opened 2024-06-24
Fixed the project header right column width to prevent layout shift

anishTP:
Banner
Screenshot 2024-07-01 at 10 27 20 PM

No banner or video
Screenshot 2024-07-01 at 10 27 03 PM

With livestream
Screenshot 2024-07-01 at 10 27 11 PM
Screenshot 2024-07-01 at 10 29 28 PM

Screen width between 768px and 992px
Screenshot 2024-07-01 at 10 30 23 PM


funnel/assets/sass/pages/project.scss
```diff
@@ -459,15 +459,15 @@
...
+      width: 33%;
     }
   }
   .project-banner.project-banner--inner {
     .project-banner__left--smaller {
       width: 25%;
     }
     .project-banner__right--smaller {
-      width: auto;
+      width: 75%;
```
jace, line 470: Shouldn't these be `max-width`, with `width` itself remaining `auto`?

anishTP: But in this case we don't want the left column to exceed 67% in any scenario. Setting `width` to `auto` for the right column + `max-width` to 33% would result in a left column larger than 67% when there's no enough content in the right column.

jace: The left column's widths are not changing, so it should not be affected by the right column having a max-width. Please test and confirm. This PR should have had before/after screenshots for all scenarios.

jace (approved):

PR #2103, opened 2024-07-31
Changed the order of details in the profile-details to improve layout

anishTP:
#### Before
Screenshot 2024-07-31 at 12 50 36 PM

#### After
Screenshot 2024-07-31 at 12 50 20 PM

### DIFF
--- a/funnel/assets/sass/pages/profile.scss
+++ b/funnel/assets/sass/pages/profile.scss
@@ -213,6 +213,13 @@
   }
   .profile-details {
     padding: 2 * $mui-grid-padding 0 0;
+    ul {
+      padding: 0;
+      display: flex;
+      flex-direction: row;
+      gap: 16px;
+      list-style: none;
+    }
   }
   .profile.profile--unverified {
     .profile__banner__box {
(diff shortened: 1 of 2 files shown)


jace:
The bio is optional and of variable length, so the metadata (joined date, website) will float. IMO, these should be above to fix their position.

jace:
This also needs screenshots in mobile layout to confirm the elements stack correctly.

anishTP:
> The bio is optional and of variable length, so the metadata (joined date, website) will float. IMO, these should be above to fix their position.

When the bio is empty there won't be any empty space, if that is what you mean by metadata will float.
Screenshot 2024-07-31 at 1 22 20 PM

anishTP:
> This also needs screenshots in mobile layout to confirm the elements stack correctly.

#### Before
Screenshot 2024-07-31 at 1 21 56 PM

#### After
Screenshot 2024-07-31 at 1 21 28 PM

zainabbawa:
Screenshot 2024-07-31 1 28 26 PM
Here's a reference for how Twitter/X.com does it.

jace:
Let's go with bio _above_ metadata then.

jace (approved):

